PR #1043, opened 2014-11-28
adding logout to the filter

atarkowska:
as indicated in http://lists.openmicroscopy.org.uk/pipermail/ome-users/2014-November/004914.html

To test. Setup public user with a following filter

```
$ bin/omero config set omero.web.public.url_filter '^/(?!webadmin|webclient/logout/|webclient/action/\w+|webclient/annotate_(file|tags|comment))'
```
1. Go to http://host/webclient. There will be no login page. You will be redirected directly to the public user data.
2. Click logout, you should be redirected to login page.
3. Log in as regular user. You should see own data.
4. Log out again. You should be redirect to public user data.


aleksandra-tarkowska:
--rebased-to #1060

PR #1055, opened 2014-12-15
Update CreateApp docs

will-moore:
Updates the web CreateApp page to address a couple of small issues:
 - previous instructions specified that the app directory itself was on the ```PYTHONPATH``` whereas it should be the parent directory that is on the ```PYTHONPATH```.
 - import statements in the urls.py example don't work with 5.0 OMERO

To test, follow the CreateApp instructions until you have the "Welcome" message at a new app /index.html.


manics:
Looks good to me

hflynn:
Please rebase to develop unless it doesn't apply there @will-moore (in which case, mark as such)

will-moore:
--rebased-to #1063

PR #1058, opened 2015-01-01
omero: Add detailed version requirements

rleigh-dundee:
Conversion of https://docs.google.com/spreadsheets/d/1TRJUhdMZ_pnHxg1TUk9Lf_oQcx2paZtwWMjTybQ7qZo/edit#gid=0 to sphinx markup

This isn't intended to mirror the installation prerequisites; it's primarily of interest to ourselves and sysadmins who need to know what's supported for the deployment timeframe of an OMERO installation.  From that point of view it's a much more detailed overview of each prerequisite, which could be cross-referenced from the install docs.  This is a detailed list of all of the released versions of our prerequisites, and their release and support timeframes.  It also details which versions are provided by each version of each distribution/operating system.  The intent is that this gives a detailed overview of what's supportable, and what's obsolete/unsupportable, and the varying degrees between these states, to enable us to clearly plan what we will support for the 5.1 (and later) releases.  The support states are certainly something which needs wider discussion; this is just the initial conversion of the spreadsheet to give us a starting point for doing that (I just filled in vaguely sensible values to mostly match the current status quo).

This might not be in the most appropriate place in the manual, it can certainly be moved to e.g. the developer part if that's better.  While it's not intended to mirror the installation instructions, it's clearly got some similarities, so if there's a way to better integrate the information here with what's already in the manuals, I'd be happy with that.

I've checked the HTML and PDF tables, but they could do with double-checking to ensure they are legible and (most importantly!) factually correct.

### DIFF
--- a/omero/sysadmins/version-requirements.txt
+++ b/omero/sysadmins/version-requirements.txt
@@ -2,6 +2,210 @@
 Version requirements
 ********************
 
+Proposal for revised minimum version requirements
+-------------------------------------------------
+
+Minimum software versions
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is a proposal to the wider community to solicit their
+feedback. For each component, we need to decide upon the minumum
+supportable version which is provided by a distribution and/or is
+installable independently. Criteria for what is considered to be
+supportable includes support by its developers upstream and within
+each operating system distribution for the lifetime of the 5.1 release
+(including security support), and also upon our resources allocated to
+CI and testing.  If we aren't actively testing it, we can't claim it's
+supported or functional.
+
+Proposed changes are made for both the forthcoming 5.1 release, and
+also for the following 5.2 release, albeit tentatively at this point,
+in order that sysadmins may plan ahead and ensure that prerequisites
+are in place to ease future upgrades.
+
+As a system administrator, your feedback regarding these proposed
+changes would be most appreciated.  We would like to ensure that we
+support deployment and upgrade of OMERO on the most commonly-used and
+supported Linux and Windows releases, and also on MacOS X, with a
+minimum of effort.  If any of the proposed changes will be problematic
+(for example, the minimum requirement for a given component is not
+possible to meet), please do provide us with the details of your
+system so that we can consider this in future revisions to this
+proposal before the requirements are made final.  Please check the
+minimum Java and PostgreSQL versions in particular.
+
+Note that the following section contains the details which were used
+to create this proposal.
+
+Operating systems
+,,,,,,,,,,,,,,,,,
+
+* Microsoft Windows
+
+  * Proposal: [5.1] 7 (client) / Server 2008R2 (server)
+  * Proposal: [5.2] raise minimum to 8 (client) / Server 2012R2
+    (server) if needed
+  * Rationale: All prior versions prior to this are/will be out of
+    support shortly except Vista (which we don't test and is not
+    widely used). We have no testing of these older systems.
+
+* MacOS X
+
+  * Proposal: [5.1] 10.8
+  * Proposal: [5.2] 10.9
+  * Proposal: Support 10.8 for client only, test server only on
+    10.9/10.10
+  * Rationale: All prior versions no longer have security support and
+    it still leaves 3 versions to support. MacOS is typically suited
+    only to client use, not serious server deployment
+
+* Linux (CentOS/RHEL)
+
+  * Proposal: [5.1] 6.6
+  * Proposal: [5.2] Recommend 7.x over 6.x; 6.x will either remain
+    supported or become unsupported
+  * Rationale: 7 is quite new and not yet widely adopted, and so it's
+    likely many institutions will wish to continue using 6 for the
+    foreseeable future. However, the versions in 6 are getting quite
+    old and this does affect the support for other components (see
+    below)
+
+* Linux (Ubuntu LTS)
+
+  * Proposal: [5.1] 12.04 supported, 14.04 recommended
+  * Proposal: [5.2] 12.04 dropped, 14.04 recommended
+  * Proposal: Only LTS versions are supported and tested. Non-LTS
+    versions following a supported LTS release are likely to work but
+    are not officially supported or tested.
+  * Rationale: Both are newer than CentOS/RHEL6.6, so both could be
+    supported since all version requirements are satisfied. However,
+    14.04 comes with Ice 3.5 and is generally more up to date, and we
+    don't have testing in place for either at present. We do not have
+    the resources to test non-LTS versions, though recent releases are
+    in use by developers.
+
+* Linux (Debian)
+
+  * Proposal: [5.1] 7 recommended, 6 dropped
+  * Proposal: [5.2] 8 recommended, 7 supported
+  * Rationale: 6 is older than Ubuntu 12.04 and only has Ice 3.3;
+    extended support will be present for the 5.1 lifetime however so
+    will be usable if Ice is upgraded, but will not have official
+    support.
+
+Bitness
+,,,,,,,
+
+Currently 32- and 64-bit systems are supported for client and server
+on all platforms with the exception of MacOS X (64-bit only).
+
+* Proposal: [5.1] Drop 32-bit support for server, retain 32-bit
+  support for clients
+* Proposal: [5.2] Drop 32-bit support entirely
+* Rationale: It’s not practical to run a server on a 32-bit system due
+  to its memory requirements, so it’s safe to drop 32-bit support for
+  5.1 given that it’s de-facto 64-bit-only today. There are still a
+  number of 32-bit Windows clients, so support should probably be
+  retained. However, new systems are almost all 64-bit so dropping
+  32-bit support is a possibility for 5.2.
+
+Components
+,,,,,,,,,,
+
+* PostgreSQL
+
+  * Proposal: [5.1] 9.2 recommended, 9.3+ supported, 8.4 dropped, 9.0
+    and 9.1 also dropped
+  * Proposal: [5.2] 9.3 recommended, 9.4+ supported, 9.2 supported or
+    dropped
+  * Rationale: Current releases available for all supported systems
+    from upstream and for CentOS/RHEL6.6 officially via SCL (software
+    collections). At a bare minimum, 8.4 and 9.0 must be dropped (end
+    of security support); dropping 9.1 removes the floating point
+    issues fixed by 9.2+. Due to its age, 8.4 no longer receives
+    adequate testing and so we can’t claim to support it. The same
+    applies to 9.0 and 9.1.
+
+* Python
+
+  * Proposal: [5.1] 2.7 recommended, 2.6 supported, (3.x not
+    supported)
+  * Proposal: [5.2] 2.7 recommended, 2.6 dropped, (3.x not supported)
+  * Rationale: 2.7 is provided by all systems except for
+    CentOS/RHEL6.6, however it is available officially via SCL so
+    could be made the minimum. Note that 3.3 is also available via
+    SCL, so there's potential for a 3.x migration in the 5.2
+    timeframe. Dropping 2.6 would unify python support to a single
+    major version, and remove a number of portability issues; now it's
+    restricted to CentOS/RHEL6, it's in practice much less tested than
+    2.7 since no developer is using it routinely. Removing 2.6 support
+    will be dependent upon internal testing of the SCL python in
+    CentOS.
+
+* GCC
+
+  * Proposal: [5.1] 4.4
+  * Proposal: [5.2] Depends upon CentOS6 and Ubuntu 12.04 support
+    status
+  * Rationale: 4.4 is the CentOS/RHEL6.6 default. It would be
+    desirable to update but is probably not what external users will
+    want given the compatibility concerns. The SCL DeveloperToolset3
+    could bring this up to GCC 4.9.1, which would make the new
+    baseline GCC 4.6 (Ubuntu 12.04 and Travis)
+
+* LLVM/clang
+
+  * Proposal: 3.4 
+  * Rationale: It's the current toolchain on MacOS/FreeBSD and is
+    working well; 3.3 dropped due to no longer being used by current
+    toolchains
+
+* MSVC
+
+  * Proposal: [5.1] 2010
+  * Proposal: [5.2] 2012
+  * Rationale: The minimum (and maximum) MSVC version is determined by
+    the Windows Ice builds provided by ZeroC, and as such is limited
+    by our Ice version requirements. Raising the limit in 5.2 will
+    permit use of newer language features.
+
+* Ice
+
+  * Proposal: [5.1] 3.4 supported, 3.5 recommended, 3.6 likely not
+    supported
+  * Proposal: [5.2] 3.4 dropped, 3.5 supported, 3.6 recommended
+    (subject to testing)
+  * Rationale: 3.4 is no longer supported by ZeroC with 3.5 being
+    current. By the release of 5.2, Ice 3.6 will have been out for
+    some time and should be well tested.
+
+* Java
+
+  * Proposal: [5.1] 7 recommended, 8 supported, 6 dropped
+  * Proposal: [5.2] 7 supported, 8 recommended
+  * Rationale: It's 2 years this Februrary since the last Java 6
+    security update. Java 7 is supported on all supported operating
+    systems above, and in most cases is the default java version on
+    those systems (it’s now 8 on some). This was mentioned recently,
+    and also by the Fiji developers; 5.1 would be an opportune time to
+    drop Java 6. Also note that we no longer actively test with a Java
+    6 JVM; all internal testing is on Java 7 or 8 (Oracle and
+    OpenJDK), and there are some client issues with OpenJDK6
+
+* Apache
+
+  * Proposal: [5.1] 2.2
+  * Rationale: 2.2 is the CentOS/RHEL6.6 and Ubuntu 12.04 version. 2.4
+    support in OMERO isn't perfect at this point, needing some manual
+    tweaking (should be resolved for 5.1)
+
+* nginx
+
+  * Proposal: [5.1] 1.0
+  * Proposal: [5.2] 1.4
+  * Rationale: All versions work; 1.0 is the Ubuntu 12.04 version, 1.4
+    is the Ubuntu 14.04 and Debian 8 version.
+
 Support levels
 --------------
 
@@ -263,8 +467,8 @@ Linux (Ubuntu)
     * - 10.04 (.4) LTS
       - from Apr 2010
       - to Apr 2015 (server only)
-      - Supported
-      - Supported
+      - Dropped
+      - Dropped
       - `Link <https://wiki.ubuntu.com/Releases>`__
     * - 10.10
       - from Oct 2010
@@ -533,7 +737,8 @@ PostgreSQL
       - 
       - 
 
-Distribution support:
+Distribution support
+,,,,,,,,,,,,,,,,,,,,
 
 .. tabularcolumns:: |l|l|L|L|L|l|l|l|
 
@@ -679,7 +884,8 @@ Python
       - Unsupported
       - `Link <https://www.python.org/dev/peps/pep-0478>`__
 
-Distribution support:
+Distribution support
+,,,,,,,,,,,,,,,,,,,,
 
 .. tabularcolumns:: |l|l|L|L|L|l|l|
 
@@ -836,7 +1042,8 @@ GCC
       - Upcoming
       - Upcoming
 
-Distribution support:
+Distribution support
+,,,,,,,,,,,,,,,,,,,,
 
 .. tabularcolumns:: |l|l|L|L|L|l|l|
 
@@ -970,7 +1177,8 @@ LLVM/clang
       - Upcoming
       - Upcoming
 
-Distribution support:
+Distribution support
+,,,,,,,,,,,,,,,,,,,,
 
 .. tabularcolumns:: |l|l|L|L|L|l|l|
 
@@ -1039,6 +1247,66 @@ versions were not much used as a default compiler and are probably not
 worth supporting; 3.4 is used in MacOS X 10.9 and 10.10 and in FreeBSD
 10.1. Until 3.4, no point releases were made for major versions.
 
+Microsoft Visual Studio
+^^^^^^^^^^^^^^^^^^^^^^^
+
+`General overview <http://support2.microsoft.com/lifecycle/search/?sort=PN&alpha=Visual+Studio>`__
+
+.. tabularcolumns:: |l|l|l|l|l|l|
+
+.. list-table::
+    :header-rows: 1
+
+    * - 
+      - 
+      - Upstream
+      - support
+      - 
+      - 
+    * - Version
+      - Release date
+      - (mainline)
+      - (extended)
+      - OME (build)
+      - OME (deploy)
+    * - 2008
+      - from Aug 2008
+      - to Apr 2013
+      - to Apr 2018
+      - Unsupported
+      - Unsupported
+    * - 2010
+      - from Jun 2010
+      - to Jul 2015
+      - to Jul 2020
+      - Supported
+      - Supported
+    * - 2012
+      - from Oct 2012
+      - to Jan 2018
+      - to Jan 2023
+      - Supported
+      - Supported
+    * - 2013
+      - from Jan 2014
+      - to Apr 2019
+      - to Apr 2024
+      - Recommended
+      - Recommended
+    * - 2015
+      - from TBA
+      - to TBA
+      - to TBA
+      - Unsupported
+      - Unsupported
+
+The version to use is largely dependent upon the Ice support for a particular Visual Studio version.
+
+- Ice 3.5 supports 2013, 2012 and 2010
+- Ice 3.4 supports 2010 and 2008
+
+In general later versions have better C++ standards conformance and
+correctness, and should be preferred where possible.
 
 Ice
 ^^^
@@ -1084,7 +1352,8 @@ Ice
       - Upcoming
       -
 
-Distribution support:
+Distribution support
+,,,,,,,,,,,,,,,,,,,,
 
 .. tabularcolumns:: |l|l|L|L|L|l|l|
 
@@ -1101,7 +1370,7 @@ Distribution support:
     * - 3.3
       - N/A
       - N/A
-      - 14.10
+      - 10.04
       - 6.0
       - N/A
       - N/A
@@ -1168,7 +1437,8 @@ Java
       - Supported
       - `Link <http://www.oracle.com/technetwork/java/eol-135779.html>`__
 
-Distribution support:
+Distribution support
+,,,,,,,,,,,,,,,,,,,,
 
 .. tabularcolumns:: |l|l|L|L|L|l|l|
 
@@ -1305,7 +1575,8 @@ OpenGL
       - Unsupported
       - Unsupported
 
-Distribution support:
+Distribution support
+,,,,,,,,,,,,,,,,,,,,
 
 .. tabularcolumns:: |l|l|l|l|l|l|J|l|
 
@@ -1465,14 +1736,15 @@ Apache
       - from Dec 2005
       - TBA
       - N/A
-      - Supported
+      - Recommended
     * - 2.4
       - from Feb 2012
       - TBA
       - N/A
-      - Recommended
+      - Supported
 
-Distribution support:
+Distribution support
+,,,,,,,,,,,,,,,,,,,,
 
 .. tabularcolumns:: |l|l|L|L|L|l|l|
 
@@ -1516,8 +1788,9 @@ Distribution support:
       - 
 
 Apache 2.2 requires mod_fastcgi (from external repository, for
-CentOS6).  Apache 2.4 requires mod_proxy_cfgi (`to be tested
-<https://trello.com/c/ROD6R0L7/180-omero-web-custom-prefix>`__).
+CentOS6).  Apache 2.4 requires mod_proxy_cfgi and isn't perfectly
+supported by default (needs a manual tweak), should be working for
+5.1.
 
 nginx
 ^^^^^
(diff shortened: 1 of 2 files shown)


omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+Version requirements
+********************
+
+Support levels
+--------------
+
+The following sections use the terminology in the table below to
+describe the support status of a given component, as it progresses
+from being new and not supported, to supported and regularly tested,
+and to finally to being old and no longer supported or tested.
```
hflynn: Remove one of these 'to's

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - Supported
+      - Supported
+    * - 10.10
+      - from Oct 2014
+      - Supported
+      - Supported
+      - Recommended
+      - Recommended
+
+Apple don't formally announce end of life for their releases, but 10.6
```
hflynn: don't -> do not

mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.1-staging/sysadmins/version-requirements.html.

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
+********************
+Version requirements
+********************
+
+Support levels
+--------------
+
+The following sections use the terminology in the table below to
+describe the support status of a given component, as it progresses
+from being new and not supported, to supported and regularly tested,
```
mtbc: not sure if "frequently" captures your meaning better than "regularly" (or if either is true!)

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - from Oct 2012
+      - to Jan 2018
+      - to Jan 2023
+      - Supported
+      - Upcoming
+      - `Link <http://support.microsoft.com/lifecycle/?p1=16526>`__
+
+XP is now unsupported upstream, and Server 2003 will be unsupported
+within the timeframe of 5.1 support.  Vista is likely supported but
+usage is minimal and it's untested, hence marked as dropped.  7/Server
```
mtbc: s/it's/it is/

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+    * - 7
+      - from June 2014
+      - to June 2024
+      - Upcoming
+      - Upcoming
+      - `Link <http://wiki.centos.org/FAQ/General#head-fe8a0be91ee3e7dea812e8694491e1dde5b75e6d>`__
+
+Only RHEL and CentOS 6 are supported at present.  Given the long life
+of enterprise releases, I'd suggest we should only support the latest
+one at any given time or else it ties us into very old dependencies;
```
mtbc: That's a good idea. :+1:

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+    * - 15.04
+      - TBA
+      - TBA
+      - Upcoming
+      - Upcoming
+      - `Link <https://wiki.ubuntu.com/Releases>`__
+
+Only the LTS and 13.10 releases are supported.  We may wish to only
+formally support e.g. the last one or two LTS releases (being a bit
+more frequent than CentOS/RHEL; 10.04 has quite old dependencies and
```
mtbc: close )

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - Supported
+      - `Link <https://wiki.ubuntu.com/Releases>`__
+    * - 15.04
+      - TBA
+      - TBA
+      - Upcoming
+      - Upcoming
+      - `Link <https://wiki.ubuntu.com/Releases>`__
+
+Only the LTS and 13.10 releases are supported.  We may wish to only
```
mtbc: "we may wish" sounds more like an internal policy memo than sysadmin docs

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - Supported
+      - Upcoming
+      - `Link <http://support.microsoft.com/lifecycle/?p1=16526>`__
+
+XP is now unsupported upstream, and Server 2003 will be unsupported
+within the timeframe of 5.1 support.  Vista is likely supported but
+usage is minimal and it's untested, hence marked as dropped.  7/Server
+2008 and 8/Server 2012 are both supported and used.  Server 2008
+builds in place; no 2012 builds or testing in place yet.  No regular
+Windows CI deployment and testing in place other than manually.  Some
```
mtbc: +is ?

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - OME (build)
+      - OME (deploy)
+      - Details
+    * - 2.5
+      - from Sep 2006
+      - to Oct 2011
+      - Dropped
+      - Dropped
+      - `Link 1 <https://mail.python.org/pipermail/python-committers/2011-October/001844.html>`__
+        `Link 2 <https://www.python.org/download/releases/2.5/>`__
```
mtbc: lots of these "Link" links, I wonder if we can find better text for these links

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - Unsupported
+      - `Link <http://legacy.python.org/dev/peps/pep-0429/>`__
+    * - 3.5
+      - TBA (probably Sep 2015)
+      - TBA
+      - Unsupported
+      - Unsupported
+      - `Link <https://www.python.org/dev/peps/pep-0478>`__
+
+Distribution support:
```
mtbc: I wonder if these would be better as some kind of sub-heading.

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - `Link <http://wiki.centos.org/FAQ/General#head-fe8a0be91ee3e7dea812e8694491e1dde5b75e6d>`__
+    * - 7
+      - from June 2014
+      - to June 2024
+      - Upcoming
+      - Upcoming
+      - `Link <http://wiki.centos.org/FAQ/General#head-fe8a0be91ee3e7dea812e8694491e1dde5b75e6d>`__
+
+Only RHEL and CentOS 6 are supported at present.  Given the long life
+of enterprise releases, I'd suggest we should only support the latest
```
mtbc: "I'd suggest" looks very internal-memo too.

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+    * - Details
+      - 
+      - `Link <https://apps.fedoraproject.org/packages/postgresql>`__
+      - `Link <http://packages.ubuntu.com/search?keywords=postgresql&searchon=names&suite=all&section=all>`__
+      - `Link <https://packages.debian.org/search?keywords=postgresql&searchon=sourcenames&suite=all&section=all>`__
+      - 
+      - 
+
+The PostgreSQL project provides `packages
+<http://www.postgresql.org/download/>`__ for supported platform.
```
mtbc: platforms plural?

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - `Python 2 <https://packages.debian.org/search?keywords=python2&searchon=sourcenames&suite=all&section=all>`__
+        `Python 3 <https://packages.debian.org/search?keywords=python3&searchon=sourcenames&suite=all&section=all>`__
+      - 
+      - 
+
+At the moment 2.7 support is present upstream for the foreseeable
+future; 3.x versions continue to be released and retired regularly in
+parallel.  The limiting factor will be distribution support for 2.7 as
+major packages are slowly switching to 3.x, and this might leave us in
+a mess if our python module dependencies are no longer available
```
mtbc: also reading rather more as internal docs

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - 
+      - 
+
+Note that clang++ 3.3 is used in MacOS 10.8 and FreeBSD 10; earlier
+versions were not much used as a default compiler and are probably not
+worth supporting; 3.4 is used in MacOS X 10.9 and 10.10 and in FreeBSD
+10.1. Until 3.4, no point releases were made for major versions.
+
+
+ICE
```
mtbc: "Ice"?

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - 
+      - `Link <https://apps.fedoraproject.org/packages/httpd>`__
+      - `Link <http://packages.ubuntu.com/search?keywords=apache&searchon=names&suite=all&section=all>`__
+      - `Link <https://packages.debian.org/search?keywords=apache&searchon=names&suite=all&section=all>`__
+      - 
+      - 
+
+Apache 2.2 requires mod_fastcgi (from external repository, for
+CentOS6).  Apache 2.4 requires mod_proxy_cfgi (`to be tested
+<https://trello.com/c/ROD6R0L7/180-omero-web-custom-prefix>`__).
```
mtbc: OME Trello links will be broken for docs.

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - Description
+    * - Upcoming
+      - unsupported/new
+      - New version not yet regularly tested and not officially supported; may or may not work (use at own risk)
+    * - Recommended
+      - supported/optimal
+      - Current version which is regularly tested, confirmed to work correctly, recommended for optimal performance/experience
+    * - Supported
+      - supported/suboptimal
+      - Older version which is tested, confirmed to work correctly, but may not offer optimal performance/experience
```
mtbc: Might we support versions that are *newer* that the recommended version?

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - Upcoming
+      - `Link <http://support.microsoft.com/lifecycle/?p1=16526>`__
+
+XP is now unsupported upstream, and Server 2003 will be unsupported
+within the timeframe of 5.1 support.  Vista is likely supported but
+usage is minimal and it's untested, hence marked as dropped.  7/Server
+2008 and 8/Server 2012 are both supported and used.  Server 2008
+builds in place; no 2012 builds or testing in place yet.  No regular
+Windows CI deployment and testing in place other than manually.  Some
+manual testing on Windows 7 VMs and machines.
```
mtbc: No version of Windows is recommended for deployment?

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - TBA
+      - Upcoming
+      - Upcoming
+      - `Link <https://wiki.ubuntu.com/Releases>`__
+
+Only the LTS and 13.10 releases are supported.  We may wish to only
+formally support e.g. the last one or two LTS releases (being a bit
+more frequent than CentOS/RHEL; 10.04 has quite old dependencies and
+could be considered for dropping.  No CI testing for any version
+except 10.04 on howe/gretzky; some developer use of 12.04 LTS, 14.04
```
mtbc: Many mentions of CI on this page. Ought there be links to pages within http://www.openmicroscopy.org/site/support/contributing/continuous-integration.html ?

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1606 @@
...
+      - Recommended
+      - `Link <https://www.debian.org/releases/wheezy/>`__
+    * - 8.0
+      - TBA
+      - TBA
+      - Upcoming
+      - Upcoming
+      - `Link <https://www.debian.org/releases/jessie/>`__
+
+Stable releases 7.x and v6.x are supported.  There is no CI testing for
```
mtbc: s/v6.x/6.x/ to match 7.x?

omero/sysadmins/version-requirements.txt
manics: I vote for keeping things simple and recommending just one version (i.e. the latest fully supported version). This also reduces the number of people running older versions when we decide to drop one.

mtbc: It occurred to me that, for instance, the latest one, while it works well enough, might from our point of view have some "regressions" for which we're still figuring workarounds. At present Java 8 is probably such a case.

rleigh-dundee: @manics While that's certainly an ideal, it's not really going to be a practical reality.  Given that each Linux distribution has a specific version (or versions) of each component, it's often the case that one or more will be a bit older or newer than our recommendation, e.g. current stable distributions providing PostgreSQL 9.1 while we might wish to mandate 9.2+, same with python, java, ice etc.

omero/sysadmins/version-requirements.txt
rleigh-dundee: @mtbc It's intentionally not Recommended because this implies we have CI deployment or some other form of regular testing in place.  We didn't when I started the document, but we might now?  Until we've got it running under server2012, that's still marked Upcoming intentionally as well.  There's a card for that.

@kennethgillen What's the current state of w2008/w2012 CI stuff, do you know?  Do we have a daily deployment of the OMERO build on w2008?  Is there a w2012 CI node yet?  Thanks.

omero/sysadmins/version-requirements.txt
rleigh-dundee: That's intentional.  I just wrote that as a placeholder for having a concrete policy, which I hope can be decided on here or as a followup.  Same for the other comments/questions in the text.

omero/sysadmins/version-requirements.txt
rleigh-dundee: This is another placeholder where we need to decide on a proper policy.

omero/sysadmins/version-requirements.txt
rleigh-dundee: We could do, but I don't think there's a huge value in doing so.  It's just referring to the general facts that we have CI builds for CentOS (most OMERO and Bio-Formats jobs), and a smaller number of MacOS and Windows jobs, rather than any specific jobs.  The main point is that there is zero testing of the other platforms (Ubuntu, Debian, etc.).

The main thing I'd like to have in place here are regular or semi-regular builds and tests on all the platforms we claim to have support for.  Docker was supposed to help here for the Linux variants, but we're waiting on that at present.  Relating to our policies, we should probably drop Recommended to Supported if we don't do proper testing on that platform.

omero/sysadmins/version-requirements.txt
rleigh-dundee: The only real purpose of these links is to document where the information came from and provide a hint for where to update it for new releases.  We can certainly make it nicer than "Link", but extended descriptions will screw up the table layouts.

omero/sysadmins/version-requirements.txt
rleigh-dundee: This can certainly be updated.  This is another area where we need to decide on a policy.  In this case it's probably quite simple (support 2.6/2.7 only for now), though there is a question of whether we should keep 2.6 support (which is tied into whether we support multiple CentOS major versions since retaining CentOS6 support forces 2.6 support).

omero/sysadmins/version-requirements.txt
rleigh-dundee: This is another thing to discuss.  If it needs testing internally, we need to do so then we can update this with the proper information.

@manics What's the status of this one?

rleigh-dundee:
I have pushed fixes for all the minor comments.

We could really do with sitting down and going through the data to properly specify what our support policies will be for 5.1 for all the marked sections.  What would be the best way to do this?  Meeting/chat sometime this week?

omero/sysadmins/version-requirements.txt
mtbc: Perhaps "source", where that's appropriate? Just a thought.

mtbc:
I've not yet got around to much helping with https://trello.com/c/E3APkGze/258-versions-of-prerequisites and this stuff does need to be figured out before we release 5.1.0, though I am way behind you in knowing the details, I also have some graphs reimplementation cleanup, and I am in imminent danger of being sucked into rendering settings work. However, I am happy to meet, especially if I can be useful, this week or next. Who should be in the meeting? Is this about putting together a clear, succinct case for proposed prerequisites, to be shared via nitpick or whatever, ready for another meeting at which they may or may not be approved?

hflynn:
Does @joshmoore need to be involved in this discussion? Might be best to schedule a time once he's back on Friday.

hflynn:
Also, can we please avoid individual Trello links in the public documentation? It's not very good practice to point externals at something they don't have permission to view.

rleigh-dundee:
@hflynn Yes.  It's not intended to be merged with the link in, just a point for something which needs looking at to ensure the information is accurate.  It does need discussion prior to being suitable for merging.

hflynn:
:+1:  just wanted to clarify

hflynn:
the build system is also complaining about the https://fedoraproject.org/wiki/Releases#Current_Supported_Releases anchor link - I know it works but given the content is nearly at the top of the page anyway, could you just remove the anchor so I don't have to add it to the ignore list to get the build green?

rleigh-dundee:
Document updated to include https://docs.google.com/document/d/1z6S6fVK6Ls7WhJ0kLOyPApegKiG-F-21kCjVyk4ivnc/edit# as Jason suggested.

I've tidied it a bit more (renamed Link to Reference,or Ref where space is limited).  Are there any outstanding changes to the wording or layout etc.?

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1884 @@
...
+-------------------------------------------------
+
+Minimum software versions
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is a proposal to the wider community to solicit their
+feedback. For each component, we need to decide upon the minumum
+supportable version which is provided by a distribution and/or is
+installable independently. Criteria for what is considered to be
+supportable includes support by its developers upstream and within
```
mtbc: I don't know if a "both" after "support" would make this sentence easier to parse. I certainly had to re-read it a couple of times.

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1884 @@
...
+supported or functional.
+
+Proposed changes are made for both the forthcoming 5.1 release, and
+also for the following 5.2 release, albeit tentatively at this point,
+in order that sysadmins may plan ahead and ensure that prerequisites
+are in place to ease future upgrades.
+
+As a system administrator, your feedback regarding these proposed
+changes would be most appreciated. We would like to ensure that we
+support deployment and upgrade of OMERO on the most commonly-used and
```
mtbc: probably don't need hyphen after "-ly" word

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1884 @@
...
+supported Linux and Windows releases, and also on MacOS X, with a
+minimum of effort. If any of the proposed changes will be problematic
```
mtbc: s/will/would/

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1883 @@
...
+Version requirements
+********************
+
+Proposal for revised minimum version requirements
+-------------------------------------------------
+
+Minimum software versions
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is a proposal to the wider community to solicit their
```
joshmoore: I'd think the PR and/or the emails to ome-users/ome-devel are the proposal and that this document should be the current status of that "version plan" (or whatever), keeping the request for feedback. If that doesn't fit with the more general intent, perhaps under `contributing/` would be better.

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1883 @@
...
+    supported
+  * Proposal: [5.2] 3.4 dropped, 3.5 supported, 3.6 recommended
+    (subject to testing)
+  * Rationale: 3.4 is no longer supported by ZeroC with 3.5 being
+    current. By the release of 5.2, Ice 3.6 will have been out for
+    some time and should be well tested.
+
+* Java
+
+  * Proposal: [5.1] 7 recommended, 8 supported, 6 dropped
```
joshmoore: Copying my comment from the gdoc: was there discussion about ImageJ/FIJI's maintaining Java 6 support?

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1883 @@
...
+The following subsections detail the versions of each dependency
+needed by OMERO which are supported by both its upstream developers
+(for security and general updates) and by OME for OMERO building and
+server and client deployment.
+
+.. note::
+    Versions in brackets are in development distributions and may
+    change without notice.
+
+Package lists
```
joshmoore: This section reads a bit strange. The links are great but it's unclear that/if this is solely what the section is for, or if there will be references made in the rest of the sections.

omero/sysadmins/version-requirements.txt
rleigh-dundee: It was noted on the trello card that the wider implications of this need investigation.  If their decision to maintain Java 6 support effectively locks us into doing the same, then we'll certainly need to retain support for Java 6 and update the proposal here.  We could perhaps then propose to drop it in 5.2 instead.

Can the ImageJ/Fiji base run plugins which are compiled for 1.7?  I'm just wondering to what extent we are tied in.

omero/sysadmins/version-requirements.txt
rleigh-dundee: This is primarily for use in updating the information below and for adding additional packages; it's a link to the package lists for each system we're potentially interested it.  It's not really directly useful for sysadmins.  It could be turned into sphinx comments to keep it out of the doc, but it might possibly be of use to others.  Certainly happy to change it if desired.

omero/sysadmins/version-requirements.txt
joshmoore: Another possible way of integrating this section with the rest of the file is to have:

```
OME (build)	OME (deploy)   OME (next-build)	OME (next-deploy)
```
(with better names, of course) so that for every current version one could see where we are headed.

rleigh-dundee: Certainly happy to reword.  This section was intended to simply contain proposed changes along with a brief rationale, for people to read and comment on.  I wasn't thinking it would be retained for the 5.1 release; after we've reached a decision upon the minimum requirements, we can update the minimum requirements and delete this entirely?

rleigh-dundee: With respect to updating the tables to include the information, it will be difficult to fit it in (for the PDF).  The top section (the proposal) was split out from the tables to keep it specific to just the proposed changes for people to look through without being overwhelmed by all the information.  The detailed tables are the source data which was used for the rationale, and the information can certainly be added there if needed, but that's not going to be a summary of the changes as intended which will be easily read and understood by the people we need feedback from.

omero/sysadmins/version-requirements.txt
rleigh-dundee: Just did a quick check with ImageJ: I converted bioformats to build with 1.7 and tested the resulting loci_tools.jar with openjdk7 and 6.  Works fine with 7, with 6 you get:

```
java.lang.UnsupportedClassVersionError: loci/plugins/LociImporter : Unsupported major.minor version 51.0
 
This plugin requires Java 1.7 or later.
```

Whether we want to do that is debatable.  I've got the latest usage stats and I'll see what the current Bio-Formats Java 6 usage is.  But it's certainly an option we could consider.

mtbc: If we do decide to keep supporting Java 6, we had better find a way to include it in our QA workflow without risking anything from its long lack of security support. Even on Debian 7 I am using Java 7 without resorting to backports, and our CI nodes are generally Java 7.

I wonder if usage stats reveal enough about platforms to allow us to identify what fraction can't easily upgrade.

joshmoore: @mtbc: good point.

@rleigh-dundee :  can I propose we add another classification **"Deprecated"** for all the items which you are proposing in 5.2 to drop but we are keeping support in 5.1? If that makes sense, let's please change Java 6 from "Dropped" to "Deprecated".

manics: Does that imply we need to be damn sure the 5.2 **Recommended** versions work properly in 5.1?

rleigh-dundee: Will do.

@manics That would certainly be preferred where possible.  They certainly need to be expected to work properly for 5.2.  But I would say it's a roadmap which could be subject to change; our plan might need to change if we encounter unexpected problems in a component.

rleigh-dundee: @mtbc One thing I just spotted is that Ice 3.6 only supports Java 6 and 7, so if we want to support Ice 3.6 in 5.2, that implies dropping Java 6 before or at that point.

omero/sysadmins/version-requirements.txt
rleigh-dundee: This will be updated to have the support status for 5.0/5.1/5.2 in the next push.

hflynn:
Sorry, didn't get on to this today, will read it over properly and comment tomorrow

rleigh-dundee:
@joshmoore The latest commit should address most of your comments.  Is the 5.0/5.1/5.2 support in the tables and the updated summary at the start OK structure-wise?  The actual support statuses may still need tweaking; I've added deprecated statuses where we would drop support in the following release.

omero/sysadmins/version-requirements.txt
joshmoore: @manics: yes, that's what it means.

@rleigh-dundee:  not necessarily. It may just mean only supporting a sparse Ice/Java matrix.

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1988 @@
...
+Summary of planned changes for OMERO 5.1 and 5.2
+================================================
+
+Criteria for what is considered to be supportable includes support
+both by its developers upstream and within each operating system
+distribution for the lifetime of the 5.1 release (including security
+support), and also upon our resources allocated to CI and testing. If
+we are not actively testing it, we cannot claim it is supported or
+functional.  Software components must be provided and supported either
+by a distribution or or the original developers.
```
hflynn: You have two 'or's here. Also the 1st sentence is a bit of a mouthful, what about starting "Criteria for what we support includes whether upstream support from developers and operating system distributions will be available for the lifetime of the…" etc, etc

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1988 @@
...
+    * - 7
+      - from June 2014
+      - to June 2024
+      - Unsupported
+      - Supported
+      - Recommended
+      - `Ref <http://wiki.centos.org/FAQ/General#head-fe8a0be91ee3e7dea812e8694491e1dde5b75e6d>`__
+
+Only RHEL and CentOS 6 are supported at present. Given the long life
+of enterprise releases, I'd suggest we should only support the latest
```
hflynn: Can you change this phrasing from 'I'd suggest…' to 'we propose only supporting the latest one at any given time to avoid tying ourselves into very old dependencies...

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1988 @@
...
+================
+
+
+Microsoft Windows
+-----------------
+
+Supports amd64 (64-bit) and i386 / x86 (32-bit) for all supported
+versions; most systems are today running x64 with most hardware being
+x64, but there is still a niche of x86 users, estimated <25% and
+falling globally (OMERO usage stats here?)
```
hflynn: Do we have these stats immediately to hand? If not, can you just remove reference to them here and we'll add them later.

omero/sysadmins/version-requirements.txt
```diff
@@ -0,0 +1,1988 @@
...
+      - `Python 2 <https://packages.debian.org/search?keywords=python2&searchon=sourcenames&suite=all&section=all>`__
+        `Python 3 <https://packages.debian.org/search?keywords=python3&searchon=sourcenames&suite=all&section=all>`__
+      - 
+      - 
+
+At the moment 2.7 support is present upstream for the foreseeable
+future; 3.x versions continue to be released and retired regularly in
+parallel. The limiting factor will be distribution support for 2.7 as
+major packages are slowly switching to 3.x, and this might leave us in
+a mess if our python module dependencies are no longer available
```
hflynn: 'this might leave us in a mess if...' -> 'will cause problems if...'

hflynn:
Few extra nitpicks about wording but otherwise I think we are good

omero/sysadmins/version-requirements.txt
rleigh-dundee: Added some stats, but they are quite crude.  I'm still working on improving this reporting.  What I've added is based on unique users of 5.0 from june-dec 2014.

hflynn:
Changes all look fine to me, thanks! I'll merge once builds are green and it should be live after lunch.

hflynn:
Ok, all green, merging.

hflynn:
--no-rebase

mtbc:
I can't help but wonder if we ought to be saying something about supported web browser versions too. @aleksandra-tarkowska, @will-moore, have we a card or anything about that?

will-moore:
@mtbc We've generally been quite conservative about deprecating browser support. We may not have perfect appearance on IE8, but we do support it and everything else since.

rleigh-dundee:
@will-moore Having a concrete list of what is supported in 5.0, what we plan to support in 5.1 and in 5.2 does give a roadmap for where we are right now and where we will be in the future.  It wouldn't be much work to add such a list to this version requirements file.  That's irrespective of whether we're being conservative or not.

But note the supported operating systems section on this page; the minimum web browser versions should match what's supported here.  e.g. Since our minimum Windows version is 7, the minimum IE versions does not need to include versions in XP/Vista.  Likewise for Safari on MacOS <8. etc.

PR #1059, opened 2015-01-04
Add double and timestamp MATLAB annotations functions to the table

sbesson:
This Pull Request adds the new annotations utility functions and a link to the `WriteData` example to the MATLAB developer page.

### DIFF
--- a/omero/developers/Matlab.txt
+++ b/omero/developers/Matlab.txt
@@ -446,91 +446,27 @@ This is useful when you need the ``Pixels`` intensity.
 Annotations
 -----------
 
-The following table lists all OMERO.matlab functions used to manipulate
-annotations from OMERO:
-
-.. list-table::
-    :header-rows: 1
-    :widths: 10, 15, 15, 15, 15, 15, 5
-
-
-    -   *
-        * Comment
-        * Double
-        * File
-        * Tag
-        * Timestamp
-        * XML
-
-    -   * Get by identifier
-        * :source:`getCommentAnnotations <components/tools/OmeroM/src/annotations/getCommentAnnotations.m>`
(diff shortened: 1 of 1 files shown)


sbesson:
--depends-on https://github.com/openmicroscopy/openmicroscopy/pull/3304

hflynn:
The table is really wide (not that I think there is much we can do about that) but all renders fine on staging. @jburel do you want to review any technical content when you are happy with the code PR?

sbesson:
@hflynn: as briefly mentioned in #3304, this table will still grow (need to add `long`, maybe `boolean`, `term` and `list` as well as `map`).
If we want to preserve the table, I am happy to swap the rows/columns to prevent uncontrolled width expansion. If you think this table does not add much value, I can alternatively link to the source folder and give a couple of examples.

jburel:
@hflynn: i will have a look.

hflynn:
Ah. The table is already too wide for my laptop screen, and swapping the rows/columns is still going to make it wider again. At the point at which you need to scroll, I'm not sure it adds any value.

sbesson:
@hflynn: as discussed, last commit should remove the table which is becoming not manageable.

jburel:
```
To create a file annotations from the content of a ``local_file_path`` and
 attach it to the image ``image_id``::
 
     fileAnnotation = writeFileAnnotation(session, local_file_path);
-    link = linkAnnotation(session, fileAnnotation, 'image', image_id);
```
Since you removed the link section. the text   "attach it to the image ``image_id``" should also be removed. You explain after how to link annotation

sbesson:
Pushed

hflynn:
All looks good to me

sbesson:
--rebased-to #1073

PR #1063, opened 2015-01-07
Update CreateApp docs (rebased onto develop)

will-moore:
This is the same as gh-1055 but rebased onto develop.

---

Updates the web CreateApp page to address a couple of small issues:
- previous instructions specified that the app directory itself was on the `PYTHONPATH` whereas it should be the parent directory that is on the `PYTHONPATH`.
- import statements in the urls.py example don't work with 5.0 OMERO

To test, follow the CreateApp instructions until you have the "Welcome" message at a new app /index.html.


will-moore:
--rebased-from #1055

PR #1066, opened 2015-01-12
Fixing broken anchor link

hflynn:
Existing link was making http://ci.openmicroscopy.org/view/Docs/job/OMERO-5.1-merge-docs/ yellow and when I hunted around, I found the topic was actually on a different page.

cc @aleksandra-tarkowska - if you are happy with this, I'll rebase to dev_5_0 once the build is green.

### DIFF
--- a/common/conf.py
+++ b/common/conf.py
@@ -174,6 +174,7 @@ extlinks = {
     'snapshot' : (cvs_root + '/snapshots/%s', ''),
     'zeroc' : ('http://zeroc.com/%s', ''),
     'zerocdoc' : ('http://doc.zeroc.com/%s', ''),
+    'djangodoc' : ('https://docs.djangoproject.com/en/1.6/%s', ''),
     'doi' : ('http://dx.doi.org/%s', ''),
     'pypi': ('https://pypi.python.org/pypi/%s', '')
     }
--- a/omero/developers/Web/CSRF.txt
+++ b/omero/developers/Web/CSRF.txt
@@ -8,7 +8,7 @@ application in which they are currently authenticated. For more details see
 
 OMERO.web provides easy-to-use protection against Cross Site Request
 Forgeries, for more information see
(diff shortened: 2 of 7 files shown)


aleksandra-tarkowska:
@hflynn I think it should be https://docs.djangoproject.com/en/1.6/topics/templates/#template-inheritance

aleksandra-tarkowska:
Looks like we need to check if we have more links like https://docs.djangoproject.com/en/dev/, they should all be 1.6 rather then dev

sbesson:
```
$ git grep "docs.djangoproject.com/en/dev" | wc
       5      15     608
```

Time to create a `djangodocs` extlink? This will also simplify maintenance the day we bump our Django version.

aleksandra-tarkowska:
@sbesson great idea, what is the best strategy to do it?

hflynn:
I'll do that on this PR, thanks for pointing that out @aleksandra-tarkowska

hflynn:
I left one /dev/ link because it was to the 1.5 notes instead.

hflynn:
Ok, I've got extra '/' in the links, I'll fix that after stand-up (realised after I left yesterday, bah!)

sbesson:
I would use this PR to find and replace all links to the django documentation using your new extlink:

```
$  git grep "docs.djangoproject.com/en" | wc
      22     175    3742
```

hflynn:
Ah, I only grepped for /en/dev/ - I'll fix this up this afternoon

hflynn:
Ok, aside from the 1.5 one already mentioned, that is all of them except those in the config.txt which I'll need to open a PR on the code base tomorrow to fix.

sbesson:
https://ci.openmicroscopy.org/job/OMERO-5.1-merge-docs/ is green. :+1: Anything else on this PR?

hflynn:
I hadn't planned to add anything unless someone spotted a mistake I missed.

aleksandra-tarkowska:
I am good

hflynn:
--rebased-to #1067

PR #1070, opened 2015-01-14
adjust createdb commands to specify Unicode databases

mtbc:
https://trello.com/c/Fv1HZVfs/779-virt-microscope-pg-9-for-testing reveals the need for a UTF8-encoded database for OMERO since 5.1 patch 12.

PostgreSQL 8.4 references can be updated once version prerequisites suggested via http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-January/004982.html are approved, so for now they are left alone.

--no-rebase

### DIFF
--- a/omero/sysadmins/unix/server-install-homebrew.txt
+++ b/omero/sysadmins/unix/server-install-homebrew.txt
@@ -244,7 +244,7 @@ Start the PostgreSQL server.
 
 ::
 
-    $ initdb /usr/local/var/postgres
+    $ initdb -E UTF8 /usr/local/var/postgres
     $ brew services start postgresql
     $ pg_ctl -D /usr/local/var/postgres/ -l /usr/local/var/postgres/server.log start
 
--- a/omero/sysadmins/unix/server-install-homebrew.txt
+++ b/omero/sysadmins/unix/server-install-homebrew.txt
@@ -248,14 +248,15 @@ Start the PostgreSQL server.
     $ brew services start postgresql
     $ pg_ctl -D /usr/local/var/postgres/ -l /usr/local/var/postgres/server.log start
 
-Create a user, a database and add the PL/pgSQL language to your database.
+Create a user, a database and, if using PostgreSQL 8.4, add the PL/pgSQL
+language to your database.
 
 ::
 
     $ createuser -P -D -R -S db_user
     Enter password for new role:       # db_password
     Enter it again:       # db_password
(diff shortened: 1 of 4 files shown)


rleigh-dundee:
Not `UTF-8` (with hyphen)?

omero/sysadmins/unix/server-install-homebrew.txt
```diff
@@ -248,14 +248,15 @@ Start the PostgreSQL server.
...
+Create a user, a database and, if using PostgreSQL 8.4, add the PL/pgSQL
+language to your database.
 
 ::
 
     $ createuser -P -D -R -S db_user
     Enter password for new role:       # db_password
     Enter it again:       # db_password
-    $ createdb -O db_user omero_database
+    $ createdb -E UTF8 -O db_user omero_database
```
sbesson: Could you also open a PR to modify `OMERO-homebrew-install.sh` to include this option and have it tested by the daily job?

mtbc:
@rleigh-dundee: http://www.postgresql.org/docs/8.4/static/multibyte.html#MULTIBYTE-CHARSET-SUPPORTED

omero/sysadmins/unix/server-install-homebrew.txt
mtbc: Good idea: https://github.com/openmicroscopy/openmicroscopy/pull/3345

manics: And whilst you at it...might as well keep https://github.com/manics/omero-install/tree/master/linux in sync

sbesson:
As discussed with @mtbc something we may add to this PR is a single command explaining how to convert the encoding of a PG database before running the upgrade script for 5.1.0 in the server upgrade page.

mtbc:
I don't know what to say for that, but I could accept PRs against this branch from others. @rleigh-dundee makes suggestions on https://trello.com/c/l6DZx1jU/252-5-1-utf-8-db-upgrade but we should wait until @kennethgillen has something well-tested and recommended.

mtbc:
Would it make sense to simply merge this as is? I'm working on model change stuff right now and can open a separate PR for DB upgrade instructions.

sbesson:
No veto to merging this in its current state.

manics:
Shouldn't this be rebased to dev_5_0 since if someone's creating a new DB there's no reason not to make it UTF-8?

PR #1076, opened 2015-01-23
OMERO model object reference

mtbc:
This is a newly created document that, as a developer, I've often wished I had. Staged at https://www.openmicroscopy.org/site/support/omero5-staging/developers/Model/EveryObject.html and fixes http://trac.openmicroscopy.org.uk/ome/ticket/12140.

--rebased-to #1081

### DIFF
--- a/omero/developers/Model.txt
+++ b/omero/developers/Model.txt
@@ -25,18 +25,18 @@ vocabulary has a representation in the :model_doc:`XML specification <ome-xml/>`
 in the database (the data model), and in code. This last representation
 is the object model with which we will concern ourselves here.
 
-Because of its complexity, the object model is generated from a 
+Because of its complexity, the object model is generated from a
 :sourcedir:`central definition <components/model>` using our own
-:sourcedir:`code-generator <components/dsl>`. It relies on no
-libraries and can be used in both the server and the RMI clients.
-The relationships among the objects are enumerated in a hyperlinked
+:sourcedir:`code-generator <components/dsl>`. It relies on no libraries
+and can be used in both the server and the RMI clients. The
+relationships among the objects are enumerated in a hyperlinked
 :doc:`reference document <Model/EveryObject>`.
-:doc:`/developers/server-blitz` uses a :sourcedir:`second
-mapping <components/blitz/resources/templates>` to
-generate |OmeroJava|, |OmeroPy|, and |OmeroCpp| classes, which can be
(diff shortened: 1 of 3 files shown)


sbesson:
How did you generate the every object page?

mtbc:
https://github.com/mtbc/openmicroscopy/commit/4825e5b44beb4802b6fb79771f2648a54d22173f more or less. Can certainly adjust formatting if need be, but it was hard to find anything more concise that Sphinx would actually permit.

mtbc:
One thing I *can* do is to close up the properties section a bit at the cost of having the "used by" links link to the class, not the specific property, if that's preferable.

mtbc:
Pushed an additional commit to show what I meant by the latest comment. Can be undone if preferred.

hflynn:
Having both 'relationships' and 'properties' under each subheading in the side menu makes this very hard to navigate - could we remove the subheadings? The 'relationships' are self-explanatory and we could use a bullet point to mark the properties subsection instead.

hflynn:
Being really nitpicky, I'd prefer 'properties are:' but this looks much better thanks! Can we also have 'Subclasses', Used by' and 'Properties' since these are new sentences? Otherwise, I'm happy with the formatting now.
What is the plan here for the page generation? Can this be merged and then an autogen PR opened @sbesson ?

sbesson:
Definitely planning to have this auto-generated at some point. This won't make sense here since the first commit is so large. Also this will require a top-level PR to be opened by @mtbc.

mtbc:
Will make the suggested changes. :smiley:

My plan is to get this and its rebase past review, then commit to develop the resulting OMERO.server class that generates them, and when the 5.1 model is finalized to submit an update docs PR to include those changes, then we oughtn't need it again until 5.2. Invocation is a bit nasty at the moment and I add the overview paragraph manually, but in the longer term I can work with @sbesson to help him to find some autogen approach that works.

hflynn:
Ok, so this PR can be merged once the final nitpicks are sorted then? @mtbc do you want to make a card on Trello for your plan?

hflynn:
Looks pretty good on staging to me, merging.

mtbc:
Opened #1081 and https://github.com/openmicroscopy/openmicroscopy/pull/3404, and created https://trello.com/c/a1Utd8JL/404-update-everyobject-txt and https://trello.com/c/jfPrRneK/261-autogen-omero-object-reference.

PR #1079, opened 2015-01-26
Add cp/_rm section for 5.0.7

joshmoore:
See: https://github.com/openmicroscopy/openmicroscopy/pull/3379

### DIFF
--- a/omero/sysadmins/in-place-import.txt
+++ b/omero/sysadmins/in-place-import.txt
@@ -295,6 +295,17 @@ Finally, the least favored option is `ln_rm`. It first performs
 a hard-link like `ln`, but once the import is complete it attempts
 to delete the original file. This is currently in testing as
 an option for DropBox but is unlikely to be of use to general users.
+Although this option is more limited than the `upload_rm` option below it
+will be much faster.
+
+"upload_rm" - uploading and deleting
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This option, available from OMERO 5.0.3, is not strictly an in-place option
+but is detailed here for convenience. It first performs a file upload like
+default import, but once the import is complete it attempts to delete the
+original files. It deletes the original files **if and only if** the import
+is successful.
 
 "cp" and "cp_rm" variants
 ^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


ximenesuk:
This looks good but in looking at the built docs I noticed that the, not strictly in-place, transfer option "upload_rm" isn't yet documented so this might be a good PR for that to sit in. The comparison between the "cp" group and the "ln" group is fine but why "cp" over "upload"? That comparison should be stated somewhere hereabouts too.

joshmoore:
Happy to add in `upload_*`. Guess the only real optimization would be performance, though I haven't tested it. Assuming, say, a 1TB file, I'd think a simple `cp` would be much healthier.

joshmoore:
@ximenesuk : copied the minimal blurb from https://github.com/openmicroscopy/ome-documentation/pull/975, but obviously you did much more there. Do we want to go further? /cc @hflynn

ximenesuk:
Looks good to me. A sentence hinting at the benefits of `cp` might head off any questions but if we have no concrete metrics then maybe leave that for now? I'm happy either way.

joshmoore:
I'm inclined to keep it a bit simple for 5.0.7. During the rebase, I'll see if it can be more tightly tied into your other fixes. @hflynn, this is all yours.

hflynn:
Erm, sorry to come at this late but if this isn't actually in-place import stuff, would it not be better documented on https://www.openmicroscopy.org/site/support/omero5-staging/sysadmins/import-scenarios.html or at least signposting further up the page ?

joshmoore:
`cp` and `cp_rm` is definitely in-place. `upload` and `upload_rm` work both via in-place and non-in-place (out-of-place? at-home?)

hflynn:
Ah ok, sorry, I obviously got the wrong end of the stick.

hflynn:
Re-running the merge build now but this is looking fine to me in that case. I'm inclined to agree about leaving the extra stuff on 5.1.0 only since we already decided not to rebase #975

ximenesuk:
Yes, I think in-place may now be a misplaced term! All 5.x imports are effectively in-place in effect, it's just what you do before and after the import that differs!

hflynn:
We do like to pick our terminology!

hflynn:
Looks fine on staging. merging.

joshmoore:
--rebased-to #1087

PR #1077, opened 2015-01-23
fix URL for Lucene Index Toolbox

mtbc:
The URL for Luke has moved.

Staged at https://www.openmicroscopy.org/site/support/omero5.1-staging/developers/Modules/Search.html.

--no-rebase for now

PR #1080, opened 2015-01-27
fix URL for Lucene Index Toolbox (rebased from develop)

mtbc:
The URL for Luke has moved.

Staged at https://www.openmicroscopy.org/site/support/omero5-staging/developers/Modules/Search.html.

--rebased-from #1077


hflynn:
Thanks!

PR #1081, opened 2015-01-27
OMERO model object reference (rebased from dev_5_0)

mtbc:
This is a newly created document that, as a developer, I've often wished I had. Staged at https://www.openmicroscopy.org/site/support/omero5.1-staging/developers/Model/EveryObject.html and fixes http://trac.openmicroscopy.org.uk/ome/ticket/12140.

--rebased-from #1076


sbesson:
Thanks

PR #1091, opened 2015-02-02
Fix link to Homebrew installation script

sbesson:
With the migration of the script to the omero-install repository, the
link on the documentation page needs to be updated

PR #1092, opened 2015-02-02
Fix link to Homebrew installation script (rebased onto dev_5_0)

sbesson:
This is the same as gh-1091 but rebased onto dev_5_0.

----

With the migration of the script to the omero-install repository, the
link on the documentation page needs to be updated

### DIFF
--- a/omero/sysadmins/unix/server-install-homebrew.txt
+++ b/omero/sysadmins/unix/server-install-homebrew.txt
@@ -14,7 +14,7 @@ OMERO.server Mac OS X installation walk-through with Homebrew
 
 The instructions provided here depend on Homebrew 0.9 or later, including
 support for the ``brew tap`` command. These instructions are implemented in a
-series of :source:`automated scripts <https://github.com/ome/omero-install/tree/master/homebrew>`
+series of `automated scripts <https://github.com/ome/omero-install/tree/master/homebrew>`_
 which install OMERO via Homebrew from a fresh configuration.
 
 Prerequisites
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #1091

sbesson:
See https://github.com/ome/omero-install/pull/11

rleigh-dundee:
This also looks fine.  Please merge.

rleigh-dundee:
Actually, the docs merge build has broken link for me: https://github.com/snoopycrimecop/openmicroscopy/blob/merge/dev_5_0/latest/https://github.com/ome/omero-install/tree/master/homebrew  (in http://www.openmicroscopy.org/site/support/omero5-staging/sysadmins/unix/server-install-homebrew.html)

sbesson:
My bad. Pushed a fix.

rleigh-dundee:
Thanks, that's much better.

rleigh-dundee:
Looks good to merge.

Note that this page is quite outdated and could do some work for 5.1:
- OSX/Xcode section describes the old CI setup which was replaced a while back
- "Java..comes as standard on OS X" has not been true since 10.6 and could be removed.
- System vs. homebrew python.  Rather than detailing all the various combinations which are possible, maybe just recommend use of homebrew python and leave it at that?

Most of this is duplication what's already on the main unix installation page, which also covers installing the prerequisites with homebrew.  Could this page be cut down to just how to do the install and delegate the rest of the detail to the main page?

hflynn:
Great suggestion @rleigh-dundee - I'll stick a card on the Docs board for that, if you or Seb want to volunteer, please go ahead :)

PR #1100, opened 2015-02-09
Updating schema link and users what's new for 5.1

hflynn:
See https://trello.com/c/bm6wvqIt/147-data-model-docs-updates

### DIFF
--- a/omero/users/whatsnew.txt
+++ b/omero/users/whatsnew.txt
@@ -1,12 +1,13 @@
 What's new for OMERO 5.1
 ========================
 
-From a user's perspective, OMERO 5.1 introduces units and some other
-stuff that others **will** write about.
+From a user's perspective, OMERO 5.1 introduces support for units and new more
+flexible user-added metadata. Other updates and new features include:
 
-- Moving and deleting data is now faster.
-
-- There is **going to be** a new :help:`Getting Started guide
-  <getting-started-5.html>` for OMERO 5.1.
-
-.. note:: :doc:`demo-server` **is now available for OMERO 5.1.**
+- improved performance for moving and deleting data
+- the ability to copy and paste rendering settings and select between
+  different users' settings in the UI
+- improved functionality of the OMERO ImageJ plugin allowing you to save ROIs
+  and overlays from ImageJ into OMERO
+- the ability to 'import to' specified groups, projects and datasets using
+  OMERO.dropbox or the Command Line Importer 
--- a/omero/users/whatsnew.txt
+++ b/omero/users/whatsnew.txt
@@ -9,5 +9,6 @@ flexible user-added metadata. Other updates and new features include:
   different users' settings in the UI
 - improved functionality of the OMERO ImageJ plugin allowing you to save ROIs
   and overlays from ImageJ into OMERO
-- the ability to 'import to' specified groups, projects and datasets using
-  OMERO.dropbox or the Command Line Importer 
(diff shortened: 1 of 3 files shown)


omero/users/whatsnew.txt
```diff
@@ -1,12 +1,13 @@
...
-stuff that others **will** write about.
-
-- Moving and deleting data is now faster.
-
-- There is **going to be** a new :help:`Getting Started guide
-  <getting-started-5.html>` for OMERO 5.1.
-
-.. note:: :doc:`demo-server` **is now available for OMERO 5.1.**
+From a user's perspective, OMERO 5.1 introduces support for units and new more
+flexible user-added metadata. Other updates and new features include:
```
mtbc: Enclose "more flexible" in commas? Just a thought -- I am not sure if even I think it would help -- leave as-is if you prefer.

omero/users/whatsnew.txt
```diff
@@ -1,12 +1,13 @@
...
+
+- improved performance for moving and deleting data
+- the ability to copy and paste rendering settings and select between
+  different users' settings in the UI
+- improved functionality of the OMERO ImageJ plugin allowing you to save ROIs
+  and overlays from ImageJ into OMERO
+- the ability to 'import to' specified groups, projects and datasets using
+  OMERO.dropbox or the Command Line Importer
```
mtbc: For CLI couldn't we do much of that for 5.0? I may be misunderstanding...

omero/users/whatsnew.txt
hflynn: I'm not sure it adds anything, I don't think the meaning is confused currently

omero/users/whatsnew.txt
hflynn: Ok, I thought there was something new here but maybe I've misunderstood what it is @ximenesuk ?

omero/users/whatsnew.txt
mtbc: Okay. :+1:

omero/users/whatsnew.txt
ximenesuk: At the moment all DropBox imports are to orphaned images and default group. No choice.

The CLI can be used to import to a specific group and target container, but in the case of a container it must already exist and its id must be known.

The intention in 5.1 is to allow that selection of group and container to be (optionally) automatic based on the source path of the original files. Here the containers would be created on demand if they didn't already exist. So, I guess the line needs re-wording a little, something like:

the ability to automatically 'import to' specified groups, projects and datasets from
OMERO.dropbox or the Command Line Importer using information from the file path

but worded a bit better?

joshmoore:
Should we hash out https://github.com/openmicroscopy/openmicroscopy/pull/3495 first?

hflynn:
Yes, this can wait til after m4. And is aimed at users rather than devs anyway.

hflynn:
Actually, given this updates the model version, it could do with going out with m4 if we get the model stuff sorted today. I'll push a fix shortly, we always have the 'not released yet' disclaimer at the top of the page to cover us ;)

hflynn:
How's this @ximenesuk @joshmoore ?

hflynn:
@joshmoore happy for me to merge this for m4 once everything is green?

joshmoore:
LGTM.

hflynn:
--no-rebase

PR #1103, opened 2015-02-12
in upgrade tell sysadmins to first ensure Unicode DB

mtbc:
Staged at https://www.openmicroscopy.org/site/support/omero5.1-staging/sysadmins/server-upgrade.html. Removes some to-5.0-specific text, refers those users to the 5.0 upgrade instructions, and adds some hints about how to ensure a Unicode-encoded OMERO DB.

--no-rebase

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -20,7 +20,7 @@ Upgrade check list
     :local:
     :depth: 1
 
-Check Prerequisities
+Check prerequisities
 ^^^^^^^^^^^^^^^^^^^^
 
 Before starting the upgrade, please ensure that you have obtained all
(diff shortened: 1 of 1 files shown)


omero/sysadmins/server-upgrade.txt
```diff
@@ -146,6 +127,31 @@ Upgrade your database
         earlier server. If upgrading from a |version| server, you do not need
         to upgrade the database.
 
+Ensure Unicode character encoding
+"""""""""""""""""""""""""""""""""
+
+OMERO 5.1 requires a Unicode-encoded database. From :command:`psql`::
+
+  # SELECT datname, pg_encoding_to_char(encoding) FROM pg_database;
```
rleigh-dundee: You can also run `\l` in psql or `psql -l` to get the same information.

rleigh-dundee:
All looks clear and good.  Just noted one place the command might be simplified, but the output in the example is simpler than the suggested changes, so depends on which you prefer.

omero/sysadmins/server-upgrade.txt
mtbc: Thank you, good point. Now pushed commits mentioning that.

rleigh-dundee:
Thanks, looks good.

PR #1105, opened 2015-02-13
Remove release process page from contributing documentation

sbesson:
As most of our release process is using Trello boards and adjustements are made on a template boards, this page is always out-of-sync and not worth the effort of maintaining it.

--no-rebase


hflynn:
Agreed, merging thanks.

PR #1101, opened 2015-02-10
Mention new templates.xml memory settings in troubleshooting section

sbesson:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12527

PR #1107, opened 2015-02-13
Mention new templates.xml memory settings in troubleshooting section (rebased onto dev_5_0)

sbesson:
This is the same as gh-1101 but rebased onto dev_5_0.

---

Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12527


sbesson:
--rebased-from #1101

PR #1111, opened 2015-02-16
Update re-indexing table

sbesson:
Following http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-February/005038.html, the re-indexing times indicated in our docs is out-of-date are largely overestimated.

This PR updates the re-indexing estimated times using:
- data from a background re-indexing of a production database - courtesy of @ehrenfeu 
- data from an off-line re-indexing of a similarly sized production database

Minor nitpicks/corrections are also added to the off-line re-indexing section.

PR #1114, opened 2015-02-18
Update re-indexing table (rebased onto dev_5_0)

sbesson:
This is the same as gh-1111 but rebased onto dev_5_0.

---

Following http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-February/005038.html, the re-indexing times indicated in our docs is out-of-date are largely overestimated.

This PR updates the re-indexing estimated times using:
- data from a background re-indexing of a production database - courtesy of @ehrenfeu 
- data from an off-line re-indexing of a similarly sized production database

Minor nitpicks/corrections are also added to the off-line re-indexing section.

### DIFF
--- a/omero/sysadmins/search.txt
+++ b/omero/sysadmins/search.txt
@@ -203,16 +203,14 @@ servers of various sizes:
   - * Background [1]_
     * 8h
     * 19TB
-    * -Xmx4800m
+    * :option:`-Xmx4800m`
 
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #1111

hflynn:
Renders fine on staging, merging since content has already had a technical review.

PR #1119, opened 2015-02-19
Updating pytables external link

hflynn:
See https://github.com/PyTables/PyTables/issues/420#issuecomment-75062287 - www.pytables.org is currently shutdown. This should make the build green again.

PR #1122, opened 2015-02-23
Updating pytables external link (rebased onto dev_5_0)

sbesson:
This is the same as gh-1119 but rebased onto dev_5_0.

---

See https://github.com/PyTables/PyTables/issues/420#issuecomment-75062287 - www.pytables.org is currently shutdown. This should make the build green again.


sbesson:
--rebased-from #1119

PR #1124, opened 2015-02-24
New version of xslt matrix graghics

qidane:
Expanded to include new schema version

--no-rebase


hflynn:
I assume there is still info to be added to the table at the top of https://www.openmicroscopy.org/site/support/omero5.1-staging/developers/Model/XsltTransformations.html
Quality image looks fine but I wonder if we should try a high resolution version of the big matrix - it's not very readable blown up on my screen.

rleigh-dundee:
Overall, the content looks OK.  With respect to the PNG image, could this be made an SVG so that it's legible at all sizes?  The PDF can be converted to an SVG quite easily, and sphinx can handle it.

The matrix is going to get increasingly large over time, so future updates might need to move to an alternative representation.

hflynn:
@qidane you ok to fix that image before I merge this?

qidane:
I thought we tried .svg before and found it did not work right on several browsers?
The original idea was to use SVG for both HTML and PDF. But that had to be changed to .pdf and .pdf

qidane:
Might be ok now:
http://caniuse.com/#feat=svg-html5

qidane:
Ah, this might have been the problem: "IE9-11 desktop & mobile don't properly scale SVG files. "
http://caniuse.com/#feat=svg

rleigh-dundee:
I think browser support was a problem last time I tried this.  Looking at IE, it looks like it also has support, the only issue is scaling.  We can specify a size in the SVG and/or in the sphinx markup if needed for this browser, but it looks like the feature itself is now present in all browsers.

qidane:
If the svg file it present is seems to be selected by the build in preference to the png file so no code changes were needed. If you wish to go back to the png simply remove the svg file.

rleigh-dundee:
If the PNG file is no longer used, can we remove it from the PR to avoid permanently wasting space in the git repo?

qidane:
How sure are we that the svg file will work for all?

rleigh-dundee:
@qidane We could test it on Windows to be sure.

rleigh-dundee:
Compared with the PNG image, the PDF (and SVG) both have massive borders on all sides which reduces its readability.  Could the borders be removed from both?

Additionally, the other PNG image doesn't have an SVG version; could that one also have a corresponding SVG for the PDF?  This also has an issue with unnecessary borders.

rleigh-dundee:
New image in Firefox:

ie-svg

New image in IE11:

ie-svg2

The physical size is a bit small for it to be clearly legible.  This might be tweakable by setting the image width property or setting the physical size in the original PDF/SVG so that it's rendered at an appropriate scale.

hflynn:
Realistically, I think we need to accept that people will have to zoom in their browser if they actually want to read this. I'll check the pdf but I suspect we can probably leave this as is.

hflynn:
Pdf looks ok zoomed in too, I'm inclined to say we leave this be now. Ok with you @rleigh-dundee ?

rleigh-dundee:
It's OK as it stands, but I think it could be made more legible by increasing the physical size (the LaTeX PDF rescales it to fit so the absolute size doesn't make any difference there, but the browser renders it at the size it specifies).

hflynn:
Do you want to fetch the branch and experiment with sizes until you're happy with it then? Then you can just let @qidane know what to set it as?

qidane:
It just needs a complete redesign as the matrix has got too big. But I will let that happen on another PR after I am gone.

rleigh-dundee:
@hflynn The change should be trivial.  The font size in the PDF/SVG just needs to be the same as in the other smaller table (or slightly smaller if desired).

Doesn't need to be done if not wanted, but it should be a few seconds of work to change and verify since it's the same in all browsers.

hflynn:
Can you just resize it quickly when you're back please @qidane ? Currently the font size is set to 6 and it doesn't resize as you expand your browser.

hflynn:
I think resetting it to 10pt should be fine

hflynn:
Looks good to me, I'll just check the pdf

hflynn:
All good, pdf is still sized to fit page but web version expands with your window to a perfectly readable font size - thanks!

PR #1127, opened 2015-03-04
adjust server config link

mtbc:
People wanting to know more about server configuration are probably better directed to the general sysadmin page than the overview. Staged at http://www.openmicroscopy.org/site/support/omero5.1-staging/developers/Server/Properties.html.

--no-rebase

### DIFF
--- a/omero/developers/Server/Properties.txt
+++ b/omero/developers/Server/Properties.txt
@@ -23,6 +23,9 @@ OMERO.server:
                                          (Name will change to "…default")
       etc/local.properties             - Local overrides for other properties (used by build only)
 
+The most useful of the properties are listed in a :doc:`glossary
+</sysadmins/config>`.
+
 During the build, these files get stored in the ``blitz.jar`` and are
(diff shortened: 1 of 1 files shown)


hflynn:
Agreed. I don't think you need 'index' though, just `/sysadmins/` should work.

Looks like this page maybe needs further updates - I see mention of 4-beta above!

mtbc:
@hflynn: That link doesn't seem to work for me locally, I don't know why not.

hflynn:
Ah sorry, must be only url links rather than the internal doc links that work that way.

mtbc:
Possibly simply remove that first sentence/paragraph altogether?

hflynn:
Yes, I think the whole section can go. Also maybe helpful to add a link to http://www.openmicroscopy.org/site/support/omero/sysadmins/config.html ?

hflynn:
:+1: Looks good, merging now. Can you rebase to dev_5_0 for tidiness sake please?

PR #1129, opened 2015-03-04
WellSample is no longer annotatable

qidane:
Remove WellSample from the list.

----

--no-rebase


hflynn:
Looks fine on staging, merging thanks.

PR #1131, opened 2015-03-04
adjust server config link (rebased from develop)

mtbc:
People wanting to know more about server configuration are probably better directed to the general sysadmin page than the overview. Staged at http://www.openmicroscopy.org/site/support/omero5.0-staging/developers/Server/Properties.html.

--rebased-from #1127


hflynn:
Good to merge once Travis is green, already built on staging.

PR #1132, opened 2015-03-05
Add brief explanation of the difference between DN and other LDAP attributes

dpwrussell:
Some basic clarification which will be useful for non-LDAP gurus trying to use an LDAP service to authenticate and get group information.

PR #1133, opened 2015-03-09
Add brief explanation of the difference between DN and other LDAP attributes (rebased onto dev_5_0)

hflynn:
This is the same as gh-1132 but rebased onto dev_5_0.

---

Some basic clarification which will be useful for non-LDAP gurus trying to use an LDAP service to authenticate and get group information.


hflynn:
--rebased-from #1132

hflynn:
See https://www.openmicroscopy.org/site/support/omero5-staging/sysadmins/server-ldap.html

PR #1134, opened 2015-03-11
Use Sphinx 1.2.3 for Travis build

sbesson:
Similar fix as in https://github.com/openmicroscopy/bioformats/pull/1657. Travis is currently failing because our documentation does not build cleanly with Sphinx 1.3.0


hflynn:
--rebased-to #1135

PR #1136, opened 2015-03-11
update 5.1 patch version to 1

mtbc:
Have the documentation show the correct patch version for the 5.1 release.

This is rather unfortunate, and should perhaps be reverted for 5.2.

--no-rebase

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -170,7 +170,7 @@ in a short time. The example below assumes you are upgrading from a
 .. parsed-literal::
 
     $ cd OMERO.server
-    $ psql -h localhost -U **db_user** **omero_database** < sql/psql/OMERO\ |version|\_\_0/OMERO\ |previousversion|\_\_0.sql
+    $ psql -h localhost -U **db_user** **omero_database** < sql/psql/OMERO\ |version|\_\_1/OMERO\ |previousversion|\_\_0.sql
     Password for user **db_user**:
     ...
     ...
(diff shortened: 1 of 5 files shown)


manics:
If it'll make maintenance easier you could add a `dbpatch` variable to `omero/conf.py`

hflynn:
That sounds like a good idea

mtbc:
Reading `omero/conf.py` I realize I don't know how to do that. (Presumably we need both `previousdbpatch` and `dbpatch`.) Would you be able to alter it, then I'll adjust the actual documentation pages?

hflynn:
Would we need previous as well? I'd assume just add `dbpatch` so you only have to update it in the config if it changes. @manics are you going to try this out or shall I?

manics:
New proposal.... just get rid of those args https://trello.com/c/8PPlaBmP/456-remove-compulsory-version-patch-args-from-db-script

mtbc:
For things like,
```
$ psql -h localhost -U **db_user** **omero_database** < sql/psql/OMERO\ |version|\_\_1/OMERO\ |previousversion|\_\_0.sql
```
this time there will still be a 0 with the previous version, but for version=5.2 it will be a 1 with previousversion=5.1.

manics:
Note `omero db script` output has now changed: https://github.com/openmicroscopy/openmicroscopy/pull/3615

hflynn:
@mtbc can you update this PR to reflect the changes above please?

mtbc:
Do we have the `conf.py` changes yet?

hflynn:
Maybe I misunderstood, i thought @manics had removed the need for the version change

mtbc:
We still need it for the part where one runs the SQL upgrade script.

manics:
I'm working on the web docs at the moment, after that I'll attempt to make the conf change here if no-one else gets to it first.

mtbc:
Now apparently will need to adjust `sysadmins/cli/db.txt`.

mtbc:
@manics will take all this over once his `conf.py` changes are ready to go.

PR #1137, opened 2015-03-11
Add a note to double check OMERO.web config

ctrueden:
When upgrading our OMERO server from 5.0.0 to 5.0.8, I had an issue
where OMERO.web did not display things properly afterward. The problem
was that although we use a "future-proof" symlink, pointing the 'omero'
folder to whatever the current version is (e.g.,
OMERO.server-5.0.8-ice34-b60), the "bin/omero web config" configuration
stanza generator used the hardcoded version-specific folder name, which
I failed to adjust when I did the 5.0.0 install. So in short, my Apache
configuration was outdated.

Just another pitfall which might be worth mentioning in the upgrade
documentation, since there is already a handy checklist.

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -48,6 +48,15 @@ The passwords and logins used here are examples. Please consult the
 sure to replace the values of **db_user** and **omero_database** with the
 actual database user and database name for your installation.
 
+OMERO.web configuration
+^^^^^^^^^^^^^^^^^^^^^^^
+
+If you generated configuration stanzas using `bin/omero web config` which
+enable OMERO.web via Apache or Nginx, they may include hard-code links to
+your previous version of OMERO. We recommend using a future-proof symlink
+if possible, so that these stanzas do not need updating with each OMERO
+server upgrade. See also the :doc:`unix/install-web` page.
+
 Web plugin updates
 ^^^^^^^^^^^^^^^^^^
 OMERO.web plugins are very closely integrated into the webclient. For this


ctrueden:
@hflynn I don't feel strongly about the actual writeup there, if you prefer different language there. Just wanted to put the idea out there, since we ran into this issue.

manics:
@ctrueden There's an unresolved Trello card for this issue https://trello.com/c/DQ79JLgo/340-check-whether-symlinks-are-kept-or-dereferenced-in-omero-web-config
If you can think of a way to improve things let us know.

ctrueden:
Thanks @manics, interesting. Maybe use `os.getenv('PWD')` as [this SO answer](http://stackoverflow.com/a/1542818) suggests?

hflynn:
this will need to web deployment page link updated to show the split between unix & Windows but otherwise reads and formats fine. @manics shall I merge this while the technical issue is on-going? Any fix wouldn't apply to older installations anyway would it?

manics:
It might not be necessary for Windows.

hflynn:
Good to merge then @manics ?

mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.1-staging/sysadmins/server-upgrade.html#omero-web-configuration.

mtbc:
A couple of suggestions provided in https://github.com/ctrueden/ome-documentation/pull/1.

mtbc:
@hflynn: Now looks fine to me. @aleksandra-tarkowska: Makes good sense to you, too?

joshmoore:
@aleksandra-tarkowska's headed out. This works for me (if that counts).

mtbc:
Sounds good then. :+1:

hflynn:
Thanks all!

hflynn:
--rebased-to #1142

PR #1142, opened 2015-03-13
Add a note to double check OMERO.web config (rebased onto dev_5_0)

hflynn:
This is the same as gh-1137 but rebased onto dev_5_0.

---

When upgrading our OMERO server from 5.0.0 to 5.0.8, I had an issue
where OMERO.web did not display things properly afterward. The problem
was that although we use a "future-proof" symlink, pointing the 'omero'
folder to whatever the current version is (e.g.,
OMERO.server-5.0.8-ice34-b60), the "bin/omero web config" configuration
stanza generator used the hardcoded version-specific folder name, which
I failed to adjust when I did the 5.0.0 install. So in short, my Apache
configuration was outdated.

Just another pitfall which might be worth mentioning in the upgrade
documentation, since there is already a handy checklist.

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -71,7 +71,16 @@ show nothing.
 which has been significantly stream-lined. See :ref:`search-reindexing`
 for more information.
 
-Web Plugin Updates
+OMERO.web configuration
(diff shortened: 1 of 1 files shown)


hflynn:
--rebased-from #1137

manics:
Looks good to me

PR #1143, opened 2015-03-16
Webgateway 51

will-moore:
This updates the WebGateway page and the Python page with 5.1 changes.

WebGateway changes are quite small, simply describing thumbnails support of z/t section from https://github.com/openmicroscopy/openmicroscopy/pull/3285 and adding tiles info to the imgData json.

Python changes are the combination of updates to examples in https://github.com/openmicroscopy/openmicroscopy/pull/3606 and https://github.com/openmicroscopy/openmicroscopy/pull/3616.

Also there are a lot of small changes caused by the flake8 fixes to examples, since the Python.txt is largely code-generated from examples by

```
$ cd examples/Training/
$ python markup.py
```

--no-rebase


jburel:
* 'rwrw--': 'READ-WRITE'}  # Not exposed in 4.4.0 clients => Remove 4.4.0
 * Map Annotation: 
Maybe add the same key twice to indicate that the key does not have to be unique
 * colour => color

hflynn:
What about the Map Annotation comment? Is this good to merge now @jburel @will-moore ?

will-moore:
Sorry - yes, I wasn't sure if here was the right place to explain the characteristics of the Map Annotation - Just code examples for how to use them. Also we're using a list in the code, and users will probably have seen the behaviour of Map Annotations in the clients first.

PR #1146, opened 2015-03-17
Extend CLI import documentation

sbesson:
- Use program/option markup
- Add description of `--debug` and `--skip` arguments

--no-rebase

### DIFF
--- a/omero/users/command-line-interface.txt
+++ b/omero/users/command-line-interface.txt
@@ -175,6 +175,9 @@ The options available to the import command are:
 
         $ bin/omero import images_folder --debug WARN
 
+Other advanced import options are described in the :ref:`inplace_import`
+section.
+
 Manage tags
 -----------
 
--- a/omero/users/command-line-interface.txt
+++ b/omero/users/command-line-interface.txt
@@ -38,11 +38,13 @@ corresponding to your system from the :downloads:`OMERO downloads <>` page.
     an existing OMERO instance.
 
 Once the server is downloaded, the |CLI| is located under the :file:`bin/`
-directory::
+directory:
+
+.. parsed-literal::
 
     $ cd OMERO.server
     $ bin/omero -h
-    OMERO Python Shell. Version 4.4.5-ice33
+    OMERO Python Shell. Version |release|-ice35
     Type "help" for more information, "quit" or Ctrl-D to exit
     omero>
 
--- a/omero/users/command-line-interface.txt
+++ b/omero/users/command-line-interface.txt
@@ -148,17 +148,24 @@ Some of the options available to the import command are:
 
         $ bin/omero import --depth 1 images_folder
 
-.. option:: --skip [SKIP [SKIP]...]
+.. option:: --skip SKIP
 
-    Specify optional import steps to skip. Multiple steps can be skipped and
-    multiple arguments can be supplied. Available options that can be skipped
-    are currently:
+    Specify optional step to skip during import.
+
+    The import of very large datasets like High-Contrent Screening data or
+    SPIM data can be time and resource consuming noth at the client and at the
+    server level. This option allows to disable some non-critical steps and
+    allow faster import of such datasets. The caveat associated with its usage
+    is that some elements are no longer generated at import time and may either
+    be re-generated during client access (e.g. thumbnails) or may need manual
+    intervention to be generated (e.g. minmax).
+
+    Available options that can be skipped are currently:
 
     all
         Skip all optional steps described below
     checksum
-        Skip all steps calculation checksums either on files after transfer
-        or on pixels
+        Skip checksum calculation either on files after transfer or on pixels
     minmax
         Skip computation of the minima and maxima pixel values
     thumbnails
@@ -169,6 +176,10 @@ Some of the options available to the import command are:
     Example of usage::
 
         $ bin/omero import large_image --skip all
+        $ bin/omero import large_image --skip minmax
+
+    Multiple import steps can be skipped by supplying multiple arguments::
+
         $ bin/omero import large_image --skip checksum --skip minmax
 
 .. option:: --debug DEBUG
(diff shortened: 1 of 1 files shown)


omero/users/command-line-interface.txt
```diff
@@ -162,15 +198,15 @@ To create a new tag annotation, use the :omerocmd:`tag create` command::
...
+    Please enter a name for this tag set: my_tag_set
+    Please enter a description for this tag set: description of my_tag_set
 
 For both tags and tag sets, the name and description can be passed using the
 :option:`--name` and :option:`--desc` options::
 
-	$ bin/omero tag create --name my_tag --desc 'description of my_tag'
-	$ bin/omero tag createset --tag 1259 1260 tag_id_2 --name my_tag_set --desc 'description of my_tag_set'
+    $ bin/omero tag create --name my_tag --desc 'description of my_tag'
+    $ bin/omero tag createset --tag 1259 1260 tag_id_2 --name my_tag_set --desc 'description of my_tag_set'
```
manics: `tag_id_2` ?

omero/users/command-line-interface.txt
```diff
@@ -127,20 +127,56 @@ To import a file image.tiff, use::
 
     $ bin/omero import image.tiff
 
-Many options can be passed to the :omerocmd:`import`. They can be listed using
-the :option:`-h` option::
+The options available to the import command are:
```
ximenesuk: Some of the options ?

manics:
Since you're modifying this page you might as well fix a few typos:

- https://github.com/openmicroscopy/ome-documentation/pull/1146/files#diff-ae91cc23e02b919798cb9420c37c0e05R16 `Most of commands`
- https://github.com/openmicroscopy/ome-documentation/pull/1146/files#diff-ae91cc23e02b919798cb9420c37c0e05R44 Remove the `-h`, and if you feel like it update the version string

Other than that looks good.

omero/users/command-line-interface.txt
```diff
@@ -127,20 +127,56 @@ To import a file image.tiff, use::
...
+
+.. option:: --depth DEPTH
+
+    Set the number of directories to scan down for files (default: 4)::
+
+        $ bin/omero import --depth 1 images_folder
+
+.. option:: --skip [SKIP [SKIP]...]
+
+    Specify optional import steps to skip. Multiple steps can be skipped and
```
ximenesuk: I think we need to say why you'd want to use this option, ie import speed, maybe others. And maybe wat that is costing you. Especially as this page is aimed at users who may have little idea what these steps are.

omero/users/command-line-interface.txt
```diff
@@ -127,20 +127,56 @@ To import a file image.tiff, use::
...
+    multiple arguments can be supplied. Available options that can be skipped
+    are currently:
+
+    all
+        Skip all optional steps described below
+    checksum
+        Skip all steps calculation checksums either on files after transfer
```
ximenesuk: I'm not sure what this line is saying. is 'all steps' meant to be here?

omero/users/command-line-interface.txt
```diff
@@ -127,20 +127,56 @@ To import a file image.tiff, use::
...
+        Skip computation of the minima and maxima pixel values
+    thumbnails
+        Skip generation of thumbnails
+    upgrade
+        Skip upgrade check for Bio-Formats
+
+    Example of usage::
+
+        $ bin/omero import large_image --skip all
+        $ bin/omero import large_image --skip checksum --skip minmax
```
ximenesuk: The `--skip SKIP1 --skip SKIP2` example isn't supported by the option given above which implies `--skip SKIP1 SKIP2`

sbesson:
Commits pushed. Note that 934a4300840018b746a74201d78e3dc289e02a94 attempts to add an more thorough description of the large datasets import problem /cc @hflynn

hflynn:
Erm, there's a separate page for CLI importing at https://www.openmicroscopy.org/site/support/omero5.1-staging/users/command-line-import.html
It would probably be better to update that and reference it from this page rather than starting another big import section here.
Sorry I didn't spot that earlier.

sbesson:
The page you linked is about the Java-only command-line importer (which is used by the import plugin of the Python CLI). And some options described by this PR only apply to the Python CLI. Maybe we need to decide of a single page for all command-line import tools. But from my point-of-view we are largely moving towards exposing the Python CLI as the entry point for all command-line actions.

hflynn:
Oh. That's not confusing at all :( Can we explain that somewhere?

sbesson:
Yes maybe we should use this opportunity to reorganize the command-line section. A proposal would be something like

```
users/command-line-interface/index.txt
users/command-line-interface/installing.txt
users/command-line-interface/import.txt
users/command-line-interface/sessions.txt
users/command-line-interface/tag.txt
```

where the `import.txt` page would discuss the Python import but also mention the standalone Java command-line importer on the same page, preventing duplication. Also it would allow to address other CLI doc tickets.

The major impact of this is that it would "break" links to `users/command-line-interface.txt` /cc @ximenesuk @joshmoore

hflynn:
If we're going to break things, now is the time to do it. I'd vote for users/cli/index etc rather than the full command-line-interface though.

sbesson:
Fair enough. Closing this then. I will open a restructuration-only PR this evening then we can review individual sections.

manics:
@sbesson Advanced warning- if you're planning to modify `omero/autogen` as part of this I've also got some pending changes https://github.com/manics/ome-documentation/compare/sys_dev_whatsnew...manics:sys_dev_whatsnew_autogen

sbesson:
Thanks for the heads-up. No I plan to touch the static Sphinx content only as part of my PR. But it is nice to see you plan to push to `omero/downloads/cli` as it matches @hflynn's suggestion .

PR #1144, opened 2015-03-17
Small fixes and improvements to top_links docs

will-moore:
Updated the top_links docs to not use "set" command since this will remove default links.
Also improved layout (indentation) and description of new 'attrs' that can be added in 5.1.

--rebased-to #1154

PR #1154, opened 2015-03-19
Use 'apppend' for top_links and fix indentation

will-moore:
This is most of #1144 rebased on dev_5_0.

Updated the top_links docs to not use "set" command since this will remove any existing links.
Also improved layout (indentation).

--rebased-from #1144


hflynn:
@will-moore you need to remove the `:property:` mark-up, those links don't exist on this branch.

PR #1158, opened 2015-03-20
Autogenerate sysadmin linux walkthrough, admin ports, db script

manics:
Note requires CI changes due to changes to `omero/autogen_docs`, in particular the addition of `ome/omero-install`

Use `omero/autogen` to:

- Copy some of the ome/omero-install Linux walkthrough files
- Generate `omero admin ports -h` text
- Guess the current and previous DB versions
- Generate the example `omero db script` output
  - Note this changes the CLI invocation to use the `--password` flag instead of prompting for the root password (makes it easier to autogenerate the example)
- Remove mentions of plpgsql (applied to postgres 8.4)
- Link to the Linux cron job in the web install

--no-rebase


omero/downloads/cli/db-script-example.txt
```diff
@@ -0,0 +1,6 @@
++ cd /home/omero/OMERO.server
```
sbesson: Arguably this line could be skipped

omero/sysadmins/cli/db.txt
```diff
@@ -6,9 +6,9 @@ Rather than try to provide the functionality of a RDBM tool like ``psql``, the
...
-    $ bin/omero db script OMERO4 0 secretpassword
-    Using OMERO4 for version
-    Using 0 for patch
-    Using password from commandline
-    Saving to /omero/OMERO4__0.sql
-    $ psql omero < OMERO4__0.sql
\ No newline at end of file
+    $ bin/omero db script --password secretpassword
+
+.. literalinclude:: /downloads/cli/db-script-example.txt
```
sbesson: The `db script` command is repeated in the `db-script-example.txt`.

sbesson:
Overall makes sense, the biggest issue is the output of the `db script` command which is likely affected by the fact the script is run within a CI environment. Some comments inline. Another thought is that `common/conf_autogen.py` is OMERO-specific and could be moved to `omero/conf_autogen.py`

Otherwise, the cross-repository doc integration builds fine.

manics:
Fixed

PR #1148, opened 2015-03-18
updating developers/web doc

aleksandra-tarkowska:
This PR update some sample code to the Django standards

To test it, for example create simple form and use ```render``` on HTTP POST. You should not get 403 as ```RequestContext``` is forced and ```{% csrf_token %}``` will be used.

cc @will-moore 
--no-rebase

PR #1159, opened 2015-03-20
updating developers/web doc (rebased onto dev_5_0)

atarkowska:
This is the same as gh-1148 but rebased onto dev_5_0.

---

This PR update some sample code to the Django standards

To test it, for example create simple form and use `render` on HTTP POST. You should not get 403 as `RequestContext` is forced and `{% csrf_token %}` will be used.

cc @will-moore


aleksandra-tarkowska:
--rebased-from #1148

hflynn:
Matches develop content which has already had technical review and looks fine on staging - merging.

PR #1161, opened 2015-03-20
Render_response

aleksandra-tarkowska:
Doc update following comment https://github.com/openmicroscopy/openmicroscopy/pull/3629#issuecomment-84020368

cc: @chris-allan @will-moore

PR #1164, opened 2015-03-23
Render_response (rebased onto dev_5_0)

hflynn:
This is the same as gh-1162 but rebased onto dev_5_0.

---

Replacement for #1161 with fixes ready for m5


hflynn:
--rebased-from #1162

PR #1175, opened 2015-03-26
Ignore pyc files

sbesson:
Related to the failure of https://ci.openmicroscopy.org/job/OMERO-5.1-release/36/. /cc @joshmoore 

Dangling `pyc` files (`omero/conf_autogen.pyc`) broke the `git describe` logic. This commit is making sure these files are ignored.

--no-rebase

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 venv
 .DS_Store
+*.pyc


manics:
Why don't the other python files create .pycs?

sbesson:
Not sure. Only thing I can think of is that maybe `conf.py` is treated specially by Sphinx?

manics:
Looks good to me

PR #1178, opened 2015-03-27
Web python docs tweaks

will-moore:
Some minor docs tweaks from reading through docs (Web & Python mostly).

Biggest change is to the instructions for running Python code samples, following @sbesson's config changes. Other changes are just minor fixes.

--no-rebase

### DIFF
--- a/omero/developers/Python.txt
+++ b/omero/developers/Python.txt
@@ -39,11 +39,13 @@ All the code examples below can be found at
 :sourcedir:`examples/Training/python`.
 
 
-Start by downloading the first example below: 
-:source:`examples/Training/python/Connect_To_OMERO.py` and edit the
-USERNAME and PASSWORD variables according to your log-in.
+If you want to run the examples, you will need to download and
+configure them to connect to your own server. E.g. ``HOST = "localhost"``
+You can edit ``HOST, PORT, USERNAME and PASSWORD`` in
+the ``Parse_OMERO_Properties.py`` file and these values will be imported
+into the other scripts.
 
-Then you can run the example:
+Then you can run the scripts:
 
 ::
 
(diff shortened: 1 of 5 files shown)


will-moore:
cc @aleksandra-tarkowska You wanna have a look at these changes today?

omero/developers/Python.txt
```diff
@@ -39,11 +39,13 @@ All the code examples below can be found at
...
 
 
-Start by downloading the first example below: 
-:source:`examples/Training/python/Connect_To_OMERO.py` and edit the
-USERNAME and PASSWORD variables according to your log-in.
+If you want to run the examples, you will need to download and
+configure them to connect to your own server. E.g. ``HOST = "localhost"``
+You can edit ``HOST, PORT, USERNAME and PASSWORD`` in
+the ``Parse_OMERO_Properties.py`` file and these values will be imported
+into the other scripts.
```
sbesson: Definitely sensible. I am not sure if we also want to document the alternative approach: create/edit an `ice.config` file with `omero.host`, `omero.user`... properly set, point the `ICE_CONFIG` environment variable it and run the examples.

will-moore: I certainly don't want to add that at this point on the page, since this is for "beginners" and should be as simple as possible. Also I'm not sure if that approach has many advantages for most users?

aleksandra-tarkowska:
looks good to me @will-moore

omero/developers/Python.txt
aleksandra-tarkowska: I have never used that in development. That is more for running tests right?

will-moore: Yes, from https://github.com/openmicroscopy/openmicroscopy/pull/3621

sbesson: I can see it either way. For power developers, I would argue `ice.config` files should be presented as the default way in order to create connections as opposed to hard-coding credentials in scripts. But for training purposes, telling people to modify the top-level script is certainly fine.

hflynn:
Will merge once it's built on staging on Monday.

PR #1182, opened 2015-03-30
homebrew: Update for current homebrew and supported platforms

rleigh-dundee:
- Delete information for MacOS 10.7 and earlier, since it's no longer supported by us or homebrew
- Use new Xcode and Java install instructions
- Use current Homebrew dependency versions
- Recommend use of Homebrew Python over system Python
- Use current Homebrew wiki links
- Use Python 2.7 only (2.6 no longer provided by MacOS X or homebrew)
- Set environment at appropriate places in the script so it's correct in the places which depend upon it (`PATH` and `PYTHONPATH`)
- Drop `/usr/local/lib/node_modules` from the user `PATH`; it's not part of any of our other installation instructions and appears to be redundant (used in the past?)
- Reword PostgreSQL installation steps
- Drop old Homebrew problems which are no longer applicable
- Don't invoke an executable shell script with a shebang using bash; just invoke it directly

--no-rebase

### DIFF
--- a/omero/sysadmins/unix/server-install-homebrew.txt
+++ b/omero/sysadmins/unix/server-install-homebrew.txt
@@ -52,7 +52,7 @@ running::
 Homebrew
 ^^^^^^^^
 
-.. _`Homebrew wiki`: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Installation.md#installation
+.. _`Homebrew wiki`: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Installation.md
 
 Follow the installation instructions on the `Homebrew wiki`_. All the
 requirements for OMERO will be installed under :file:`/usr/local`.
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
@joshmoore @sbesson I'm not too familiar with this having never actually used it.  If you could possibly look it over to make sure there's nothing been removed which was still needed that would be great.

sbesson:
Had a quick scan at the proposed changes and did not spot any crucial information removed. Thanks for the cleanup.

omero/sysadmins/unix/server-install-homebrew.txt
```diff
@@ -209,70 +212,45 @@ If you encounter problems with the installation script, please take a look at
...
+#. Create a new database cluster::
 
-::
+    $ initdb -E UTF8 /usr/local/var/postgres
 
-    $ brew tap gapple/services
+#. Start the PostgreSQL server.
 
-::
+  - If you wish to use ``brew services`` to start postgresql, you will
```
hflynn: These bullet points are giving the stupid plone highlighting in staging, although I'm not sure whether you can manage to get bullet points within a numbered section otherwise… I'll have a play in a bit

hflynn: Looking at the structure of points further down though, I think you could just remove the bullet points and have the sentences and literal blocks without them.

hflynn:
Aside from the plone annoyance, this looks fine on staging and reads well. The merge build is complaining about https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Installation.md#installation though, and although it works fine for me, I don't think we actually really need the #installation bit at the end so might be easier just to remove that (the build gets stroppy about functional anchors sometimes but it does make spotting actual errors tougher if we leave them in).

rleigh-dundee:
@hflynn Should be fixed now.

hflynn:
Looks good thanks, I'll merge once everything is green.

PR #1183, opened 2015-03-31
Add paragraph about memo file invalidation post-upgrade

sbesson:
This notice should warn sysadmins of the memo file invalidation/re-generation post upgrade as noticed by @dominikl during testing. 

--no-rebase


hflynn:
:+1:

PR #1187, opened 2015-04-06
Upgrade minor fix

sbesson:
See https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=7778

Check:
- there is no more 4.4 mention on the ugprade page
- the note about using `SQL Shell` describes the workflow corresponding to the forum thread

--no-rebase


manics:
Good to merge

PR #1191, opened 2015-04-16
Doc Fix: Omero Java bindings

bramalingam:
Fixes all the bugs/deprecated methods part of the documentation.

### DIFF
--- a/omero/developers/Java.txt
+++ b/omero/developers/Java.txt
@@ -749,7 +749,7 @@ In the following example, we create an image and delete it.
     img.setAcquisitionDate(omero.rtypes.rtime(1000000));
     img = (Image) entryUnencrypted.getUpdateService().saveAndReturnObject(img);
 
-    HashMap<String, List<Long>> targetObject = new HashMap<String, List<Long>>();
+    java.util.Map<String, List<Long>> targetObject = new HashMap<String, List<Long>>();
     List<Long> idlist = new ArrayList<Long>();
     idlist.add(img.getId().getValue());
     targetObject.put("Image", idlist);
--- a/omero/developers/Java.txt
+++ b/omero/developers/Java.txt
@@ -749,8 +749,8 @@ In the following example, we create an image and delete it.
     img.setAcquisitionDate(omero.rtypes.rtime(1000000));
     img = (Image) entryUnencrypted.getUpdateService().saveAndReturnObject(img);
 
-    Hashtable targetObject = new Hashtable();
-    ArrayList<Long> idlist = new ArrayList();
+    HashMap<String, List<Long>> targetObject = new HashMap<String, List<Long>>();
+    List<Long> idlist = new ArrayList<Long>();
     idlist.add(img.getId().getValue());
     targetObject.put("Image", idlist);
     Delete2 deleteCommands = new Delete2(targetObject, null, false);
@@ -911,4 +911,4 @@ please see |OmeroClients|.
 --------------
 
 .. seealso::
-	ZeroC_, |OmeroGrid|, :ref:`build#OmeroTools`, |OmeroApi|
+    ZeroC_, |OmeroGrid|, :ref:`build#OmeroTools`, |OmeroApi|
--- a/omero/developers/Java.txt
+++ b/omero/developers/Java.txt
@@ -438,7 +438,7 @@ created:
     originalFile.setName(omero.rtypes.rstring(name));
     originalFile.setPath(omero.rtypes.rstring(path));
     originalFile.setSize(omero.rtypes.rlong(file.length()));
-    originalFile.setSha1(omero.rtypes.rstring(generatedSha1));
+    originalFile.setHash(omero.rtypes.rstring(generatedSha1));
     originalFile.setMimetype(omero.rtypes.rstring(fileMimeType)); // or "application/octet-stream"
     // now we save the originalFile object
     originalFile = (OriginalFile) iUpdate.saveAndReturnObject(originalFile);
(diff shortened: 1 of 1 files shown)


omero/developers/Java.txt
```diff
@@ -745,18 +749,18 @@ In the following example, we create an image and delete it.
     img.setAcquisitionDate(omero.rtypes.rtime(1000000));
     img = (Image) entryUnencrypted.getUpdateService().saveAndReturnObject(img);
 
-    DeleteCommand[] cmds = new DeleteCommand[1];
-    //Command to delete the image.
-    cmds[0] = new DeleteCommand("/Image", img.getId().getValue(), null);
-    DeleteHandlePrx handle = entryUnencrypted.getDeleteService().queueDelete(cmds);
+    Hashtable targetObject = new Hashtable();
```
mtbc: a `HashMap` suffices, this is single-threaded code

omero/developers/Java.txt
```diff
@@ -745,18 +749,18 @@ In the following example, we create an image and delete it.
...
+    ArrayList<Long> idlist = new ArrayList();
```
rleigh-dundee: `List<Long> idlist`

rleigh-dundee:
I'm afraid I'm not sufficiently familiar with the OMERO API to review this properly, so you might want someone else to take a look at it.

As a general point, it would be nice to have these examples tested as part of a normal build.  For example, in bioformats look at the examples in `docs/sphinx/developers/cpp/examples` used in `docs/sphinx/developers/cpp/tutorial.txt`.  These are built as part of a regular build, so they can never get out of date.  For the examples in ome-documentation, these could be moved out of `Java.txt` into separate .java files which can be built and run, and they could be built from the ome-documentation submodule as part of a regular build in openmicroscopy.git, and tested against a running server as part of the integration tests.  This would ensure they are always matching the current API and work correctly.

bramalingam:
@rleigh-dundee : I have done that already, waiting for @jburel to get back, but a few of these are run on the daily build (Java examples) now..! Would be good to have them all running as you have suggested.

rleigh-dundee:
I'm not sure how that can be the case when the Java.txt file is not compilable (since it's not Java source).  If you look at the examples I referred to, the examples are not embedded in the Sphinx markup, they are sourced from it and what goes into the documentation is built and tested.  This doesn't need doing for this PR, but it is a defect in the code samples in the docs.

bramalingam:
Oh sorry , what I meant was- I have the source code as a .java file and I have tested it.I was gonna wait for @jburel to take a decision , on how we can move forward on that regard.

sbesson:
@rleigh-dundee: the logic behind these examples should be stored under `examples/Training` in the openmicroscopy repository and is run via https://ci.openmicroscopy.org/job/OMERO-5.1-merge-training/.

The complex question is the one of the synchronization between the code and the docs. As it stands, there are 2 strategies:
- the Python code samples are copied into the documentation as code snippets (via https://github.com/openmicroscopy/openmicroscopy/blob/develop/examples/Training/markup.py) /cc @will-moore. This is similar to what BF C++ is doing.
- the MATLAB page is making use of source code links to point at the full example files

I can see valid points for both approaches (code samples embedding vs links). It is also hard to know whether there should be a general policy or if this is a case-by-case decision. /cc @hflynn

hflynn:
My preference would be code snippets for shortish examples and source code links where the example  is long and/or doesn't need to be read in conjunction with the doc page for the doc page to make sense.

hflynn:
(so case-by-case)

rleigh-dundee:
@hflynn Sphinx lets you do literalincludes of ranges of lines (or the entire) source files, so they do appear directly on the doc page.  But the important point is that they are actually proper source files you can build and run, which means the documentation examples can't get out of date without it being noticed, and the code samples are directly testable.

I don't think it should be case-by-case; it should be done for every example to ensure they are correct.

hflynn:
Sorry, I meant case-by-case on code samples in proper source files and then included in the docs via the autogen job when they are updated vs just linking to code source. I agree we should have code examples pasted from the source that need to be manually updated separately.

rleigh-dundee:
@hflynn There's no real need to copy the examples here.  They could be present in ome-documentation for use by the docs, but also built from openmicroscopy since it can use the ome-documentation submodule to get them.  This gets us both direct inclusion in the docs, plus the ability to link to the whole source file in the docs using `:download:`, and also being built and run as part of the standard build and integration tests.

joshmoore:
Ok, guys. Let's move this to voice. There are too many pieces here and we need to make sure this remains maintainable. Can we finish up @bramalingam's PR (2 comments, @bramalingam, that need addressing) and handle this elsewhere?

hflynn:
Sorry, I think Roger and I are actually saying the same thing anyway, I'm just failing to explain my point in his language!

rleigh-dundee:
Just to be clear about what I'm suggesting, for anyone who didn't look at the examples:

- This is a sample source file https://github.com/openmicroscopy/bioformats/blob/df26d52d34fe79589a79ff013ef0dc8f7a1bc0f7/docs/sphinx/developers/cpp/examples/model-io.cpp (note the `-start` and `-end` comments)
- This is the Sphinx code using it https://github.com/openmicroscopy/bioformats/blob/df26d52d34fe79589a79ff013ef0dc8f7a1bc0f7/docs/sphinx/developers/cpp/tutorial.txt#L228
- This is the final result: http://www.openmicroscopy.org/site/support/bio-formats5.1/developers/cpp/tutorial.html#ome-xml-data-model-objects

So the small code examples in the text are sourced from real source files, which can be build and tested independently of their use in the documentation, so the reader is guaranteed that the examples are correct and functional.

This is doable with ome-documentation example and openmicroscopy with a little reorganisation.

joshmoore:
All agreed, let's just not do this here. @hflynn : any final comments here?

hflynn:
Nope, if Marks' happy with the fixes I'll merge once we're green. Did you want this released live?

joshmoore:
Barring the follow-on discussion re: snippets, think everyone is happy with this. Thanks, @bramalingam. @hflynn : probably no get this live, but also no barrier to doing so. Enjoy your weekend, all.

hflynn:
Will merge once everything is green, thanks.

hflynn:
--no-rebase

PR #1197, opened 2015-04-27
travis: Switch from saucy to trusty

rleigh-dundee:
saucy has been removed from the mirrors and was no longer supported.

Note the reordering of the graphviz install to use precise;
for some reason the trusty version fails to install due to
some (unrevealed) dependency problem installing libgvc6.

--------

Testing: travis should be green.


sbesson:
Thanks. merging.

PR #1201, opened 2015-04-28
known limitations warning for DB client encoding

hflynn:
See https://trello.com/c/CVUflk8g/490-encoding-for-db-init-upgrade-script
Documents how to update upgrade script for DB until we get it fixed properly

### DIFF
--- a/omero/sysadmins/limitations.txt
+++ b/omero/sysadmins/limitations.txt
@@ -26,6 +26,29 @@ correctly. OpenJDK and OracleJDK versions 6 and 7 are also supported.
 Windows OS issues
 -----------------
 
+Failure response on import for new databases
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you have import failures with an error such as::
+
+   java.lang.RuntimeException: Failure response on import!
+   Category: ::omero::grid::ImportRequest
+   Name: import-request-failure
+   Parameters: {stacktrace=java.lang.RuntimeException: omero.ValidationException
+       serverStackTrace = "ome.conditions.ValidationException: could not insert: [ome.model.core.Pixels];
+
+when trying to import data to a new database (or possibly one just upgraded to
+5.1), it may be an issue with the database initialization. If you are working
+with a fresh install, the easiest fix would be to drop the OMERO database and
+re-create it. If you already deleted the initialization script
+(OMERO5.1__0.sql), recreate it by running ``bin/omero db script``. Edit the
+script to add ``SET CLIENT_ENCODING TO 'UTF8';`` as the first line of the
+script. Connect to the database via psql and run the script with ``\i [path to
+script]OMERO5.1__0.sql``.
+
+Alternatively, switch the language in pgadmin and then recreate the DB and DB
+User.
+
 UTF-8 character limitations in passwords
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/omero/sysadmins/limitations.txt
+++ b/omero/sysadmins/limitations.txt
@@ -35,19 +35,25 @@ If you have import failures with an error such as::
    Category: ::omero::grid::ImportRequest
    Name: import-request-failure
    Parameters: {stacktrace=java.lang.RuntimeException: omero.ValidationException
-       serverStackTrace = "ome.conditions.ValidationException: could not insert: [ome.model.core.Pixels];
+       serverStackTrace = "ome.conditions.ValidationException: could not
+       insert: [ome.model.core.Pixels]; SQL [insert into pixels (creation_id,
+       external_id, group_id, owner_id, permissions, update_id,
+       dimensionOrder, methodology, physicalSizeXUnit, physicalSizeX,
+       physicalSizeYUnit, physicalSizeY, physicalSizeZUnit, physicalSizeZ,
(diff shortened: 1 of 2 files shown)


joshmoore:
I'd think we need some text around what's actually happening, namely that the language encoding during database initialization leads to corrupt unicode characters so that any import containing those unicode characters (like μ) will fail with the above exception. cc: @dominikl

manics:
`Parameters: {stacktrace=java.lang.RuntimeException: omero.ValidationException
 serverStackTrace = "ome.conditions.ValidationException: could not insert: [ome.model.core.Pixels];`

sounds like a fairly generic error message, can you add something about checking whether it is a charset problem?

omero/sysadmins/limitations.txt
```diff
@@ -26,6 +26,29 @@ correctly. OpenJDK and OracleJDK versions 6 and 7 are also supported.
...
+If you have import failures with an error such as::
+
+   java.lang.RuntimeException: Failure response on import!
+   Category: ::omero::grid::ImportRequest
+   Name: import-request-failure
+   Parameters: {stacktrace=java.lang.RuntimeException: omero.ValidationException
+       serverStackTrace = "ome.conditions.ValidationException: could not insert: [ome.model.core.Pixels];
+
+when trying to import data to a new database (or possibly one just upgraded to
+5.1), it may be an issue with the database initialization. If you are working
```
rleigh-dundee: Maybe be a bit more specific here about what the "issue" is.  If it's e.g. due to using the wrong encoding for the database it should state that rather than leaving the reader guessing!

dominikl: The database encoding is correct. It's the psql shell which picks up the wrong encoding (from the windows locale) when executing the initialization/upgrade sql scripts.

hflynn:
Hopefully better now

omero/sysadmins/limitations.txt
```diff
@@ -26,6 +26,35 @@ correctly. OpenJDK and OracleJDK versions 6 and 7 are also supported.
...
+       ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)];
+       nested exception is org.hibernate.exception.DataException: could not
+       insert: [ome.model.core.Pixels]
+
+when trying to import data to a new database (or possibly one just upgraded to
+5.1), it may be an issue with the language encoding during database
+initialization leading to corrupt unicode characters, such that any import
+containing these characters (e.g. µ) will fail with the above exception. If
+you are working with a fresh install, the easiest fix would be to drop the
+OMERO database and re-create it. Run ``SET CLIENT_ENCODING TO 'UTF8';`` in the
```
rleigh-dundee: Is this bit correct?  You should have run `bin/omero db script` before running the psql shell, and then source the script with `\i` *after* setting the client encoding?

hflynn: Ah sorry, I have obviously misunderstood @dominikl 's comment on devteam

rleigh-dundee: Have we looked at setting PGCLIENTENCODING in the environment as an alternative; it might make the instructions simpler since you won't have to run psql interactively.

dominikl:
Sorry for the confusion. Actually I'd say, it's not really a 'known limitation' or a bug. It would just need an update of the installation and upgrade documentation for windows.
http://www.openmicroscopy.org/site/support/omero5.1/sysadmins/windows/server-installation.html
 Where it says "2. Execute the following to populate your database..." 
```
omero=> \i C:/OMERO.server/OMERO5.1__1.sql
...
...
omero=> \q
```
This should be 
```
omero=> SET client_encoding = 'UTF8';
omero=> \i C:/OMERO.server/OMERO5.1__1.sql
...
...
omero=> \q
```

hflynn:
Ok, pushed change to installation page - @joshmoore do you want me to remove the limitation or update it to make clear people need that step added to their database creation?

omero/sysadmins/limitations.txt
```diff
@@ -26,6 +26,35 @@ correctly. OpenJDK and OracleJDK versions 6 and 7 are also supported.
...
+       pixelsType, relatedTo, sha1, significantBits, sizeC, sizeT, sizeX,
+       sizeY, sizeZ, timeIncrementUnit, timeIncrement, version, waveIncrement,
+       waveStart, image, image_index, id) values (?, ?, ?, ?, ?, ?, ?, ?, ?,
+       ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)];
+       nested exception is org.hibernate.exception.DataException: could not
+       insert: [ome.model.core.Pixels]
+
+when trying to import data to a new database (or possibly one just upgraded to
+5.1), it may be an issue with the language encoding during database
+initialization leading to corrupt unicode characters, such that any import
```
mtbc: Unicode

hflynn:
Ok, should be good to go now, in discussion with @joshmoore decided to leave upgrade page for now as we would just be creating noise that probably doesn't apply to many people at all. Can revisit that if we get further reports of this issue.

joshmoore:
:thumbsup: Thanks, @hflynn. We'll try to get a better handle on this incl. `PGCLIENTENCODING` etc. for 5.1.2. Merging.

PR #1190, opened 2015-04-15
Public repo docs

hflynn:
Updated public repo documentation, now moved to sysadmin docs.

Will require https://www.openmicroscopy.org/site/support/omero5.1/developers/Web/PublicData.html to be redirected to /sysadmins/public.html when released.

Primary review:
 * https://www.openmicroscopy.org/site/support/omero5.1-staging/sysadmins/public.html

PR #1202, opened 2015-04-28
fix minor typos

mtbc:
Trivial fixes, but if I do them as I see them, they'll mostly end up done!


hflynn:
Hi Mark, can you remove the change to the public data page? I already fixed that in #1190

mtbc:
Ah, thank you; done.

hflynn:
thanks, will merge once green

PR #1208, opened 2015-05-08
CLI envvars 5.1.2

sbesson:
This PR documents the feature additions in https://github.com/openmicroscopy/openmicroscopy/pull/3789

- use `OMERO_TMPDIR` rather than `OMERO_TEMPDIR` across the documentation
- mention `OMERO_USERDIR` (for alternate sessions storage directory) under `omero/cli/sessions`
- mention and describe `bin/omero help env` in the overview page
- add a table listing all the client directories and their environment variables

### DIFF
--- a/omero/sysadmins/system-requirements.txt
+++ b/omero/sysadmins/system-requirements.txt
@@ -75,7 +75,7 @@ the user's HOME directory in :file:`$HOME/omero/tmp`,
 :file:`$HOME\\omero\\sessions`). If your home
 directory is stored on a network, possibly NFS mounted (or similar),
 then these temporary files are being written and read over the
-network. This can slow access down.  The :envvar:`OMERO_TEMPDIR`
+network. This can slow access down.  The :envvar:`OMERO_TMPDIR`
 environment variable may used to set the directory used for temporary
 files, for example on fast local storage.
 
--- a/omero/sysadmins/system-requirements.txt
+++ b/omero/sysadmins/system-requirements.txt
@@ -64,20 +64,46 @@ The recommended client specification is:
 
 Large imports may require 4GB RAM.
 
+.. _client_directories:
+
 Client configuration
 ^^^^^^^^^^^^^^^^^^^^
 
 When performing some operations the clients make use of temporary file
-storage and log directories. By default these files are stored below
-the user's HOME directory in :file:`$HOME/omero/tmp`,
-:file:`$HOME/omero/log`, and :file:`$HOME/omero/sessions`
-(resp. :file:`$HOME\\omero\\tmp`, :file:`$HOME\\omero\\log`, and
-:file:`$HOME\\omero\\sessions`). If your home
-directory is stored on a network, possibly NFS mounted (or similar),
-then these temporary files are being written and read over the
-network. This can slow access down.  The :envvar:`OMERO_TMPDIR`
-environment variable may used to set the directory used for temporary
-files, for example on fast local storage.
+storage and log directories. The table below indicates the default values for
+each directory and the environment variables allowing to set their locations:
+
+.. list-table::
+  :header-rows: 1
+
+  - * Directory
+    * Environment variable
+    * Default (Linux)
+    * Default (Windows)
+  - * OMERO user directory
+    * :envvar:`OMERO_USERDIR`
+    * :file:`$HOME/omero`
+    * :file:`%HOMEPATH%\\omero`
+  - * Temporary files
+    * :envvar:`OMERO_TMPDIR`
+    * :file:`$HOME/omero/tmp`
+    * :file:`%HOMEPATH%\\omero\\tmp`
+  - * Local sessions
+    * :envvar:`OMERO_SESSIONDIR`
+    * :file:`$HOME/omero/sessions`
+    * :file:`%HOMEPATH%\\omero\\sessions`
+  - * Log files
+    *
+    * :file:`$HOME/omero/log`
+    * :file:`%HOMEPATH%\\omero\log`
+
+
+If your home directory is stored on a network, possibly NFS mounted (or similar), then these temporary files are being written and read over the
+network. This can slow access down.
+
+.. seealso::
+   :ref:`client_performance`
+       Troubleshooting section about client performance issues on NFS
 
 Software
 --------
--- a/omero/sysadmins/system-requirements.txt
+++ b/omero/sysadmins/system-requirements.txt
@@ -71,7 +71,7 @@ Client configuration
 
 When performing some operations the clients make use of temporary file
 storage and log directories. The table below indicates the default values for
-each directory and the environment variables allowing to set their locations:
+each directory and the environment variables for overriding their locations:
 
 .. list-table::
   :header-rows: 1
--- a/omero/sysadmins/system-requirements.txt
+++ b/omero/sysadmins/system-requirements.txt
@@ -76,10 +76,10 @@ each directory and the environment variables allowing to set their locations:
 .. list-table::
   :header-rows: 1
 
-  - * Directory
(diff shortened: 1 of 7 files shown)


omero/sysadmins/system-requirements.txt
```diff
@@ -75,7 +75,7 @@ the user's HOME directory in :file:`$HOME/omero/tmp`,
 :file:`$HOME\\omero\\sessions`). If your home
```
manics: RFE: If an example Windows path is included it should also use the appropriate windows env-var (which appears to be either `%HOMEPATH%` or `%APPDATA%` depending on the system). Probably easier to just remove it.

manics: Alternatively since this info is also in http://www.openmicroscopy.org/site/support/omero5.1-staging/sysadmins/troubleshooting.html#performance-issues-with-the-clients you could remove the duplication and add a link instead.

manics:
Otherwise good to merge.

sbesson:
Last two commits introduce a table to improve the clarity about the various client directories, their default locations and the way they can be customized. Do you know under which conditions `APPDATA` is used as the default Windows location?

omero/sysadmins/troubleshooting.txt
```diff
@@ -331,6 +331,8 @@ more details. There are a few known possibilities:
    (:doc:`UNIX instructions <unix/install-web>` or :doc:`Windows instructions 
    <windows/install-web>`).
 
+.. _client_performance:
+
 Performance issues with the clients
```
rleigh-codelibre: Maybe change to "Troubleshooting performance ..." since it makes the section and its cross-references read as though the clients have performance issues!

omero/sysadmins/system-requirements.txt
```diff
@@ -64,20 +64,48 @@ The recommended client specification is:
...
+  :header-rows: 1
+
+  - * Client directory
+    * Environment variable
+    * Default location (UNIX)
+    * Default location (Windows)
+  - * OMERO user directory
+    * :envvar:`OMERO_USERDIR`
+    * :file:`$HOME/omero`
+    * :file:`%HOMEPATH%\\omero`
```
rleigh-codelibre: Why not `%HOME%` to match the UNIX case?

omero/sysadmins/system-requirements.txt
```diff
@@ -64,20 +64,48 @@ The recommended client specification is:
...
-:file:`$HOME/omero/log`, and :file:`$HOME/omero/sessions`
-(resp. :file:`$HOME\\omero\\tmp`, :file:`$HOME\\omero\\log`, and
-:file:`$HOME\\omero\\sessions`). If your home
-directory is stored on a network, possibly NFS mounted (or similar),
-then these temporary files are being written and read over the
-network. This can slow access down.  The :envvar:`OMERO_TEMPDIR`
-environment variable may used to set the directory used for temporary
-files, for example on fast local storage.
+storage and log directories. The table below indicates the default values for
+each directory and the environment variables allowing to set their locations:
```
rleigh-codelibre: Maybe change `allowing to set` to `for overriding` ?

omero/sysadmins/system-requirements.txt
```diff
@@ -64,20 +64,48 @@ The recommended client specification is:
...
+
+.. list-table::
```
rleigh-codelibre: Needs the latex tabular columns?; I see strange vertical alignment of column two.  May be the font, but it doesn't affect the :file: usage in the following columns.

omero/sysadmins/troubleshooting.txt
```diff
@@ -344,18 +346,19 @@ three things to check:
...
-process is running as. As the server makes heavier use of these folders than
-the clients, if the OMERO.server user's home(~) is stored on a network drive
-then slow performance can occur.
+the temporary and log folders of the user the server process is running as. As
+the server makes heavier use of these folders than the clients, if the
+OMERO.server user directory is stored on a network drive then slow performance
+can occur.
 
-To resolve this, you can define an :envvar:`OMERO_TEMPDIR` environment
+To resolve this, you can define an :envvar:`OMERO_TMPDIR` environment
```
rleigh-codelibre: "you can define an :envvar:`OMERO_TMPDIR` environment variable pointing to" would be less woolly and more direct as "define the :envvar:`OMERO_TMPDIR` evironment variable to point to"

omero/sysadmins/troubleshooting.txt
```diff
@@ -344,18 +346,19 @@ three things to check:
...
 variable pointing to a temporary directory located on the local file system
-(e.g. `/tmp/`).
+(e.g. :file:`/tmp/omero_tmp`).
```
rleigh-codelibre: Is the `_tmp` really required here?  Is it not implied that it's temporary by the fact it's localed in `/tmp/`?

rleigh-codelibre:
Generally looks good, some minor nits and suggestions made.

omero/sysadmins/system-requirements.txt
sbesson: This change was addressing https://github.com/openmicroscopy/ome-documentation/pull/1208#discussion_r30024171

sbesson:
Fixes pushed.

omero/sysadmins/system-requirements.txt
rleigh-dundee: Is it correct though?  I see a single use of HOMEPATH in etc/env.bat, but I can't see any obvious use of this batch file, making me wonder if it's used at all.
Also, see
```
components/tools/OmeroPy/src/omero/util/__init__.py:        homeprop = shell.SHGetFolderPath(0, shellcon.CSIDL_APPDATA, 0, 0)
```

Is it using APPDATA instead by default?

Note APPDATA is the "roaming profile" which is possibly not local, and is synced back to a server over the network even if local.  We should be using "%APPDATA%\Local" to ensure it's local and stays local, I think.

manics:
> Do you know under which conditions APPDATA is used as the default Windows location?

https://github.com/sbesson/openmicroscopy/blob/46229e660ce282bfbc45d346c6d5466d96d95678/components/tools/OmeroPy/src/omero/util/__init__.py#L842 suggests it's always AppData.

However Insight seems to use `%HOMEPATH%\omero`, so maybe it's a Java/Python inconsistency?

omero/sysadmins/system-requirements.txt
manics: Could argue it both ways, if large files are being stored you wouldn't want to sync it for performance reasons, but for storing preferences (e.g. last login) it makes sense to use the roaming profile.

sbesson:
There are definitely clients differences. I think Insight is using `System.getProperty('user.dir')` by default. As pointed out above, `APPDATA` is used by default by the CLI. 

I would suggest the next body of work will be to extend this table to
- give an exhaustive list of the default locations for the different clients 
- list the scope of usage of the various environment variables (currently Python (CLI) only)
From there we can decide what needs to be unified/modified. Happy to do this here or in a separate PR.

omero/sysadmins/system-requirements.txt
rleigh-dundee: You're right, I was thinking server rather than client.  For client it definitely makes sense to sync it back, so long as it's not gigabytes of useless logs...  For the configuration, definitely agreed.

manics:
Definitely a separate PR, and perhaps unify all the clients?

rleigh-dundee:
@sbesson Yes, probably best left for another PR; since what we have here is correct, just incomplete for some clients (and we don't specify which).  I've added https://trello.com/b/4TjIW0Bl/devx-board for discussion.

Looks good to merge.

PR #1210, opened 2015-05-14
Recommend chmod 700 OMERO.server directory

manics:
- https://trello.com/c/xDMnFnPd/498-fix-omero-web-var-permissions
- https://github.com/openmicroscopy/openmicroscopy/pull/3818

`etc` and `var` contain sensitive data, and on shared machines should be protected.

### DIFF
--- a/omero/sysadmins/unix/server-installation.txt
+++ b/omero/sysadmins/unix/server-installation.txt
@@ -650,6 +650,13 @@ Configuration
          Creating var/registry
          No descriptor given. Using etc/grid/default.xml
 
+-   If multiple users have access to the system running OMERO you should
+    restrict access to the :file:`OMERO.server/etc` and
+    :file:`OMERO.server/var` directories, for example by changing the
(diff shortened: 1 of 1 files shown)


joshmoore:
Is it possible/appropriate to do what's suggested here in, say, https://github.com/ome/omero-install/blob/f8fc523c1bedf2d4ce853a3a6ea22d868f16c7df/linux/system_setup.sh#L4 ?

manics:
Yes, I'll take out the chmod $HOME and chmod the server directory. Are there any other security issues worth mentioning (bearing in mind the walkthrough needs to be as simple as possible)?

manics:
Doh. The OMERO.web static files are in `lib/python/omeroweb`

joshmoore:
@manics : I would assume that's feasible. Worst-case, we add a simple README.txt to `var/`.

joshmoore:
Reads fine & is a good start before we try to go too far in warning new users.

PR #1212, opened 2015-05-19
server-installation: fix Homebrew package names

ctrueden:
The Python-based Homebrew package names do not have a `python/` prefix (anymore?). The `postgresql` package installs the latest, currently 9.4.

### DIFF
--- a/omero/sysadmins/unix/server-installation.txt
+++ b/omero/sysadmins/unix/server-installation.txt
@@ -68,6 +68,11 @@ Fedora and Scientific Linux.
 
   where *package* is the package name to install.
 
+* For homebrew you need to enable the "python" and "science" taps::
+
+    $ brew tap homebrew/python
+    $ brew tap homebrew/science
+
 .. _server-requirements:
 
 Overview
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
The changes look fine and are good to merge.

Note that this requires running "brew tap homebrew/python" first, and possibly also "brew tap homebrew/science" which we don't describe at the top of the page where homebrew is first mentioned.

ctrueden:
Indeed, I do have `homebrew/python` tapped. If you don't tap `homebrew/python` first, then presumably the `python/pillow` etc. syntax is correct? Too bad that `python/pillow` doesn't work when it is tapped! So there is no universally working command...

rleigh-dundee:
If it would be possible to add around line 70, something like:

```
* For homebrew you need to enable the "python" and "science" taps::

    $ brew tap homebrew/python
    $ brew tap homebrew/science
```

I think that would cover it and make everything internally consistent.

sbesson:
From the CI/doc point of view, we are:
- requiring to tap `homebrew/science`  (https://www.openmicroscopy.org/site/support/omero5.1/sysadmins/unix/server-install-homebrew.html#homebrew) mostly for `hdf5`
- using `pip` to install the Python dependencies although `homebrew/python` + brew install is a fine (but untested daily) alternative

ctrueden:
@rleigh-dundee Good idea; I added your patch as 568970f7e60a8d76060793e1cb4b1a946cdaef14.

rleigh-dundee:
All looks good here.  Please merge.

PR #1151, opened 2015-03-18
CLI restructuring

sbesson:
See https://github.com/openmicroscopy/ome-documentation/pull/1146#issuecomment-83057548 for the genesis of this PR.

This PR does not modify existing content but re-organizes the structure of the CLI documentation. Main (breaking) changes are the following:
- the landing pages for each section of the CLI documentation are `users/cli/index.tx`, `sysadmins/cli/index.txt` and `developers/cli/index.txt`
- the large users and sysadmin CLI sections are split in separate pages (roughly per command) with a mini-toc in the landing page
- the Java command-line-importer section is moved as a subsection of the CLI import page
- the sysadmin Command line interface landing page is moved to the top-level of the sysadmin section under the installation sections

To review this PR, check everything builds as expected and that the new layout makes sense. Content for each page should be reviewed separately and is outside the scope of this PR.

--no-rebase

PR #1213, opened 2015-05-19
Split CLI developer documentation

sbesson:
Following the CLI restructuring carried out in https://github.com/openmicroscopy/ome-documentation/pull/1151, this PR now splits the CLI developer documentation into separate pages.

### DIFF
--- a/omero/developers/cli/extending.txt
+++ b/omero/developers/cli/extending.txt
@@ -1,0 +1,21 @@
+Extensions
+----------
+
+Plugins can be written and put in the ``lib/python/omero/plugins``
+directory. On execution, all plugins in that directory are registered
+with the |CLI|. Alternatively, the "--path" argument can be used to point to
+other plugin files or directories.
+
+Thread-safety
+^^^^^^^^^^^^^
+
+The ``omero.cli.CLI`` should be considered thread-*un*\ safe. A single
+connection object is accessible from all plugins via
+``self.ctx.conn(args)``, and it is assumed that changes to this object
+will only take place in the current thread. The |CLI| instance itself,
+however, can be passed between multiple threads, as long as only one
+accesses it sequentially, possibly via locking.
+
+.. seealso::
+    |ExtendingOmero|
+        Other extensions to OMERO
--- a/omero/developers/cli/extending.txt
+++ b/omero/developers/cli/extending.txt
@@ -3,13 +3,13 @@ Extensions
 
 Plugins can be written and put in the ``lib/python/omero/plugins``
 directory. On execution, all plugins in that directory are registered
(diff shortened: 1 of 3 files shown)


omero/developers/cli/extending.txt
```diff
@@ -0,0 +1,21 @@
...
+
+Plugins can be written and put in the ``lib/python/omero/plugins``
+directory. On execution, all plugins in that directory are registered
+with the |CLI|. Alternatively, the "--path" argument can be used to point to
+other plugin files or directories.
+
+Thread-safety
+^^^^^^^^^^^^^
+
+The ``omero.cli.CLI`` should be considered thread-*un*\ safe. A single
```
ximenesuk: I don't think the emphasis works here and not sure about the compound. How about:

should *not* be considered thread-safe

or

should be considered *not* thread-safe

omero/developers/cli/extending.txt
```diff
@@ -0,0 +1,21 @@
+Extensions
+----------
+
+Plugins can be written and put in the ``lib/python/omero/plugins``
+directory. On execution, all plugins in that directory are registered
+with the |CLI|. Alternatively, the "--path" argument can be used to point to
```
ximenesuk: Does `--path` need to be quoted as a option?

omero/developers/cli/extending.txt
```diff
@@ -0,0 +1,21 @@
...
+The ``omero.cli.CLI`` should be considered thread-*un*\ safe. A single
+connection object is accessible from all plugins via
+``self.ctx.conn(args)``, and it is assumed that changes to this object
+will only take place in the current thread. The |CLI| instance itself,
+however, can be passed between multiple threads, as long as only one
+accesses it sequentially, possibly via locking.
+
+.. seealso::
+    |ExtendingOmero|
+        Other extensions to OMERO
```
ximenesuk: Not sure if I should insist on a proper line ending here.

ximenesuk:
Otherwise a useful split. It also raises the question of where to put pages about `delete` and `chgrp`: here with object creation or over is the sysadmin docs?

sbesson:
Agreed `delete`, `chgrp` but also `obj` plugins belong to the complex "power user" category. I do not have any strong opinion but it might be worth cross-referencing pages? My personal preference would go either to the developers section or to the users section (like `login` and `import`) /cc @hflynn

ximenesuk:
Maybe we need a separate section for cli power users? It would make some sense as there are commands that cross all boundaries even if some have limits for non-admin users.

ximenesuk:
Good to merge

joshmoore:
Assuming RFEs are captured elsewhere.

PR #1221, opened 2015-05-27
Signposting CLI docs better & minor tidy

hflynn:
As I commented on Colin's last CLI user docs PR, these docs are quite buried so this tries to signpost them better now that they are more extensive. I've also tried to make the page titles clearer and have them distinguish between the doc sections because as @mtbc has pointed out before, if you use the search box on the website to find CLI docs, the results only give the page title, meaning you have to follow the link or note the full URL to see whether you're in the users, sysadmins or developer section docs.

### DIFF
--- a/omero/developers/cli/index.txt
+++ b/omero/developers/cli/index.txt
@@ -1,6 +1,5 @@
-OMERO Command Line Interface
-============================
-
+Command Line Interface as an OMERO development tool
+===================================================
 
 
 .. toctree::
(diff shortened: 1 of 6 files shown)


sbesson:
Deployed on https://www.openmicroscopy.org/site/support/omero5.1-staging/. Looks good to me.

hflynn:
@mtbc has also looked over it and found no issues so merging.

PR #1222, opened 2015-05-28
Unit queries

joshmoore:
As a part of https://trello.com/c/vdyTStEq/330-unit-documentation,
this copies work from:

 * https://github.com/openmicroscopy/openmicroscopy/pull/3831 and
 * https://github.com/joshmoore/ome-units/blob/master/units/Fields.txt

into `Model/Units.txt`. @chris-allan rightly pointed out that this was
sorely lacking and should be clearly linked for anyone who needs to
upgrade their code from 5.0.

PR #1223, opened 2015-05-29
Adding units docs cross-references

hflynn:
See https://trello.com/c/vdyTStEq/330-unit-documentation - adds to #1222 by cross-referencing Data Model and OMERO developer units docs pages. Also adds API doc links to Laser wavelength on the Model Objects doc page.

### DIFF
--- a/formats/developers/ome-units.txt
+++ b/formats/developers/ome-units.txt
@@ -22,6 +22,10 @@ Units of Measurement API implementation developed at
 https://www.eclipse.org/uomo/ if future work on that project provides a
 more complete implementation.
 
+.. seealso:: 
+    :omero_doc:`OMERO developer documentation on units <developers/Model/Units.html>`
+    for further details of how this aspect of the Data Model is implemented in
+    OMERO.
 
 OME system of measurements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


hflynn:
Building on https://www.openmicroscopy.org/site/support/omero5.1-staging/ and https://www.openmicroscopy.org/site/support/ome-model-staging/developers/ome-units.html now

hflynn:
Ah ok, all the enumeration items on https://www.openmicroscopy.org/site/support/omero5.1/developers/Model/EveryObject.html need linking to the correct unit type in the javadocs, looks like I'll be finishing this next week!

hflynn:
Ok @mtbc I've added these manually so we can get the updated docs out asap but @joshmoore informs me you need to add this to your generation logic so that the changes aren't overridden next time you generate it! I'll make you a card for that.

joshmoore:
:+1: for the Model/Units.txt javadoc links and the other cross-linking.

The links from EveryObject.txt are also a huge improvement (thanks again for doing then manually!), but when it comes time to auto-generate then, rather than:
```
power.unit: enumeration (optional) from LightSource, see PowerI
power.value: double (optional) from LightSource, see PowerI
```
maybe `power.unit` could be linked to the `UnitsPower` enum, and somehow it be made clear that what the type of `power` is `PowerI` for which one must see the javadoc. **OR** we include `PowerI` as an object on this page itself (in which case, we could come back to the question of how to embed javadoc links in this page in the first place).

Assuming no immediate veto from @mtbc, I'd say merge and leave the two of us to hammer through code-generation fixes in a follow-up PR.

hflynn:
In the absence of any comment from @mtbc when I know he's been about, I'm going to take that as assent and merge and release this for now. You two can sort out how you want to handle it for the next release.

mtbc:
I'll see if I can automate this for 5.2.

PR #1233, opened 2015-07-01
Updating Upgrade Check page

hflynn:
Removing an internal section reported by Kenny (see https://trello.com/c/PAXHlAne/344-ome-registry-version-update-in-public-docs) and also stopping the link to the upgrade URL which redirects to version 4 docs being live after the discussion in devteam the other day (I realise the link needs updating in the qa repo and this is on the list somewhere, but there is no really no need to encourage people reading the docs to click on the link by having it live in the docs page, especially as once it is fixed, it will redirect you to the same page you are already on!).

### DIFF
--- a/omero/sysadmins/UpgradeCheck.txt
+++ b/omero/sysadmins/UpgradeCheck.txt
@@ -6,7 +6,7 @@ On each startup the OMERO server checks for available upgrades via the
 <components/common/src/ome/system/UpgradeCheck.java>`.
 An HTTP GET call is made to the URL configured in
 :source:`etc/omero.properties` as ``omero.upgrades.url``, currently
-`http://upgrade.openmicroscopy.org.uk` by default (note that viewing that link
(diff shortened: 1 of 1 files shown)


omero/sysadmins/UpgradeCheck.txt
```diff
@@ -6,7 +6,7 @@ On each startup the OMERO server checks for available upgrades via the
 <components/common/src/ome/system/UpgradeCheck.java>`.
 An HTTP GET call is made to the URL configured in
 :source:`etc/omero.properties` as ``omero.upgrades.url``, currently
-http://upgrade.openmicroscopy.org.uk by default (note that viewing that link
+`http://upgrade.openmicroscopy.org.uk` by default (note that viewing that link
```
sbesson: If you don't want to apply the verbatim markup, an alternative is to prepend the URL with `/`.

sbesson:
One minor comment/suggestion. Otherwise looks good.

sbesson:
:+1:

PR #1239, opened 2015-07-20
do not show settings.env header comment on walkthrough page

mtbc:
Omits the header comment introduced by https://github.com/ome/omero-install/pull/35 from the literal include on http://www.openmicroscopy.org/site/support/omero5.1-staging/sysadmins/unix/server-linux-walkthrough.html.


sbesson:
This file is actually a copy of https://github.com/ome/omero-install/blob/master/linux/settings.env. Want to open a PR against the upstream repo?

mtbc:
Thank you, now adjusted this PR accordingly.

hflynn:
Travis doesn't like this

mtbc:
Let's see if it needs a manual copy of that file from the other repository.

PR #1242, opened 2015-07-20
Removal of locked

hflynn:
See https://trello.com/c/o67lmdg6/34-removal-of-locked - adds breaking change of '-locked' removal to developers what's new page

### DIFF
--- a/omero/developers/whatsnew.txt
+++ b/omero/developers/whatsnew.txt
@@ -36,8 +36,8 @@ further details.
 OMERO.web template and view writers need to use the ``@render_response()``
 decorator, see :ref:`render-response` for more details.
 
-The permissions attribute on data manager has also been cleaned up, removing
-`-locked`.
+The redundant ``-locked`` flag has been removed from data manager tree nodes
+e.g. ``dataset-locked`` is now simply ``dataset``.
 
 Graphs
 ^^^^^^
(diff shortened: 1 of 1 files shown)


omero/developers/whatsnew.txt
```diff
@@ -36,6 +36,9 @@ further details.
 OMERO.web template and view writers need to use the ``@render_response()``
 decorator, see :ref:`render-response` for more details.
 
+The permissions attribute on data manager has also been cleaned up, removing
+`-locked`.
```
aleksandra-tarkowska: There is no need to mention about removal as that was removed long time ago. I just cleaned up few places it was missed

aleksandra-tarkowska: Wouldn't sound better if we say: attribute on **tree** data manager has?

will-moore:
I would say something like:
"The redundant '-locked' flag has been removed from the 'rel' attribute of data manager tree nodes. E.g. 'dataset-locked' is now simply 'dataset'."
or maybe even simpler:
"The redundant '-locked' flag has been removed from data manager tree nodes. E.g. 'dataset-locked' is now simply 'dataset'."

hflynn:
Fix pushed thanks

will-moore:
Thanks @hflynn. You OK with this @aleksandra-tarkowska?

aleksandra-tarkowska:
Fine by me.

PR #1241, opened 2015-07-20
updating mail doc

aleksandra-tarkowska:
better formatting

PR #1243, opened 2015-07-20
Fixing formatting on omero.mail page

hflynn:
Replaces #1241
Should get rid of the purple highlighting in plone and also clarifies the mail.from property


sbesson:
:+1:

PR #1250, opened 2015-08-03
Link query references

mtbc:
Adds useful links for those new to HQL. Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/developers/GettingStarted.html#examples.

PR #1253, opened 2015-08-03
Link query references (rebased from develop)

mtbc:
Adds useful links for those new to HQL. Staged at http://www.openmicroscopy.org/site/support/omero5.1-staging/developers/GettingStarted.html#examples.

--rebased-from #1250


hflynn:
All good on staging, merging.

PR #1254, opened 2015-08-03
Removing contributing & model docs from 5.1 branch

hflynn:
See https://trello.com/c/3yYuGGxL/155-dev-5-1-branch-creation-tasks on docs board - these two sets of docs are only maintained on develop as we don't have more than one version on the go at once.

--no-rebase


sbesson:
:+1:

PR #1262, opened 2015-08-11
Review description of all OMERO.matlab getter functions

sbesson:
- Specify the owner/group context for the various signatures
- Fix loaded graph information
- Add documentation about group/owner parameter/key value

Documents https://github.com/openmicroscopy/openmicroscopy/pull/3839

### DIFF
--- a/omero/developers/Matlab.txt
+++ b/omero/developers/Matlab.txt
@@ -188,13 +188,14 @@ indicating how to load projects, datasets, screens…
 
 -  **Projects**
 
-The projects owned by the user currently logged in can be retrieved using the
-:source:`getProjects <components/tools/OmeroM/src/io/getProjects.m>`
-function::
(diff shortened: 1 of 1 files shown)


jburel:
```To load the whole graph (projects, datasets, images), pass `true` as an optional argument::``` => Maybe we add a warning. Since that call could load a large amount of data

jburel:
```To specify the owner or group of the projects...```: 
Not very clear
I will break it into 2 steps. 
 * To retrieve projects owned by a given member do...
 * To retrieve projects in a given group do...

jburel:
```If the dataset identifiers are known, they can retrieved independently of``` => missing be they can be retrieved

sbesson:
Pushed

jburel:
Looks good. Merging

sbesson:
--rebased-to #1273

PR #1266, opened 2015-08-18
Contributing: snoopy URLs

sbesson:
With the snoopycrimecop CI branch names now being unified, we can consider them as stable for development purposes. This PR implements this by setting up extlinks to the scc branches used by the CI jobs. All source/commit extlinks under contributing are also prefixed to distinguish between them.

To test this PR:
- check the https://ci.openmicroscopy.org/view/Docs/job/CONTRIBUTING-merge-docs/ is still green
- check there is no missing branch that needs to be replaced in the same manner

--no-rebase

### DIFF
--- a/contributing/ci-bio-formats.txt
+++ b/contributing/ci-bio-formats.txt
@@ -95,7 +95,8 @@ The branch for the 5.1.x series of Bio-Formats is develop.
 		#. Builds the Java generated sources using ``ant clean ome-xml-src``
 		#. Builds the C++ generated sources using ``cmake .. && make gensrc``
 		#. Commits all the generated sources
-		#. Pushes the branch to develop/latest/generated-sources. This branch
+		#. Pushes the branch to 
+		   :bf_scc_branch:`develop/latest/generated_sources`. This branch
 		   can be used as the baseline for comparison with other generated
 		   source jobs like :term:`BIOFORMATS-5.1-merge-generated-sources`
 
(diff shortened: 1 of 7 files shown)


contributing/ci-bio-formats.txt
```diff
@@ -140,12 +140,11 @@ The branch for the 5.1.x series of Bio-Formats is develop.
...
-		#. Checks out the develop/merge/daily branch of the
-		   snoopycrimecop fork of bioformats.git_
+		#. Checks out :bf_scc_branch:`develop/merge/daily`
 		#. Builds the Java generated sources using ``ant clean ome-xml-src``
 		#. Builds the C++ generated sources using ``cmake .. && make gensrc``
-		#. Commits all the generated sources and pushes to the
-		   develop/merge/generated-sources branch
+		#. Commits all the generated sources and pushes to
+		   :bf_scc_branch:`develop/merge/generated_sources` branch
 		#. Computes the diff with develop/latest/generated-sources and archives
```
hflynn: Very minor nitpick, but shouldn't this be 'develop/latest/generated_sources with an underscore given it is referring to the branch? And do you want to make this one a link too?

sbesson:
Pushed

hflynn:
I think this is all fine now, although I'm amused you managed such a long extlink (I'd have been inclined to just use bf_scc etc but I guess we *might* want to link to something on snoopy that isn't a branch at some point…)

hflynn:
I'll merge this tomorrow morning when I've rechecked it with fresher eyes

contributing/ci-docs.txt
```diff
@@ -184,14 +185,13 @@ dev_5_1.
 		This job is used to build the merge auto-generated pages for the
 		dev_5_1 branch of the Bio-Formats documentation
```
hflynn: If you are fixing the autogen branch references to still be develop, it would probably be less confusing if we stick with labelling the BF 5.1 docs as the develop branch for now.

hflynn:
Ok sorry, still one minor quibble because I think the mix of dev_5_1 and develop branches in the same job for the BF docs is confusing, but otherwise this is all good now, thanks.

jburel:
Few small edits noticed:

* contributing/ci-omero.txt
526 Push => Pushes

* contributing/ci-bio-formats.txt
310 and 317 Run => Runs

* contributing/ci-docs.txt
177, 192 Check => Checks

* contributing/schema-changes.txt
87: Extra space before ```.  The :omero_source:`sql/README.txt```

sbesson:
Pushed.

PR #1269, opened 2015-08-20
Modify CLI tag page

ximenesuk:
This PR makes a couple of chnages to the CLI tag documenation. One is related to a commit on https://github.com/openmicroscopy/openmicroscopy/pull/4061 allowing tags to be created without a description.

The second relates to deleting tags and is in response to https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7884&p=16139#p16137 

A question regarding this second commit is whether the examples should be added to the `delete` page and a simple link be provided or if it is okay to have this explanation of one aspect of a command on anthor command page.

--rebased-to #1270

### DIFF
--- a/omero/users/cli/delete.txt
+++ b/omero/users/cli/delete.txt
@@ -112,6 +112,8 @@ contained elsewhere, excluding any images in those datasets but including
 any file annotations linked to the deleted datasets. In this case the images
 that are not otherwise contained in datasets will be orphaned.
 
+For an example on deleting tags directly see :ref:`delete_tags`.
+
 Further options
 ^^^^^^^^^^^^^^^
 
--- a/omero/users/cli/tag.txt
+++ b/omero/users/cli/tag.txt
@@ -64,3 +64,18 @@ command. The object must be specified as ``object_type:object_id``. To link
 the tag of identifier ``1260`` to the Image of identifier ``1000``, use::
 
     $ bin/omero tag link Image:1000 1260
+
+Delete tags
+^^^^^^^^^^^
+
+Tags can be deleted using the :omerocmd:`delete` command. The tag or tag set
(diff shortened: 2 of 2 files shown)


hflynn:
You can use ref mark-up to add a pointer to new the new section on the delete page.
Add `.. _deleting_tags` above your new heading and then you can use ``:ref:`deleting_tags` to add a reference on the delete page and it will automatically use your heading for the link text.

hflynn:
There's a reference to TagAnnotation in the first section of https://www.openmicroscopy.org/site/support/omero5.1-staging/users/cli/delete.html, I'd stick the ref at the end of "Additionally, objects of the three annotation types, FileAnnotation, TagAnnotation and TermAnnotation are not deleted by default when the objects to which they are linked are deleted."

hflynn:
Changes to this page look fine on staging, probably good to merge once the ref is added.

hflynn:
This will need rebasing to develop once merged btw

ximenesuk:
@hflynn, thanks. I'll add in the reference.

Re https://github.com/openmicroscopy/ome-documentation/pull/1269#issuecomment-133019341, this is not quite the same thing as on the tag page it is about directly deleting tags rather than as a side-effect of deleting other objects that may be tagged.

The PR this relates to has not yet been merged - more chnages migh be added to this if that PR is chnaged. The develop PR has yet to be opened (which is slightly different) but yes I'll rebase this as soon as possible.

hflynn:
Fine to add the ref somewhere else.

If you'd rather do a separate PR for develop to match the different code changes (rather than rebase and then add further), go ahead, I didn't realise there were going to be different code changes there.

hflynn:
should be good to merge once everything is green

PR #1272, opened 2015-08-25
docs: Use urllib conditionally

ghost:
While this shouldn't be encountered in practice (it's a default python2 module?), it was mentioned as a problem by a user on IRC.  This moves the urllib import into the existing try block so that if the import fails, it's not a fatal error.  May also ease future python3 porting, since the docs are likely already buildable directly with a python3 version of sphinx-build.

---

Testing: Check the jobs are green.

### DIFF
--- a/common/conf.py
+++ b/common/conf.py
@@ -329,11 +329,11 @@ if not (sys.version_info[0] == 2 and sys.version_info[1] <= 5):
 
 # Regular expressions that match URIs that should not be checked when doing a linkcheck build
 linkcheck_ignore = []
-import urllib
-brokenfiles_url = 'https://raw.github.com/openmicroscopy/sphinx-ignore-links/master/broken_links.txt'
 try:
-   brokenlinks = urllib.urlopen(brokenfiles_url)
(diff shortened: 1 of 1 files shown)


manics:
Was the user trying to build on Python 3?
https://docs.python.org/2/library/urllib.html

>The urllib module has been split into parts and renamed in Python 3 to urllib.request, urllib.parse, and urllib.error.

rleigh-dundee:
@manics I think that's almost certainly the reason.  The workaround should at least make things work.  But the underlying design here--pulling a file online out of another git repo, is not a good one.

sbesson:
Works fine in terms of the CI. In general, if we think we have reached enough stability to have the linkcheck list  included in the main repository (which was the primary reason for splitting), let's have a follow-up PR or a card to do so /cc @hflynn

hflynn:
Does this need rebasing to dev_5_1 for 5.1.4 in the meantime? If not, can we get a no rebase label on this please?

sbesson:
--no-rebase

PR #1278, opened 2015-09-08
Matlab: add documentation for write functions group support

sbesson:
See https://trello.com/c/gSKqCAe2/49-omero-matlab-group-support

This PR:
- reviews the `writexxAnnotation` functions giving concrete examples for each existing annotation type
- adds example of annotation writing command in a different group
- documents the project/dataset/screen/plate creation functions with and without group support

### DIFF
--- a/omero/developers/Matlab.txt
+++ b/omero/developers/Matlab.txt
@@ -643,7 +643,7 @@ function::
     tagAnnotation = writeTagAnnotation(session, 'tag name');
     % Create a timestamp annotation
     timestampAnnotation = writeTimestampAnnotation(session, now);
-    % Create a XML annotation
+    % Create an XML annotation
     xmlAnnotation = writeXmlAnnotation(session, xmlString);
 
 File annotations can also be created from the content of a
--- a/omero/developers/Matlab.txt
+++ b/omero/developers/Matlab.txt
@@ -692,15 +692,17 @@ Writing data
 Projects and datasets can be created in the context of the session group
 using the :source:`createProject <components/tools/OmeroM/src/io/createProject.m>` and :source:`createDataset <components/tools/OmeroM/src/io/createDataset.m>` functions::
 
-    % Create a new project and a dataset
+    % Create a new project in the context of the session group
     newproject = createProject(session, 'project name');
+    % Create a new dataset in the context of the session group
     newdataset = createDataset(session, 'dataset name');
 
 Writing projects/datasets in a different context than the session group can be
 achieved by passing the group identifier using the `group` parameter::
 
-    % Create a new project and a linked dataset in the specified group
+    % Create a new project in the specified group
     newproject = createProject(session, 'project name', 'group', groupId);
+    % Create a new dataset in the specified group
     newdataset = createDataset(session, 'dataset name', 'group', groupId);
 
 When creating a dataset, it is possible to link it to an existing project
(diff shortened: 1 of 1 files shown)


omero/developers/Matlab.txt
```diff
@@ -631,14 +631,37 @@ known::
...
+    xmlAnnotation = writeXmlAnnotation(session, xmlString);
+
+File annotations can also be created from the content of a
+:file:`local_file_path`::
 
-    tagAnnotation = writeTagAnnotation(session, tag_name);
+    fileAnnotation = writeFileAnnotation(session, local_file_path);
 
-To create a file annotations from the content of a ``local_file_path``::
+Each annotation creation function will use the context of the session group by
```
mtbc: "uses" present tense?

sbesson:
Staged at https://www.openmicroscopy.org/site/support/omero5.1-staging/developers/Matlab.html

imunro:
At the risk of sounding like a broken record on this I would suggest adding after the examples at lines 709 
% Create a new project and a linked dataset in the specified group
+    newproject = createProject(session, 'project name', 'group', groupId);
+    newdataset = createDataset(session, 'dataset name', project, 'group', groupId);

and 732 
 % Create a new screen and a linked plate in the specified group
+    newscreen = createScreen(session, 'screen name', 'group', groupId);
+    newplate = createPlate(session, 'plate name', screen, 'group', groupId); 

a line explaining that passing the groupId in the second line isn't actually required as the command will default to -1 in this case and is shown only as an example. Also if a groupId is passéd then it must match that of the group to which the project/screen belong.

sbesson:
@imunro: I am pretty sure that with the current implementation of https://github.com/openmicroscopy/openmicroscopy/pull/4145 passing the group is currently mandatory to create a linked datasets/plate in a different group since we are not retrieving the group of the parent object.

If not passing a group option is the preferred way, we can modify the implementation and systematically reload the parent object to retrieve its group details. In this case, this leaves the question of whether passing the groupId in the following call:

```
newdataset = createDataset(session, 'dataset name', project, 'group', groupId);
```

should:

1- override the project group with `groupId` i.e. leave the full granularity in the API
2- still use the project group unconditionally, i.e. be a no-op
3- optionally throw a warning in the case of 1) or 2)
4- not be allowed
This is the same question as https://github.com/openmicroscopy/openmicroscopy/pull/4145#issuecomment-138594704.

imunro:
From my testing it appears that the above command:
newdataset = createDataset(session, 'dataset name', project, 'group', groupId);
will only work when groupId matches the group of "project" . My feeling is that this should be disallowed to simplify the interface - however it is currently allowed and therefore IMO requires explanation.

sbesson:
Last commit should update the object creation documentation to match the new API semantics proposed in #4145

imunro:
Changes to documentation look fine to me. Good to Merge.

jburel:
* an XML annotation
* 
```
 % Create a new project and a linked dataset in the specified group
  newproject = createProject(session, 'project name', 'group', groupId);
  newdataset = createDataset(session, 'dataset name', 'group', groupId);
```
Not clear since in that case the dataset is not linked to the project. This is done later on in the example.

Same comment for screen plate below

sbesson:
:+1: Commit pushed to update the descriptions.

jburel:
Shouldn't it be an XML annotation instead of "a XML"?

jburel:
travis is green

sbesson:
--rebased-to #1279

PR #1279, opened 2015-09-10
Matlab: add documentation for write functions group support (rebased onto develop)

sbesson:
This is the same as gh-1278 but rebased onto develop.

---

See https://trello.com/c/gSKqCAe2/49-omero-matlab-group-support

This PR:
- reviews the `writexxAnnotation` functions giving concrete examples for each existing annotation type
- adds example of annotation writing command in a different group
- documents the project/dataset/screen/plate creation functions with and without group support


sbesson:
--rebased-from #1278

sbesson:
See https://www.openmicroscopy.org/site/support/omero5.2-staging/developers/Matlab.html. Note the last commit was missed from the original PR and will have to be rebased to `dev_5_1`

jburel:
looks good.

PR #1280, opened 2015-09-11
Migrate to Travis Docker container

sbesson:
This PR  migrates the Travis build to using Docker containers in order to speed up build times as well as the queuing for the organization.

Like https://github.com/openmicroscopy/bioformats/pull/1968, one caveat of this PR is that it drops the `pdf` target of the top-level build command. This target is however tested by the various Jenkins doc jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,18 @@
 language: python
 
-python:
-  - "2.6"
+# This (sudo: false) is needed to "run on container-based infrastructure" on
+# which cache: is available
+# http://docs.travis-ci.com/user/workers/container-based-infrastructure/
+sudo: false
+
+# http://docs.travis-ci.com/user/caching/
+cache: pip
+
 
 env:
-  - SPHINXOPTS="-W" CLASSPATH="/usr/share/java/ant-contrib.jar"
+  - SPHINXOPTS="-W"
 
 before_install:
-  - sudo add-apt-repository -y "deb http://archive.ubuntu.com/ubuntu/ trusty main universe"
-  - travis_retry sudo apt-get update
-  - travis_retry sudo apt-get -y install ant ant-contrib ant-optional
-  - travis_retry sudo apt-get -y install texlive-latex-base texlive-latex-recommended texlive-xetex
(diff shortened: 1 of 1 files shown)


joshmoore:
Looks good to me! And the ~1min run time is great. As discussed w.r.t 1968 as well, the most likely annoyance of this change is that a PR with PDF errors will get merged which will then fail the jenkins build. If that becomes a concern, I'd propose we copy out the job to `*-pdf` which will then need handling separately.

joshmoore:
Oops, one quick question: what was the previous purpose of the `travis-merge` call?

sbesson:
Good catch. `travis-merge` was likely merging dependent PRs typically PRs requiring unmerged auto-generated changes. My feeling is that we should be able to not be dependent on this logic by including dependent commits as part of the same PR. But I can also restore it if we think there might be an use case.

hflynn:
A pdf with errors would break the merge build surely? So as long as nothing is merged without running through Jenkins, it should be fine. And since we know it's an issue, we can also check pdf builds locally before/when PRs are opened.

hflynn:
Are there any other implications for this? Because if it's just the pdf thing, I'm happy to merge.

sbesson:
Mostly the `travis-merge` discussion, i.e do we preserve the ability for Travis to merge multiple dependent PRs for building or do we force each PR to be buildable independently?

hflynn:
Did we ever use that functionality in the docs anyway? I know the staging build merges autogen stuff in that isn't merged yet, but I can't remember having a docs PR that depended on unmerged autogen content.

sbesson:
Looking at the repo history, the last time I remember having to use it was https://github.com/openmicroscopy/ome-documentation/pull/893, i.e. 5.0.3. I would be inclined to say these are exceptional cases that could also be handled by including the autogenerated commit in the PR itself.

hflynn:
I agree, does seem exceptional and given we have a way to handle it if needed, not a blocker for this PR.

sbesson:
--rebased-to #1281

PR #1281, opened 2015-09-14
Migrate to Travis Docker container (rebased onto dev_5_1)

sbesson:
This is the same as gh-1280 but rebased onto dev_5_1.

----

This PR  migrates the Travis build to using Docker containers in order to speed up build times as well as the queuing for the organization.

Like https://github.com/openmicroscopy/bioformats/pull/1968, one caveat of this PR is that it drops the `pdf` target of the top-level build command. This target is however tested by the various Jenkins doc jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,18 @@
 language: python
 
-python:
-  - "2.6"
+# This (sudo: false) is needed to "run on container-based infrastructure" on
+# which cache: is available
+# http://docs.travis-ci.com/user/workers/container-based-infrastructure/
+sudo: false
+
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #1280

hflynn:
Build looks fine, beyond checking the jenkins build is green with this included, is there anything I need to test here?

sbesson:
@hflynn: not really, the Travis greenness is the main check. Since the change was in `.travis.yml` only, this should have no impact on Jenkins but you can re-run the jobs to be on the safe side.

hflynn:
Jenkins job green too, merging.

PR #1302, opened 2015-10-23
bump model doc reference to new 5.2 OMERO

mtbc:
Effects https://trello.com/c/lUQDSthL/81-update-model-docs-to-point-at-omero-5-2-docs.

--no-rebase

### DIFF
--- a/formats/conf.py
+++ b/formats/conf.py
@@ -44,7 +44,7 @@ model_extlinks = {
     # API
     'javadoc' : (downloads_root + '/latest/bio-formats5.1/api/%s', ''),
     # Doc links
-    'omero_doc' : (oo_site_root + '/support/omero5.1/%s', ''),
+    'omero_doc' : (oo_site_root + '/support/omero5.2/%s', ''),
     'bf_doc' : (oo_site_root + '/support/bio-formats5.1/%s', ''),
     # Downloads
     'bf_downloads' : (downloads_root + '/latest/bio-formats5.1/%s', ''),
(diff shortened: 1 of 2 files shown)


hflynn:
May as well leave this open and then decide whether to bother with it later next week thanks.

sbesson:
Can you also update https://github.com/openmicroscopy/ome-documentation/blob/develop/omero/themes/globalomerotoc.html ?

formats/conf.py
```diff
@@ -44,7 +44,7 @@
     # API
     'javadoc' : (downloads_root + '/latest/bio-formats5.1/api/%s', ''),
     # Doc links
-    'omero_doc' : (oo_site_root + '/support/omero5.1/%s', ''),
+    'omero_doc' : (oo_site_root + '/support/omero5.2/%s', ''),
```
hflynn: I wonder whether it would actually be better to remove the version numbers here altogether - links only go to the model, units & structured annotations sections of the dev docs anyway.

formats/conf.py
```diff
@@ -44,7 +44,7 @@
...
     'bf_doc' : (oo_site_root + '/support/bio-formats5.1/%s', ''),
     # Downloads
     'bf_downloads' : (downloads_root + '/latest/bio-formats5.1/%s', ''),
```
hflynn: We probably want to make this BF downloads link versionless too so it's just the most recent latest link.

hflynn:
Thanks!

PR #1310, opened 2015-10-28
Ice36 ref

jburel:
Remove remaining ref to ice 3.6 
Not supported in 5.2.0

### DIFF
--- a/omero/sysadmins/version-requirements.txt
+++ b/omero/sysadmins/version-requirements.txt
@@ -149,7 +149,7 @@ The planned changes for 5.2 depend upon our support of
   * [5.2] 3.4 dropped, 3.5 deprecated, 3.6 not supported
   * [5.3] 3.4 dropped, 3.5 deprecated, 3.6 recommended
   * Rationale: 3.4 is no longer supported by ZeroC with 3.5 being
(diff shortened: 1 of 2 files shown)


omero/sysadmins/version-requirements.txt
```diff
@@ -149,7 +149,7 @@ The planned changes for 5.2 depend upon our support of
   * [5.2] 3.4 dropped, 3.5 deprecated, 3.6 not supported
```
hflynn: I don't think we can really say 3.5 is deprecated if we aren't supporting 3.6. This will have to be changed to "3.5 recommended,"

jburel: good point. We also need to discuss if 3.4 is dropped cf. https://trello.com/c/QvS3yU4h/95-blocker-revert-requirement-changes

hflynn: I was under the impression 3.4 had already been removed from downloads and testing but I defer to @sbesson & @joshmoore

jburel:
@hflynn: updated.

mtbc:
Good to merge. Looking forward to reverting some of it for 5.3!

PR #1307, opened 2015-10-26
Update auto-generated documentation

snoopycrimecop:
Repository: openmicroscopy/ome-documentation
Already up-to-date.

Generated by build [OMERO-DEV-latest-docs-autogen#149](https://ci.openmicroscopy.org/job/OMERO-DEV-latest-docs-autogen/149/).

---

--no-rebase

PR #1311, opened 2015-10-28
Use global replacements in autogen_docs sed commands

sbesson:
This should replace multiple instances of $WORKSPACE in the same line like
in the Apache configuration file.

### DIFF
--- a/omero/autogen_docs
+++ b/omero/autogen_docs
@@ -31,10 +31,10 @@ echo "Generating DB script example"
 
 echo "Generating Web configuration templates"
 # Nginx / WSGI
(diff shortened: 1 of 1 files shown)


sbesson:
See https://github.com/snoopycrimecop/ome-documentation/commit/2e46dba1cdb3e0898ef7afbce85e85741e86471b for the application on this PR on top of #1307.

mtbc:
The diff looks plausible.

manics:
LGTM

rleigh-dundee:
Looks fine to me also.

PR #1316, opened 2015-10-29
Update link

jburel:
Update link to point to a more suitable page.

cc @rleigh-dundee


hflynn:
That page doesn't mention Django, does it still need adding to the server installation pages?

hflynn:
Does help if I check the right staging pages! Looks good, thanks.

hflynn:
Will let @rleigh-dundee reply before I merge but this works for me.

rleigh-dundee:
@hflynn It's already documented on the server installation page, so it's all fine with me.

PR #1305, opened 2015-10-26
OMERO.web upgrade

atarkowska:
This PR update OMERO.web related documentation regarding:
- https://github.com/openmicroscopy/openmicroscopy/pull/4283  removing FastCGI support
- https://github.com/openmicroscopy/openmicroscopy/pull/4236 + https://github.com/openmicroscopy/openmicroscopy/pull/4277 Django18 upgrade

cc: @sbesson

PR #1320, opened 2015-10-29
Small text update

atarkowska:
updating text as @rleigh-dundee said in https://github.com/openmicroscopy/ome-documentation/pull/1305

> Reads ambiguously to me. "the web framework is no longer shipped with Django" means that Django no longer provides the web framework, when I think this means to say that "the OMERO web framework no longer bundles a copy of Django", which is the opposite way around.

cc: @jburel

### DIFF
--- a/omero/developers/Web/Deployment.txt
+++ b/omero/developers/Web/Deployment.txt
@@ -11,8 +11,8 @@ external OMERO server. You can add the server to the
 You should enable :property:`omero.web.debug` and start a lightweight
 development Web server on your local machine.
 
-.. note:: Since OMERO 5.2, the web framework is no longer shipped with the
-    Django package and requires manual installation of the Django dependency.
+.. note:: Since OMERO 5.2, the OMERO web framework no longer bundles a copy
+    of Django package and requires manual installation of the Django dependency.
     It is highly recommended to use `Django 1.8`_ (LTS) which requires
     Python 2.7. For more information see :ref:`python-requirements` on the
     :doc:`/sysadmins/version-requirements` page.
--- a/omero/developers/Web/Deployment.txt
+++ b/omero/developers/Web/Deployment.txt
@@ -11,10 +11,11 @@ external OMERO server. You can add the server to the
 You should enable :property:`omero.web.debug` and start a lightweight
 development Web server on your local machine.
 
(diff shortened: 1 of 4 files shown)


hflynn:
Reads better to me

omero/developers/Web/Deployment.txt
```diff
@@ -11,8 +11,8 @@ external OMERO server. You can add the server to the
 You should enable :property:`omero.web.debug` and start a lightweight
 development Web server on your local machine.
 
-.. note:: Since OMERO 5.2, the web framework is no longer shipped with the
-    Django package and requires manual installation of the Django dependency.
+.. note:: Since OMERO 5.2, the OMERO web framework no longer bundles a copy
+    of Django package and requires manual installation of the Django dependency.
```
ximenesuk: 'a copy of Django package' -> 'a copy of the Django package'

omero/sysadmins/unix/install-web.txt
```diff
@@ -38,8 +38,8 @@ Prerequisites
           :djangodoc:`how to install Django 1.8 <topics/install/#install-the-django-code>`
           or :djangodoc:`hot to upgrade Django to 1.8 <topics/install/#remove-any-old-versions-of-django>`.
 
-   .. note:: Since OMERO 5.2, the web framework is no longer shipped with the
-       Django package and requires manual installation of the Django
+   .. note:: Since OMERO 5.2, the OMERO web framework no longer bundles a copy
+       of Django package and requires manual installation of the Django
```
ximenesuk: 'a copy of Django package' -> 'a copy of the Django package'

omero/developers/Web/Deployment.txt
ximenesuk: I think the whole sentence might be better as something like:

> Since OMERO 5.2, the OMERO web framework no longer bundles a copy
> of the Django package,  instead manual installation of the Django dependency
> is required.

in both cases.

aleksandra-tarkowska:
@ximenesuk see now

ximenesuk:
This all looks good in the commits. It's probably worth double checking on staging before merging.

hflynn:
Staging is fine thanks, merging.

PR #1329, opened 2015-11-10
Include specification samples

sbesson:
As reported in https://github.com/openmicroscopy/bioformats/pull/2068#issuecomment-155421039, the minimal specification on our formats documentation does not pass xmlvalid. This PR modifies the specifications to include external OME-XML files via `literalinclude`. These OME-XML samples can be tested like the rest our [samples](https://www.openmicroscopy.org/Schemas/Samples/2015-01/). /cc @rleigh-dundee @mtbc @joshmoore @melissalinkert

### DIFF
--- a/formats/specifications/compliant-hcs.txt
+++ b/formats/specifications/compliant-hcs.txt
@@ -16,67 +16,7 @@ This sample structure is based on Sample One in the
 OME Compliant File Specification 
 (see :doc:`index`):
 
-::
-
-    <?xml version="1.0" encoding="UTF-8"?>
-    <OME:OME xmlns:ROI="http://www.openmicroscopy.org/Schemas/ROI/2015-01"
-        xmlns:OME="http://www.openmicroscopy.org/Schemas/OME/2015-01"
-        xmlns:BIN="http://www.openmicroscopy.org/Schemas/BinaryFile/2015-01"
-        xmlns:SPW="http://www.openmicroscopy.org/Schemas/SPW/2015-01"
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2015-01
-                            http://www.openmicroscopy.org/Schemas/OME/2015-01/ome.xsd">
-
-        <SPW:Plate 
-            ID="Plate:1" 
-            Name="Control Plate" 
-            ColumnNamingConvention="letter" 
-            RowNamingConvention="number" 
-            Columns="12" 
-            Rows="8"
-            >
-            <SPW:Description></SPW:Description>
-
-            <!-- repeat SPW:Well for # of wells in the plate that contain images -->
-            <SPW:Well ID="Well:1" Column="0" Row="0">
-                <!-- repeat SPW:WellSample for # of images taken in the well -->
-                <SPW:WellSample ID="WellSample:1" Index="0">
-                    <!-- 
-                            if there is an image associated with this SPW:WellSample
-                            it is linked using an SPW:ImageRef
-                        -->
-                    <OME:ImageRef ID="Image:0"/>
-                </SPW:WellSample>
-            </SPW:Well>
-        </SPW:Plate>
-        <!-- plus one more Plate for each Plate in this set -->
-
-        <!-- SPW:Screen is not required -->
-
-        <!-- The OME:Image element follows the structure for the OME Compliant File Specification -->
-        <OME:Image ID="Image:0" Name="Series 1">
-            <OME:AcquisitionDate>2008-02-06T13:43:19</OME:AcquisitionDate>
-            <OME:Description>An example OME compliant file, based on Olympus.oib</OME:Description>
-            <OME:Pixels DimensionOrder="XYCZT" ID="Pixels:0" 
-                PhysicalSizeX="0.207" PhysicalSizeY="0.207" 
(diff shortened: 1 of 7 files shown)


ghost:
The other thing here is the missing `BinaryFile.xsd` in `schemaLocation`.  Could that also be part of the original problem?

joshmoore:
After checking through a number of tools, I wonder to what degree this is actually a problem of the configuration of our validator as opposed to the file itself. Both the original, as well as @sbesson's version, as well as this diff:

```
# jamoore@joshuas-mbp:/tmp$ diff orig.ome.xml min.ome.xml
2c2
<     <ome:OME xmlns:bf="http://www.openmicroscopy.org/Schemas/BinaryFile/2015-01"
---
> <ome:OME
3a4
>         xmlns:bf="http://www.openmicroscopy.org/Schemas/BinaryFile/2015-01"
6c7,9
<                             http://www.openmicroscopy.org/Schemas/OME/2015-01/ome.xsd">
---
>         http://www.openmicroscopy.org/Schemas/OME/2015-01/ome.xsd
>         http://www.openmicroscopy.org/Schemas/BinaryFile/2015-01
>         https://www.openmicroscopy.org/Schemas/BinaryFile/2015-01/BinaryFile.xsd">
```

all validate with:

```
$ cat build.xml
<project default="valid">
    <target name="valid">
        <schemavalidate>
            <fileset dir="${basedir}" includes="${file}"/>
        </schemavalidate>
    </target>
</project>
$ ant -Dfile=new.ome.xml
Buildfile: /private/tmp/build.xml

valid:

BUILD SUCCESSFUL
```

sbesson:
Last commit should migrate the samples to external files included via `literalinclude` to simplify diffs. I can certainly apply your Ant schema validation to all our sample files.
Indeed our validation tools expects a non-null default XML namespace as per https://github.com/openmicroscopy/bioformats/blob/v5.1.5/components/formats-common/src/loci/common/xml/ValidationSAXHandler.java#L75. So adding a `xmlns` attribute to any of our sample files is sufficient to restore validation using our tools:

```
sbesson@ls30630:specifications (minimal_specification) $ git diff
diff --git a/formats/specifications/minimum-specification.ome b/formats/specifications/minimum-specification.ome
index 9429dd3..aef7294 100644
--- a/formats/specifications/minimum-specification.ome
+++ b/formats/specifications/minimum-specification.ome
@@ -2,6 +2,7 @@
 <ome:OME xmlns:bf="http://www.openmicroscopy.org/Schemas/BinaryFile/2015-01"
     xmlns:ome="http://www.openmicroscopy.org/Schemas/OME/2015-01"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="http://www.openmicroscopy.org/Schemas/OME/2015-01"
     xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2015-01
                         http://www.openmicroscopy.org/Schemas/OME/2015-01/ome.xsd">
     <ome:Image ID="Image:1" Name="Name92">
sbesson@ls30630:specifications (minimal_specification) $ xmlvalid minimum-specification.ome 
Parsing schema path
http://www.openmicroscopy.org/Schemas/OME/2015-01/ome.xsd
Validating minimum-specification.ome
No validation errors found.
sbesson@ls30630:specifications (minimal_specification) $ ant -Dfile=minimum-specification.ome 
Buildfile: /opt/ome/ome-documentation/formats/specifications/build.xml

valid:

BUILD SUCCESSFUL
Total time: 17 seconds
```

joshmoore:
> So adding a xmlns attribute to any of our sample files is sufficient to restore validation using our tools:

Or perhaps just warn if only one or the other is null?

sbesson:
As the discussion might be migrated to https://github.com/openmicroscopy/bioformats/pull/2080 has for now, I propose to review the PR in its current state, i.e. mostly migrating the OME-XML samples as standalone files to simplify future diffs and maintenance (description updated).

hflynn:
Staging pages look fine, is this waiting on @bramalingam 's BF PR linked being reviewed?

sbesson:
I think the next step is to add these to the set of sample files regularly tested by the full-repository job (see https://trello.com/c/s7LvGwoQ/5-have-schema-samples-tested-by-the-full-repository-job). Any thoughts/suggestions on the file naming?

sbesson:
Now the upstream PR is merged, this PR should be ready to merge. The next step will be to:
- migrate these OME-XML under the tested samples files - see https://trello.com/c/s7LvGwoQ/5-have-schema-samples-tested-by-the-full-repository-job
- rename the compliant files e.g. `olympus.ome`, `deltavision.ome` (or use `.xml` as the extension for these files /cc @jburel )? /cc @hflynn

mtbc:
An `.xml` extension might be nice for opening the samples in editors that are aware of syntax.

hflynn:
Shouldn't the extensions be consistent between all the compliant files? i.e if we rename the two samples as .xml, the HCS and minimum-specification files should be .xml as well?

hflynn:
I'll merge this and open a PR to update the image sample links. Compliant files can be renamed in the docs once we've settled on a scheme and migrated them. Would `ome.xml` work or is that a silly question ;) ?

sbesson:
`.ome.xml` is certainly what I have seen elsewhere and makes things explicit enough. So :+1: from me.

hflynn:
I'd go `olympus.ome.xml`  and `deltavision.ome.xml` in that case then. Will do in my new PR, then we are covered regardless of whether the files get migrated in time

PR #1334, opened 2015-11-16
Document Homebrew jobs split

sbesson:
With decoupling, a single OME Homebrew job for Bio-Formats and OMERO does not
scale. This PR documents the job split.

/cc @mtbc


hflynn:
All fine on staging, merging and releasing.

PR #1336, opened 2015-11-23
Add superseded note to 5.1 docs

hflynn:
See https://trello.com/c/1VCpNEFp/195-google-shows-old-versions-as-recent
As an interim help, while we get the hang of getting Google to point at the new docs more quickly, I propose sticking a banner of the 5.1 docs to make it clear these aren't the most up to date ones any more.


hflynn:
Building now, check any page under https://www.openmicroscopy.org/site/support/omero5.1-staging/ for review

PR #1337, opened 2015-11-23
Removing broken zeroc link

hflynn:
Zeroc have updated their website and seem to have removed this content. Doesn't seem very important to me so I've just removed it - we use Ice so discussing a different thing we don't use is somewhat unneeded.


hflynn:
https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-merge-docs/129/warnings3Result/ - only homebrew showing as a broken link now

sbesson:
Makes sense to me. The broken Homebrew link will be reviewed and fixed as part of the OMERO 5.2.1 deployment epic.

PR #1338, opened 2015-11-24
Linux walkthrough

jburel:
Prepare the links to support new naming of install scripts


manics:
What would be nice is a cross-check of which files are copied by the autogen and which ones are included in the docs, so that if someone adds/removes a file from an external repo but doesn't update the docs it immediately causes an error.

hflynn:
Ok, so does this need to be merged before the job can be fixed? Sorry, I've rather lost track of the plan here.

jburel:
http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/unix/server-linux-walkthrough.html
Changes now on staging
We have one broken link for homebrew

jburel:
The links have been fixed but the content needs to be reviewed

hflynn:
This looks ok on staging. The technical content is all merged on omero-install already isn't it? I'd be inclined to merge and move on to the homebrew walk-through on a different PR and then restructuring separately.

jburel:
All the files have been merged on omero-install

hflynn:
Does your comment on #1342 mean you are planning more work here @jburel ? Or did you want input on whether to split this page up first?

jburel:
not in this PR
I will start working on another one

hflynn:
ok, shall I just merge this then and we can look at the content again on the new PR?

jburel:
This one was only to fix the broken links. So yes please

PR #1339, opened 2015-11-25
Cross-referencing key-value pairs/map annotation docs

hflynn:
Finding the docs for map annotations is a bit of a nightmare at the moment as Jason noticed - no mention of them in the Model docs where you might expect and because they are described as key value pairs in the OMERO docs, searching isn't very helpful either. This should hopefully redress that for now (next schema these annotations will probably change anyway (shouldn't be text annotations) and we should aim to add docs then).

See https://www.openmicroscopy.org/site/support/ome-model-staging/ for review and also try the search box on the sphinx pages.

### DIFF
--- a/formats/developers/index.txt
+++ b/formats/developers/index.txt
@@ -60,8 +60,8 @@ Support for additional dimensions is also covered:
 
 -  :doc:`6D, 7D and 8D Storage <6d-7d-and-8d-storage>`
 
-The new 'key-value pairs' Map Annotations introduced in the
-:doc:`/schemas/january-2015` are a type of Structured Annotation. Further
+Map Annotations, storing 'key-value pairs', are a type of Structured
+Annotation which were introduced in the :doc:`/schemas/january-2015`. Further
 information is available in the :omero_doc:`OMERO developer documentation on
 Key-value pairs <developers/Model/KeyValuePairs.html>`.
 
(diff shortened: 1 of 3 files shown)


formats/developers/index.txt
```diff
@@ -60,6 +60,11 @@ Support for additional dimensions is also covered:
 
 -  :doc:`6D, 7D and 8D Storage <6d-7d-and-8d-storage>`
 
+The new 'key-value pairs' Map Annotations introduced in the
```
ximenesuk: I wonder if it is worth dropping 'new' now so that we don't need to later? That they are introduced in the Jan 2015 schema may be enough to express the newness without it becoming outdated on the next schema release?

Could use of 'key-value pairs' as an adjective be avoided? Something like:

`Map Annotations, using key-value pairs, introduced in the...`

formats/developers/structured-annotations.txt
```diff
@@ -13,6 +13,11 @@ For more information on SAs from an OMERO-centric perspective, see the
 :omero_doc:`OMERO page on structured
 annotations <developers/Model/StructuredAnnotations.html>`.
 
+The new 'key-value pairs' Map Annotations introduced in the
```
ximenesuk: ditto from above.

hflynn:
Hopefully better now @ximenesuk

ximenesuk:
Thanks @hflynn, looks good to merge.

PR #1341, opened 2015-11-26
Add Ant install to README

will-moore:
Since you need Ant installed, added this to the README.


hflynn:
Renders fine, thanks.

PR #1348, opened 2015-12-08
Centos6

jburel:
Add walkthrough for centos6 python 2.7 (from ius) django
Some links are broken since some PRs have not yet been merged i.e. https://github.com/ome/omero-install/pull/64 and https://github.com/ome/omero-install/pull/63)

This PR contains some code snippets, the files will have to be "broken" up in ome/omero-install in order avoid that.
This will happen later on once we are happy with the content.

cc @kennethgillen

PR #1351, opened 2015-12-10
linux walkthrough

jburel:
Clarify that the page is for Ubuntu and Centos7


dominikl:
Looks fine. There are two things I noticed following the walkthrough:
- The installation of the OMERO server is done as "omero" user. In order to switch to this user, you have to set a password for the omero user at some point, e. g. as root `passwd omero`. Don't know if that has to be mentioned explicitly, guess an ubuntu sysadmin will be aware of that anyway. Also how to get a root shell, because in ubuntu you can usually just `sudo` as default.
- For some reason (maybe it's the osx terminal app?), when I tried to copy/paste the commands from the doc, this

```
apt-get -y install \
    unzip \
    wget \
    python-{imaging,matplotlib,numpy,pip,scipy,tables,virtualenv} \
    openjdk-7-jre-headless \
    ice-services python-zeroc-ice \
    postgresql
```

gets pasted into that (and doesn't work, note the dots)

```
root@ubuntu:~# apt-get -y install \
> .unzip \
> .wget \
> .python-{imaging,matplotlib,numpy,pip,scipy,tables,virtualenv} \
> .openjdk-7-jre-headless \
> .ice-services python-zeroc-ice \
> .postgresql
```

Can't really do anything about it, just curious, if it's just my terminal programm.

manics:
> In order to switch to this user, you have to set a password for the omero user at some point, e. g. as root passwd omero.

If you're logged on as root you can just do `su - omero` to switch to the `omero` acccount without a password.

> Also how to get a root shell, because in ubuntu you can usually just sudo as default.

One of `su -` or `sudo -i` should work, but it depends on how the base OS was setup.

dominikl:
That's true `su - omero` works without password, didn't know that. To get a root shell one has to do `sudo -i`, `su -` doesn't work. Mention in the docs? But like I said someone familiar with ubuntu probably knows that anyway.

manics:
OK, I wasn't sure whether you were referring to ubuntu only, or linux in general

jburel:
@dominikl: your last comment, I will add it to the testing card https://trello.com/c/QeAwP5Gz/104-installation-instructions-testing
The card will then be converted in a doc on ome-internal

dominikl:
Ah sorry about the confusion, yes I was refering to ubuntu, on centos both works, also `su -` .

dominikl:
The first command in `step05_centos7_apache24.sh` says `cp setup_omero_apache24.sh ~omero` But where is this file `setup_omero_apache24.sh`? (Testing following the staging doc)

jburel:
I need to add a line indicating to download it in that case

dominikl:
I tried the apache part several times with a virtualbox vm, everything runs through smoothly, omero starts fine, omero.web starts fine, but I can't get it in the browser. Maybe someone else can try it, too. Apparently I just can't get along with virtualbox.

omero/sysadmins/unix/server-linux-walkthrough.txt
```diff
@@ -1,7 +1,7 @@
-OMERO.server Linux walkthrough
-==============================
+OMERO.server installation on Ubuntu and Centos7
```
atarkowska: or CentOS 7

omero/sysadmins/unix/server-linux-walkthrough.txt
```diff
@@ -72,8 +78,39 @@ Nginx configuration directory, and disable the default configuration:
...
+As **root** install Apache 2.4::
+
+	apt-get -y install apache2 libapache2-mod-wsgi
+
+As the **omero** system user, configure OMERO.web:
+	
+.. literalinclude:: walkthrough/setup_omero_apache24.sh
+
+When installing on Ubuntu, edit the Apache OMERO configuration ``apache.conf.tmp``
+to set ``WSGISocketPrefix``::
```
atarkowska: We need it in all cases. Just say explicitly where exactly placing sockets and lock files used by Apache modules. Also add note to not put the sockets in the system temporary working directory.

atarkowska:
Thanks for updating doc, just small comments

omero/sysadmins/unix/server-linux-walkthrough.txt
jburel: the displayed is not ideal. Thanks

omero/sysadmins/unix/server-linux-walkthrough.txt
```diff
@@ -134,11 +182,13 @@ Copy this script into the appropriate location:
...
+
 SELinux
 -------
 
 If you are running a system with SELinux enabled (for example, it is
-`enabled by default on CentOS6 <http://wiki.centos.org/HowTos/SELinux>`_)
+`enabled by default on CentOS 6 <http://wiki.centos.org/HowTos/SELinux>`_)
 and are unable to access OMERO.web you may need to adjust the security policy:
 
 .. literalinclude:: walkthrough/setup_centos6_selinux.sh
```
sbesson: As mentioned in https://github.com/openmicroscopy/ome-documentation/pull/1363#issuecomment-16467943, this should be renamed to `setup_centos_selinux.sh`

hflynn:
Once this is fixed I'll be able to get the version history PR on staging

jburel:
@hflynn: done

hflynn:
All fixed now thanks, can we get this signed off or is there testing still going on?

omero/sysadmins/unix/server-linux-walkthrough.txt
```diff
@@ -26,12 +26,17 @@ either use your own values, or alternatively source the following file:
...
+
+Install Ice, Java, PostgreSQL:
 
 .. literalinclude:: walkthrough/step01_ubuntu1404_deps.sh
 
 | :download:`step01_ubuntu1404_deps.sh <walkthrough/step01_ubuntu1404_deps.sh>`
-| :download:`step01_centos6_deps.sh <walkthrough/step01_centos6_deps.sh>`
+| :download:`step01_centos7_deps.sh <walkthrough/step01_centos7_deps.sh>`
+
+.. note:: PostgreSQL needs to be reconfigured to accept TCP connections on CentOS 7.
```
joshmoore: But following the scripts, there's no manual action needed, right? During my testing, I certainly do anything extra

https://github.com/ome/omero-install/blob/develop/linux/step01_centos6_py27_deps.sh#L48

omero/sysadmins/unix/server-linux-walkthrough.txt
```diff
@@ -60,26 +67,66 @@ the omero user.
...
+
+  #see https://code.google.com/p/modwsgi/wiki/ConfigurationIssues
+  WSGISocketPrefix run/wsgi
+  # Use this on Ubuntu/Debian systems:
+  # WSGISocketPrefix /var/run/wsgi
+
+Copy the Apache OMERO configuration file into the
+Apache configuration directory, and disable the default configuration::
+
+    cp ~omero/OMERO.server/apache.conf.tmp /etc/apache2/sites-available/omero-web.conf
```
joshmoore, line 103: Where is this section copied from? Seems a bit complicated and is a potential source for bugs.

aleksandra-tarkowska: Why you need to disable default config, is that not what prefix is for?

omero/sysadmins/unix/server-linux-walkthrough.txt
```diff
@@ -60,26 +67,66 @@ the omero user.
...
+The default location set in ``apache.conf.tmp`` needs to be modified when 
+installing OMERO on an Ubuntu/Debian system::
+
+  #see https://code.google.com/p/modwsgi/wiki/ConfigurationIssues
+  WSGISocketPrefix run/wsgi
+  # Use this on Ubuntu/Debian systems:
+  # WSGISocketPrefix /var/run/wsgi
+
+Copy the Apache OMERO configuration file into the
+Apache configuration directory, and disable the default configuration::
```
atarkowska, line 101: Why you need to disable default config, is that not what prefix is for?

jburel, line 101: the default location set in the template file is run/wsgi
but for ubuntu we use /var/run/wsgi

atarkowska, line 101: what you mean exactly by `default configuration` ? sorry but this line doesn't refer to `WSGISocketPrefix` but to the 3 command below.

atarkowska, line 101: form what I read it says disable default apache config for http://host/, that reads like the only option is to deploy on root host. OMERO.web can be deployed under custom prefix why just not use it and leave config

omero/sysadmins/unix/server-linux-walkthrough.txt
jburel: That's step is required in centos7
to make it clearer we will need a dedicated a centos7 as suggested for I can add in the note below the links ``see line in script`

omero/sysadmins/unix/server-linux-walkthrough.txt
jburel, line 103: read the page https://code.google.com/p/modwsgi/wiki/ConfigurationIssues
Happy to reduce it:

```
The default value of ``WSGISocketPrefix``  in ``apache.conf.tmp`` needs to be modified when 
+installing OMERO on an Ubuntu/Debian system
```

omero/sysadmins/unix/server-linux-walkthrough.txt
jburel, line 101: Sorry I thought you were referring to WSGISocketPrefix

omero/sysadmins/unix/server-linux-walkthrough.txt
joshmoore: Understood. Or, "if you are not using the provided scripts for CentOS 7, you'll need to make reconfigure PostgreSQL to accept TCP connections"

omero/sysadmins/unix/server-linux-walkthrough.txt
joshmoore, line 103: Can some of the above be included in https://github.com/ome/omero-install/blob/develop/linux/step05_centos7_apache24.sh so that it'll be tested?

omero/sysadmins/unix/server-linux-walkthrough.txt
```diff
@@ -60,26 +67,66 @@ the omero user.
...
+
+Install Apache 2.4::
+
+	apt-get -y install apache2 libapache2-mod-wsgi
 
+As the **omero** system user, configure OMERO.web:
+	
+.. literalinclude:: walkthrough/setup_omero_apache24.sh
+
+In the Apache OMERO configuration ``apache.conf.tmp``, we explicitly set
```
atarkowska: as that is only Ubuntu Mac OS X specific maybe that shoudl go to note

omero/sysadmins/unix/server-linux-walkthrough.txt
joshmoore, line 103: or perhaps unnecessary as @aleksandra-tarkowska mentioned elsewhere.

omero/sysadmins/unix/server-linux-walkthrough.txt
atarkowska, line 101: No I mean https://github.com/openmicroscopy/ome-documentation/pull/1351/files#diff-ceb4317a407a746a651045face0ef7ebR115. That is not needed. We should use prefix https://www.openmicroscopy.org/site/support/omero5.2/sysadmins/config.html#omero-web-prefix.

jburel:
Discussed with @hflynn @joshmoore: breaking that page into 2

omero/sysadmins/unix/server-linux-walkthrough.txt
manics: @jburel Is this centos 7 specific? What changes did you need to make?

omero/sysadmins/unix/server-linux-walkthrough.txt
manics, line 101: Does that mean the default generated web config should be changed? Doesn't the same occur with nginx, where the default site also needs to be disabled?

omero/sysadmins/unix/server-linux-walkthrough.txt
manics, line 103: https://github.com/ome/omero-install/blob/develop/linux/step05_ubuntu1404_apache24.sh#L10 ?

manics:
Also a more general question, maybe for 5.3: Do we want to put the focus on Ubuntu 14.04 or CentOS 7 in the docs?

jburel:
@manics 
currently breaking into dedicated pages one for Ubuntu and one for centos7 since mixing the 2 was too confusing when we went offer the page for fresh install

jburel:
with the current config the following `a2dissite 000-default.conf` is needed

atarkowska:
@jburel 000-default.conf file is a catch-all site and should be left alone. I don't feel it is our responsibility to provide that deep details as in the real prod server you will have multiple  domain multiple configs and adding to the default has no sense. @manics correct me if I am wrong

atarkowska:
If we want to go this way then you will need to create your new/additional site(s) in your `/etc/apache2/site-available/`

joshmoore:
Comments on the latest state:
- I'd likely drop "-walkthrough" from the C7 file name. Only the now misnamed "linux-walkthrough" has it in the name, the others (centos6 and homebrew) are without.
- The ubuntu page should likely use the version in the title "OMERO.server installation on Ubuntu 14.04"
- Separation of just centos is **very** nice. Looking forward to the separation of apache in the next go :smile: Perhaps "Installing a web server" could be the title of the section, and each of "nginx" and "apache" could be subsections for now?
- I'm still slightly concerned the selinux section will be overlooked. @manics: under what situations is is _not_ ok to run that snippet?
- The note under http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/unix/server-linux-walkthrough.html#installing-apache has a green marker (Chrome/OS X)

For future refactorings:
- the centos6 page should use the same headers as the new centos7 and the original ubuntu pages.
- re: "should be read in conjunction with OMERO.server installation " -- I'm not sure how they should be read in conjunction, but @hflynn likely has a plan.
- either all pages should have "install" in the names or none
- the download links for the scripts should likely come above the blocks of code

jburel:
`re: "should be read in conjunction with OMERO.server installation " -- I'm not sure how they should be read in conjunction, but @hflynn likely has a plan.`
was kept from original text. Happy to remove it

atarkowska:
> The note under http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/unix/server-linux-walkthrough.html#installing-apache has a green marker (Chrome/OS X)

Why this still shows old text, I saw https://github.com/openmicroscopy/ome-documentation/pull/1351#discussion_r47480463  it was already updated?

manics:
@aleksandra-tarkowska The original purpose of the scripts was to provide an example complete working installation, and not to cater for any advanced configuration, or other applications running on the same server. The easiest way to do this is to assume OMERO.web is the only site on Apache/Nginx (and also that OMERO.server is the only thing running), because it provides a base from which we can troubleshoot if necessary.

jburel:
@joshmoore: adjusted sections
I have also noticed this green border when inserting code in note in other page e.g.
http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/unix/install-web.html 
Probably something in formatting in sphinx. I have not found a way to fix it

jburel:
@aleksandra-tarkowska: for the note, I went for a simplified version (but still more explicit that the first one) of the text to address @joshmoore's worries.

omero/sysadmins/unix/server-linux-walkthrough.txt
```diff
@@ -60,26 +62,69 @@ the omero user.
...
+.. note::
+        The default value set for the ``WSGISocketPrefix`` directive in ``apache.conf.tmp``
+        needs to be modified:
+
+        ::
+
+            #see https://code.google.com/p/modwsgi/wiki/ConfigurationIssues
+            WSGISocketPrefix run/wsgi
+            # Use this on Ubuntu/Debian systems:
+            # WSGISocketPrefix /var/run/wsgi
```
manics, line 118: Since there's a dedicated script for Ubuntu you could script this change
https://github.com/ome/omero-install/blob/develop/linux/install_ubuntu1404_apache24.sh

E.g. `sed -i -r -e 's|(WSGISocketPrefix run/wsgi)|#\1|' -e 's|# (WSGISocketPrefix /var/run/wsgi)|\1|' apache.conf.tmp`

jburel, line 118: true.  still need to be indicated in the doc since we do not yet have a literalinclude with markers

atarkowska:
@manics I understand and agree with that, is the same purpose for the doc?

hflynn:
The green highlighting is a plone CSS issue due to too much indenting, it should be avoidable unless you have something like a note box within an already indented block for a bullet point or something. I'll have a look but don't worry about it too much.

atarkowska:
@jburel thanks wasn't sure

jburel:
@aleksandra-tarkowska: the code snippet will be automatically extracted from the script (this round we did not have time to use markers and literalinclude) so it will show up in the doc.

omero/sysadmins/unix/server-centos7.txt
```diff
@@ -0,0 +1,178 @@
...
+Nginx configuration directory, and disable the default configuration:
+
+.. literalinclude:: walkthrough/step05_centos7_nginx.sh
+
+| :download:`step05_centos7_nginx.sh <walkthrough/step05_centos7_nginx.sh>`
+
+Apache
+^^^^^^
+
+**The following steps are run as root unless specified.**
```
atarkowska: Sorry for last minute comment but I think this is a bit confusing. Could we move line 93 `As the **omero**...` up to 95 `walkthrough/setup_omero_apache24...`  before installing apache then all the appropriate commands needs to be run as root. This will much two scripts down below.

jburel:
Fix the comment: Waiting on @manics for SELinux comment
Also modify layout of centos6 page https://github.com/openmicroscopy/ome-documentation/pull/1348
to match the layout used for centos7 and ubuntu

omero/sysadmins/unix/server-centos7.txt
```diff
@@ -0,0 +1,178 @@
...
+Apache configuration directory::
+
+    cp ~omero/OMERO.server/apache.conf.tmp /etc/httpd/conf.d/omero-web.conf
+
+    rm -rf /run/httpd/* /tmp/httpd*
+
+    systemctl enable httpd.service
+    systemctl start httpd
+
+| :download:`step05_centos7_apache24.sh <walkthrough/step05_centos7_apache24.sh>`
```
atarkowska, line 147: Could we swap the order of these scripts as well? or even move omero config one higher up.

omero/sysadmins/unix/server-centos7.txt
jburel: @aleksandra-tarkowska: it should have been put first. Thanks

joshmoore:
Added a few comments to the RFEs under https://github.com/openmicroscopy/ome-documentation/pull/1351#issuecomment-164767245 for later; otherwise:
- In [prereqs](https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/unix/server-centos7.html#prerequisites), "Workaround to get postgresql running inside Docker" might be confusing. I wonder if in this case it's worth hand-stripping that out (if possible) since we don't have any other mentions of docker on the pages.
- The web subitems looks good to me.
- In general, I think the user specification ("run as") are _much clearer_. The final sections ("Securing" and "Tasks") might need similar additions, but that can likely come with the later cleanup of `selinux`.
- Note: I haven't looked at the PDF or re-run the tests.

hflynn:
(Let's just not look at the pdf, I'm pretty sure no-one else does)

jburel:
For point 1: I will remove the include for this round and use marker next time
For point 3: will add that in this PR
I will do the same on centos6 when applicable

joshmoore:
Just looked at centos6 as well. Again, some adjustment re: the final tasks section (e.g. should "webapps" be on all pages?) but we can tackle that later.

jburel:
@joshmoore: The only reason I added the webapps section is b/c of usage of virtualenv.
I discussed with @pwalczysko and the conclusion that until we have a dedicated section elsewhere in the doc  that will probably avoid confusion

omero/sysadmins/unix/server-linux-walkthrough.txt
jburel, line 118: PR opened https://github.com/ome/omero-install/pull/76

sbesson:
Merging this and we can fix comments in follow up PRs if needed.

PR #1352, opened 2015-12-10
Removing out-of-date content & permissions fix

hflynn:
Removes limitations relating to OS we don't support any more anyway, 4.4 movies which are now hopelessly out-of-date (and replaced by help site content) and one remaining hardcoded reference to 5.1.x that had been missed previously.

### DIFF
--- a/omero/sysadmins/limitations.txt
+++ b/omero/sysadmins/limitations.txt
@@ -133,8 +133,8 @@ the shared images by cutting the links.
 Changing group permissions
 --------------------------
 
-If you have data which has been linked by other users, you cannot make the
-group that data is in into a private group.
+If a group contains a projection made by one member from data owned by another
+user, you cannot make the group into a private group.
 
 File format support
 -------------------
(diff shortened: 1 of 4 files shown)


jburel:
the following section can now be removed cf. https://trello.com/c/d7pYClU9/64-windows-py2-7-django-1-8-iis

```
Django 1.8 will not work with IIS on Windows
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

OMERO.web has been upgraded to use Django 1.8 but support for deploying it
using IIS on Windows is still upcoming and you must use Django 1.6 for now. We
aim to add support for Django 1.8 in OMERO.web within the 5.2.x development
line.
```

hflynn:
Is this good to merge now?

jburel:
very minor: Changing group permissions `etc` should be `etc.`

jburel:
``````
If you are using the advanced configuration setting ``FlexReaderServerMaps``
for importing Flex data split between multiple directories for use with
:doc:`OMERO.tables </developers/analysis>`, you should not upgrade to 5.1.0.```

we are now at 5.2.0. I still think it is still the case.
``````

hflynn:
Thanks @jburel - fixes pushed.

jburel:
Looking good. Ready to merge

PR #1355, opened 2015-12-11
Fix out-of-date postgres minimum version refs

hflynn:
These pages were obviously missed when the version requirements were updated for 5.2.

### DIFF
--- a/omero/sysadmins/unix/server-postgresql.txt
+++ b/omero/sysadmins/unix/server-postgresql.txt
@@ -9,8 +9,8 @@ version and that it is installed and configured correctly.
 Ensuring you have a valid PostgreSQL version
 --------------------------------------------
 
-For OMERO |version|, PostgreSQL version 9.2 or later is required; version 9.3
-or later is recommended. Make sure you are using a
+For OMERO |version|, PostgreSQL version 9.3 or later is required; version 9.4
+is recommended. Make sure you are using a
 `supported version <http://www.postgresql.org/support/versioning/>`_.
 
 You can check which version of PostgreSQL you have installed with any of
--- a/omero/sysadmins/unix/server-postgresql.txt
+++ b/omero/sysadmins/unix/server-postgresql.txt
@@ -19,11 +19,11 @@ the following commands:
 ::
 
        $ createuser -V
-       createuser (PostgreSQL) 9.3.4
+       createuser (PostgreSQL) 9.4.1
        $ psql -V
-       psql (PostgreSQL) 9.3.4
+       psql (PostgreSQL) 9.4.1
        $ createdb -V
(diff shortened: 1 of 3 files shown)


omero/sysadmins/windows/server-postgresql.txt
```diff
@@ -12,8 +12,8 @@ For Windows-specific installation instructions, first see
 Ensuring you have a valid PostgreSQL version
 --------------------------------------------
 
-For OMERO |version|, PostgreSQL version 9.2 or later is required; version 9.3
-or later is recommended. Make sure you are using a `supported version <http://www.postgresql.org/support/versioning/>`_.
+For OMERO |version|, PostgreSQL version 9.3 or later is required; version 9.4
+is recommended. Make sure you are using a `supported version <http://www.postgresql.org/support/versioning/>`_.
 
 If your existing PostgreSQL installation is version 9.2 or earlier, it
```
hflynn: Do we want to make this 9.3 too?

ximenesuk: It would make sense to otherwise 9.3 becomes a sort of limbo version.

ximenesuk:
The changes look fine to me barring the line comment above.

hflynn:
Should be sorted now @ximenesuk thanks

hflynn:
Ah, just found a bunch more places this has been missed. Another commit coming.

omero/sysadmins/unix/server-postgresql.txt
```diff
@@ -25,7 +25,7 @@ the following commands:
        $ createdb -V
        createdb (PostgreSQL) 9.3q.4
```
ximenesuk: I guess these code examples are okay as they are illustrating how to find the version. Does `9.3q.4` exist? Just wondering if the three version strings should be the same? On my system those three commands give:

```
$ createuser -V
createuser (PostgreSQL) 9.4.1
$ psql -V
psql (PostgreSQL) 9.4.1
$ createdb -V
createdb (PostgreSQL) 9.4.1
```

ximenesuk:
The other changes look consistent but it might be worth changing the command line example to get rid of the questionable `q`.

hflynn:
I expect the 'q' is a missed typing-in-the-wrong-place error ;) I've updated the output to match your example now anyway though.

ximenesuk:
Looks good to merge!

joshmoore:
NB: moving forward, should these number become variables like `|PG_MIN_VER|`?

jburel:
@joshmoore: we should introduce parameters like the one you suggested for all key ones: java, python, ice etc.
@hflynn: do we have a card already?

hflynn:
No, I'll stick one on the docs board.

PR #1356, opened 2015-12-11
Install indexes

hflynn:
In lieu of restructuring in a point release, this tries to make the installation index page clearer
see https://trello.com/c/YhIMg4wf/50-unix-index-html-improvements

### DIFF
--- a/omero/sysadmins/unix/index.txt
+++ b/omero/sysadmins/unix/index.txt
@@ -19,4 +19,3 @@ additional pages listed below.
     server-binary-repository
     server-postgresql
     install-web
-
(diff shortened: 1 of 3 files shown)


jburel:
Looks good.

joshmoore:
```
OMERO.server Mac OS X installation walk-through with Homebrew
OMERO.server installation on Ubuntu and Centos7
OMERO.server installation on Centos 6 with Python 2.7 and Ice 3.5
```

"...installation on OS X with Homebrew" ? or are the pages sufficiently different to warrant a different pattern?

hflynn:
Good to merge now?

joshmoore:
No objections from me (naturally). There was some suggestion today that the OS X page was more developer-focused. Not sure if that can/should impact the index.

hflynn:
I think quite a lot of the sysadmin content is only divided from developer content by a very fine line. The page is clearly defined at the top and where it is linked, I think that'll do for now.

PR #1358, opened 2015-12-13
Auto-gen update

jburel:
Add docker_virtual files to the list.

cc @hflynn

### DIFF
--- a/omero/autogen_docs
+++ b/omero/autogen_docs
@@ -51,6 +51,7 @@ for f in \
     omero-* \
     settings.env \
     setup_* \
+    docker_virtualenv*\
     ; do
     cp $WORKSPACE/omero-install/linux/$f $DIRECTORY
 done


omero/autogen_docs
```diff
@@ -51,6 +51,7 @@ for f in \
     omero-* \
     settings.env \
     setup_* \
+    docker_virtualenv*\
```
manics, line 54: Are these files only for Docker? If so why are they needed in the main docs?

jburel, line 54: We picked that term, I could not find a better one. I will review them post release. They are used in the server-centos6 installation

joshmoore, line 54: I think the idea is that these should be specific to the testing, and shouldn't be needed in the docs.

jburel, line 54: We have one link pointing to one the file (Apache24 support)
I can remove it and point to the omero-install instead so it is not included in the doc.

jburel:
Closing this PR, I will edit the file in the documentation

PR #1361, opened 2015-12-14
Some windows install fixes

dominikl:
Adds the instructions to explicitly install the python libraries for OMERO.web via *.whl files, like @manics  scripts does.


dominikl:
Please still hold off for a while. This didn't fix the problems either.

hflynn:
New commit reads fine to me

jburel:
read fine. Maybe some minor adjustment after tomorrow's rc2 review but I would think we should merge it now

hflynn:
Ok, merging now on the basis that a new PR can always be opened tomorrow if needed.

PR #1364, opened 2015-12-15
Update 5.1 ref on upgrade page

hflynn:
Spotted this when checking over docs for release, former wording was misleading, implying these are 5.1 docs.


hflynn:
Would be good for this whole page to be checked for the release while I have a PR open anyway.

kennethgillen:
:+1: for the "5.1 requires" fix - that made me think, and clarify in devteam, during Nightshade upgrade. Reading rest...

PR #1365, opened 2015-12-15
Updating Insight contributing page

hflynn:
Out of date Java is a quick fix, other quick fixes can be added, everything else can go on a Trello card for 5.2.2.

https://trello.com/c/XF44F6RN/396-update-build-instructions-contributing-docs

### DIFF
--- a/omero/developers/Insight/Contributing.txt
+++ b/omero/developers/Insight/Contributing.txt
@@ -38,34 +38,7 @@ is also set to UTF-8.
 Build system
 ^^^^^^^^^^^^
 
-Ant
-"""
-
-The compilation, testing, launch, and delivery of the application are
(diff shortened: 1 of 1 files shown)


jburel:
@hflynn: 
I will remove the section `ant` since the command indicated like `java build` will not work
and for now point to http://www.openmicroscopy.org/site/support/omero5.2-staging/developers/build-system.html
we can rework the actual `ant` section (we will change the title) in 5.2.2

hflynn:
Pushed, should be good to merge now.
I wonder if actually, rather than messing about with this for 5.2.2, there is any point in keeping this page longer term? Maybe we just delete it for 5.3 instead.

jburel:
@hflynn: probably delete it in 5.3
Merging

PR #1370, opened 2015-12-16
install-web changed "omero.webadmin" references

kennethgillen:
Removed omero.webadmin disambiguation where there is non required any more. Reading this as it stands, It seems like there's an additional "omero.webadmin" which is separate to the webclient.

### DIFF
--- a/omero/sysadmins/unix/install-web.txt
+++ b/omero/sysadmins/unix/install-web.txt
@@ -71,12 +71,6 @@ to access the OMERO.webclient:
 
    OMERO.web login
 
-.. note::
-	This starts the server in the foreground. It is your
-	responsibility to place it in the background, if required, and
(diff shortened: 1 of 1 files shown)


hflynn:
This issue occurs on the windows page too, do you want to fix that in this PR or would you prefer I open another?

ximenesuk:
Aside: @kennethgillen historically they used to be separate clients.

hflynn:
Fixed for windows page in linked PR

kennethgillen:
Is the note

> "This starts the server in the foreground. It is your responsibility to place it in the background, if required, and manage its shutdown." still applicable?   

It doesn't seem to have any context.

jburel:
It is out of context, remove it
`the manage shutdown` can be added as a note under Maintenance section

PR #1374, opened 2015-12-17
Exclude CI servers from the linkcheck in the contributing documentation

sbesson:
All these servers are currently under firewall and can be intermittently
unavailable either due to redeployment or major work. Running the linkcheck
on them is currently causing more noise than necessary.
/cc @hflynn


hflynn:
:+1: thanks! Build green again now

PR #1376, opened 2016-01-05
Minor fix

atarkowska:
To make clear that Django upgrade is required.

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -59,6 +59,13 @@ your previous version of OMERO. We recommend using a future-proof symlink
 if possible, so that these stanzas do not need updating with each OMERO
 server upgrade.
 
+.. note:: Since OMERO 5.2, the OMERO web framework no longer bundles
+    a copy of the Django package, instead manual installation of
+    the Django dependency is required. It is highly recommended to use
+    `Django 1.8`_ (LTS) which requires Python 2.7. For more information
+    see :ref:`python-requirements` on the
+    :doc:`/sysadmins/version-requirements` page.
+
 Web plugin updates
 ^^^^^^^^^^^^^^^^^^
 OMERO.web plugins are very closely integrated into the webclient. For this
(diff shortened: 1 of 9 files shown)


omero/sysadmins/server-upgrade.txt
```diff
@@ -59,6 +59,13 @@ your previous version of OMERO. We recommend using a future-proof symlink
 if possible, so that these stanzas do not need updating with each OMERO
 server upgrade.
 
+.. note:: Since OMERO 5.2, the OMERO web framework no longer bundles
```
hflynn: OMERO.web for consistency, otherwise this formats and reads well.

atarkowska: Do you want me to update across entire doc? This bit was copied over from another location.

mtbc:
See note at http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/server-upgrade.html#omero-web-configuration.

omero/sysadmins/server-upgrade.txt
```diff
@@ -59,6 +59,13 @@ your previous version of OMERO. We recommend using a future-proof symlink
 if possible, so that these stanzas do not need updating with each OMERO
 server upgrade.
 
+.. note:: Since OMERO 5.2, the OMERO.web framework no longer bundles
+    a copy of the Django package, instead manual installation of
+    the Django dependency is required. It is highly recommended to use
+    `Django 1.8`_ (LTS) which requires Python 2.7. For more information
+    see :ref:`python-requirements` on the
+    :doc:`/sysadmins/version-requirements` page.
```
mtbc, line 67: I'm wondering if we need separate links both to a section and to the page the section's on? Though if you think it adds value I'm fine with it.

mtbc:
I wonder if we (should) have a card noting where in docs we mention version numbers and other such stuff that might need routine updates as things change.

mtbc:
This PR seems fine to me to merge, just one possible quibble noted.

hflynn:
I have a card to replace version numbers with variables set in the conf.py already (https://trello.com/c/uripejjj/394-add-min-version-for-postgres-etc-to-conf-py) so that in future only that file will need updating. Hopefully get on to it for the next release or for 5.3 at the latest.

omero/sysadmins/server-upgrade.txt
hflynn, line 67: I can't say I think it hurts, readers are of course free to ignore hyperlinks. I think it's good to reference the page and by the time we've mentioned it, we may as well link too.

PR #1377, opened 2016-01-05
Example for importing images

dominikl:
This PR adds an example for importing images to the Java examples for developers.

Based on @bramalingam 's  ICY plugin code: https://github.com/bramalingam/OmeroICY/blob/master/TestPlugin/src/plugins/bramalingam/OMEROICY/testOmero.java#L128

### DIFF
--- a/omero/developers/Java.txt
+++ b/omero/developers/Java.txt
@@ -43,11 +43,12 @@ Connect to OMERO
 
 ::
 
-    LoginCredentials cred = new LoginCredentials();
-    cred.getServer().setHostname(hostName);
-    cred.getServer().setPort(port);
-    cred.getUser().setUsername(userName);
-    cred.getUser().setPassword(password);
+    LoginCredentials cred = new LoginCredentials(userName, password, host, port);
+
+    // Alternative using args array:
+    // args = new String[] { "--omero.host=" + hostName, "--omero.port=" + port,
+    //                "--omero.user=" + userName, "--omero.pass=" + password };
+    // LoginCredentials cred = new LoginCredentials(args);
 
     //Create a simple Logger object which just writes
     //to System.out or System.err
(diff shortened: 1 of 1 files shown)


hflynn:
Formats correctly on staging (https://www.openmicroscopy.org/site/support/omero5.2-staging/developers/Java.html) - I leave someone else to comment on the technical content.

hflynn:
Is this fulfilling https://trello.com/c/yXt1nOAr/237-python-upload-example or would further additions be needed?

joshmoore:
My primary thought looking at this is that for the general user this is more complicated than it needs to be, but code changes would be necessary to have something simpler/nicer.

dominikl:
@joshmoore Instead of using OMEROMetadataStoreClient I could alternatively create an example using the gateway, in particular the TransferFacility: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/blitz/src/omero/gateway/facility/TransferFacility.java#L73 But that's still a bit experimental; this method is covered by an integration test, but it's actually not used by Insight yet (although I copied it from Insight's `OMEROGateway` ). Would that be better?

joshmoore:
@dominikl : It might just be that we're in a transition phase. If `TransferFacility` is the new API, we could certainly work toward a production-grade example of that and then hide this example, or at least mark it as "low-level" or something.

jburel:
@dominikl: I agree that we should push the usage of the gateway and make example using gateway
Time to have an "extended" minimal-omero-client
e.g. minimal viewer, minimal importer

jburel:
Is the example run as part of training?

also I will be more explicit for the image imported or use fake.

dominikl:
No, good point, forgot about that.

dominikl:
Updated the doc to use the easier `LoginCredentials(String userName, ...)` respectively `LoginCredentials(String[] args0` constructor. As image placeholder `pathTo/image.dv` is used. I think that's actually quite clear, or should I really add the code for creating and using `.fake` image to the doc?

jburel:
I will not use `fake` in the doc, better to use a "real" format

jburel:
Thanks @dominikl

PR #1380, opened 2016-01-12
Fixing codahale metrics link

hflynn:
Link isn't loading so I've changed it to the home page. Alternatively, we could link to http://codahale.com/projects or directly to https://github.com/codahale/metrics - happy to switch to one of those if it's deemed more suitable.

### DIFF
--- a/omero/sysadmins/server-performance.txt
+++ b/omero/sysadmins/server-performance.txt
@@ -237,7 +237,7 @@ Useful metrics include:
       Time taken to generate tiled-pyramids for a big image. (PixelData-only)
 
 .. _Check_MK: https://mathias-kettner.de/check_mk.html
-.. _Codahale: http://metrics.codahale.com/
+.. _Codahale: http://codahale.com
 .. _JConsole: http://openjdk.java.net/tools/svc/jconsole/
 .. _JVisualVM: http://visualvm.java.net/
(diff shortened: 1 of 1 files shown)


hflynn:
Build green - https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-merge-docs/198/

manics:
Something under https://github.com/codahale/ looks better, though https://github.com/codahale/metrics says `This is not the Java library.` so it might be the wrong one.

hflynn:
https://github.com/dropwizard/metrics ?

hflynn:
Or actually http://metrics.dropwizard.io which looks like it is the actual replacement for the broken link

mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/server-performance.html#metrics. Would it be better to make "“Metrics” library" the actual link? Also it seems to be two words, "Coda Hale", it's a person, not sure if "Coda Hale's ..." is thus better.

hflynn:
Fix pushed and staging rebuilt - better now @mtbc ?

mtbc:
I'd remove "the", then it'll be perfect. :smiley:

hflynn:
Ah, good catch! Done

mtbc:
Good to merge once Travis turns green. :+1:

PR #1381, opened 2016-01-13
Document Tables.slice()

manics:
Also clarify Tables.read()

See https://trello.com/c/wZW74mw3/29-omero-tables-slice-isn-t-documented

### DIFF
--- a/omero/developers/Tables.txt
+++ b/omero/developers/Tables.txt
@@ -196,7 +196,8 @@ Main methods
 
     .. method:: slice(colNumbers, rowNumbers)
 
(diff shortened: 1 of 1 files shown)


simleo:
The docs build with no warnings and the html output looks fine. I'd replace "Read a subset of non-consecutive columns and rows" with "Read a subset of columns and rows (may be non-consecutive)" (same wording as above) to make sure we're not implying that the rows / columns _must_ be non-consecutive, but that's rather nit-picking.

Good to merge, optionally with the change discussed above.

manics:
@simleo Updated

simleo:
Good to merge.

PR #1384, opened 2016-01-17
Add section about coupled OMERO/Bio-Formats development

sbesson:
This section documents the development workflow while working on coupled
Bio-Formats/OMERO branches e.g. during breaking model changes.

Documents the workflow described in https://github.com/openmicroscopy/openmicroscopy/pull/4411

### DIFF
--- a/omero/developers/build-system.txt
+++ b/omero/developers/build-system.txt
@@ -248,3 +248,49 @@ it is put, and what role it plays in the distribution.
     By default, |OmeroCpp| are not built. Use ``build-all`` for that.
 
 These files are then zipped to OMERO.server-<version>.zip via ``release-zip``
+
+Coupled development
+-------------------
+
+Since OMERO 5.1.3, Bio-Formats is decoupled from the OMERO build system who
+consumes Bio-Formats artifacts from the OME Maven repository via Ivy_.
+
+While this decoupling matches most of the development use cases, it is
+sometimes necessary to work on coupled Bio-Formats and OMERO branches
+especially during breaking changes of the OME Data Model or the Bio-Formats
+API.
+
+The generic rule for coupled branches is to build each component in their
+dependency order and use the local Maven repository under :file:`~/.m2/repository` to share artifacts.
+
+Building Bio-Formats
+^^^^^^^^^^^^^^^^^^^^
+
+From the top-level folder of the Bio-Formats repository,
+
+#. adjust the version of Bio-Formats which will be built, installed locally
+   and consumed by OMERO e.g. for 5.2.0-SNAPSHOT::
+
+     $ ./tools/bump_maven_version.py 5.2.0-SNAPSHOT
+
+#. run the Maven command allowing to build and install the artifacts under the
+   local Maven cache::
+
+     $ mvn clean install
+
+Building OMERO
+^^^^^^^^^^^^^^
+
+From the top-level folder of the OMERO repository,
+
+#. adjust the ``versions.bioformats`` property under
+   :source:`etc/omero.properties` to the version chosen for the Bio-Formats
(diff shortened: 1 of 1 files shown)


omero/developers/build-system.txt
```diff
@@ -248,3 +248,49 @@ it is put, and what role it plays in the distribution.
     By default, |OmeroCpp| are not built. Use ``build-all`` for that.
 
 These files are then zipped to OMERO.server-<version>.zip via ``release-zip``
+
+Coupled development
+-------------------
+
+Since OMERO 5.1.3, Bio-Formats is decoupled from the OMERO build system who
```
joshmoore: I think of OMERO as a `which` rather than `who`, but considering Ms. Fair Boot and all ... :smile:

sbesson: I'll use `which` as this will avoid us the awkward `he/she` argument :)

omero/developers/build-system.txt
```diff
@@ -248,3 +248,49 @@ it is put, and what role it plays in the distribution.
...
+
+Building OMERO
+^^^^^^^^^^^^^^
+
+From the top-level folder of the OMERO repository,
+
+#. adjust the ``versions.bioformats`` property under
+   :source:`etc/omero.properties` to the version chosen for the Bio-Formats
+   build
+
```
joshmoore, line 289: What does "optionally" mean here? i.e. if I don't do this, there's some chance that `SNAPSHOT` items will come from artifactory rather than `~/.m2/repository`?

omero/developers/build-system.txt
hflynn: :+1:

omero/developers/build-system.txt
hflynn, line 289: Yeah, this is confusing. It could also be read as 'you need to do either 1. or 2.'

sbesson, line 289: Removed. This is adding confusion and the default resolver looks first into the local Maven repo before the OME artifactory. So forcing local resolution only should be unnecessary.

omero/developers/build-system.txt
```diff
@@ -248,3 +248,45 @@ it is put, and what role it plays in the distribution.
...
+Since OMERO 5.1.3, Bio-Formats is decoupled from the OMERO build system which
+consumes Bio-Formats artifacts from the OME Maven repository via Ivy_.
+
+While this decoupling matches most of the development use cases, it is
+sometimes necessary to work on coupled Bio-Formats and OMERO branches
+especially during breaking changes of the OME Data Model or the Bio-Formats
+API.
+
+The generic rule for coupled branches is to build each component in their
```
ximenesuk: generic -> general

omero/developers/build-system.txt
ximenesuk, line 289: On this, I had to explicitly set the resolver to `ome-resolver` when building with @mtbc's regions branch as that had a different resolver that didn't work for this workflow but presumably works for Mark. I don't want to complicate the docs here but I assume the correct resolver is set unless it has been changed locally. If there's any reason why the code in our repo might have a different resolver then some note may be useful, if not then it is fine as it stands.

sbesson, line 289: Fair enough, I can specify the resolver has to be `ome-resolver` (i.e. using local Maven resolution first).

sbesson:
Pushed.

ximenesuk:
Looks good to merge.

hflynn:
Should I hold off merging until the code PR is in, or are we good to go here @sbesson ?

hflynn:
Ah, I see the code is already cleared for merge. Thanks!

PR #1387, opened 2016-01-20
expand notes on inter-patch upgrade scripts

mtbc:
The current notes say too little about the inter-patch SQL upgrade scripts: this PR should at least improve the situation. Staged at https://www.openmicroscopy.org/site/support/contributing-staging/schema-changes.html.

@aleksandra-tarkowska may suggest further ways to improve the process.

--no-rebase as on contributing

### DIFF
--- a/contributing/conf.py
+++ b/contributing/conf.py
@@ -45,6 +45,7 @@ contributing_extlinks = {
     # Doc links
     'omero_doc' : (oo_site_root + '/support/omero/%s', ''),
     'bf_doc' : (oo_site_root + '/support/bio-formats/%s', ''),
+    'model_doc' : (oo_site_root + '/support/ome-model/%s', ''),
     }
 extlinks.update(contributing_extlinks)
 
--- a/contributing/schema-changes.txt
+++ b/contributing/schema-changes.txt
@@ -142,8 +142,8 @@ Review workflow
 
 Any Pull Request introducing a schema change will be initially marked and
 reviewed with a `breaking` label (see :doc:`ci-introduction`) so after
-opening your Pull Request please have it labeled as being breaking. This implies
-the Pull Request will first be included in the :ref:`omero_breaking` and
-:ref:`bf_breaking` exclusively. Once positively reviewed, the `breaking` label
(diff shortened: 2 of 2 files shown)


hflynn:
I have a possibly out of date card for updating this page - https://trello.com/c/6BUPmEP5/231-update-https-www-openmicroscopy-org-site-support-contributing-schema-changes-html Do you think it might benefit from further updates on top of this commit @mtbc ?

mtbc:
That's an interesting one: this page (about OMERO, not OME-XML) is basically covering the very tail end of the larger process (which isn't documented elsewhere, as far as I know). I could certainly add a couple of commits that make the relationship clearer.

hflynn:
Thanks, that would be helpful. I suspect the card may be aimed at documenting the OME-XML bit...

mtbc:
I think it is.

hflynn:
Well, if you want to volunteer to document your work on the model changes be my guest ;)

mtbc:
Maybe something to be taken time over some summer, I could probably piece much of it together from current PRs. (I'm hoping not to become _the_ model person!)

atarkowska:
@mtbc I think we could mentioned also OMERO5.0__0.sql and need of duplicating patch there as this is direct way of upgrade from the previous version.

atarkowska:
Also we do not say much about changing version number in the sql files (we only say change omero.properties). It has to be done in few places.

mtbc:
Those are good suggestions, thank you. Is it okay to save them for a card / later PR? One can figure them out from looking at the previous patch's scripts and CI will typically catch mistakes where local testing doesn't. I opened this PR more promptly because how to make the git commit history nice is rather less obvious and things will seem to work even if it's horrible.

hflynn:
@mtbc go ahead and save suggestions on a card if you don't have time to implement them all now, we certainly shouldn't hold up your Model work for this.

mtbc:
Noted on https://trello.com/c/9mDeCCUa/403-expand-notes-on-new-db-patches, thanks.

atarkowska:
thanks @mtbc

PR #1388, opened 2016-01-26
Adding note about VA being ditched

hflynn:
See https://trello.com/c/q14NpdkH/5-va-docs - no-one has downloaded the VA since we moved to the new downloads site in autumn 2013 so we aren't releasing it any more. This PR makes that clear for the 5.2 docs, for 5.3 they will be removed altogether.


mtbc:
Good to merge.

PR #1389, opened 2016-01-28
Fixing typo on homebrew install page

hflynn:
See https://trello.com/c/gtfC9e9F/405-bug-typo-in-osx-omero-install-instructions typo reported by Gus.

Will be staged at https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/unix/server-install-homebrew.html

### DIFF
--- a/omero/sysadmins/unix/server-install-homebrew.txt
+++ b/omero/sysadmins/unix/server-install-homebrew.txt
@@ -66,7 +66,7 @@ Install PostgreSQL database server::
 
     $ export LANG=${LANG:-en_US.UTF-8}
     $ export LANGUAGE=${LANGUAGE:-en_US:en}
-    $ brew install postgres
+    $ brew install postgresql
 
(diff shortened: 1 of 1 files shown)


jburel:
Note that the script under omero-install needs to be updated too

hflynn:
Ah, I wasn't sure if this came from there or not! Will open a PR there too.

sbesson:
Is the command not working on a given OSX version? In Homebrew, `postgres` is an alias of `postgresql` so the installation command is expected to work using either name. That being said, it does not hurt to use the extensive name.

jburel:
it did not work on el-capitan
so to avoid issue we should stick to the extensive name

sbesson:
Thanks, the change is certainly sensible, even more if it fixes an issue. I'll try to look at the Homebrew repository and cross-reference any relevant issues.

sbesson:
As per https://github.com/ome/omero-install/pull/85#issuecomment-176638902, ready to merge.

PR #1400, opened 2016-02-15
Removing old reference to controlling foreground/background as per #1370

kennethgillen:
Same text as removed in omero/sysadmins/unix/install-web.txt during https://github.com/openmicroscopy/ome-documentation/pull/1370


hflynn:
Thanks!

PR #1402, opened 2016-02-18
Adding unit & map annotation sample links

hflynn:
See https://trello.com/c/lIBeSsQy/93-new-2015-01-samples
OME-XML samples for map annotations and units were only recently added for the 2015/01 schema. This PR adds specific links indicating they are now available.

### DIFF
--- a/formats/developers/ome-units.txt
+++ b/formats/developers/ome-units.txt
@@ -26,6 +26,8 @@ more complete implementation.
     :omero_doc:`OMERO developer documentation on units <developers/Model/Units.html>`
     for further details of how this aspect of the Data Model is implemented in
     OMERO.
+    :omexml_downloads:`OME-XML samples <>` demonstrating the use of
+    instrument units.
 
 OME system of measurements
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


hflynn:
Will be staged at https://www.openmicroscopy.org/site/support/ome-model-staging/ once built

formats/developers/ome-units.txt
```diff
@@ -26,6 +26,8 @@ more complete implementation.
     :omero_doc:`OMERO developer documentation on units <developers/Model/Units.html>`
     for further details of how this aspect of the Data Model is implemented in
     OMERO.
+    :omexml_downloads:`OME-XML samples <>` demonstrating the use of
```
sbesson: Missing newline to separate this from the previous see also.

sbesson:
One formatting comment. Overall reads fine. Given the specificity of the sample links in this context, is it worth linking to the individual files directly rather than the samples directory?

hflynn:
I did wonder about just linking to the files. I'll do that and fix the formatting now.

hflynn:
Should be sorted now @sbesson

hflynn:
Live docs updated.

PR #1404, opened 2016-02-22
simplify model object glossary

mtbc:
This PR slightly simplifies the ginormous model object glossary at http://www.openmicroscopy.org/site/support/omero5.2-staging/developers/Model/EveryObject.html in two ways:
1. The list of which properties a class is "used by" now omits subclasses of those property-declaring classes. For instance, `AnnotationAnnotationLink` is used by `Annotation.annotationLinks` so we don't now bother also mentioning `BooleanAnnotation.annotationLinks`, etc.
2. We don't mention `details.permissions.perm1` because that is an internal implementation detail we would probably rather not have external code rely upon. With @joshmoore's work to extend the permissions system (e.g., download restrictions) it's probably now a rather incomplete view of what's permitted anyway.

Apart from those points, the glossary should remain as before.

### DIFF
--- a/omero/developers/Model/EveryObject.txt
+++ b/omero/developers/Model/EveryObject.txt
@@ -24,7 +24,6 @@ Used by: :ref:`LogicalChannel.mode <OMERO model class LogicalChannel>`
 
 Properties:
   | details.externalInfo: :ref:`ExternalInfo <OMERO model class ExternalInfo>` (optional)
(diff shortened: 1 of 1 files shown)


joshmoore:
> We don't mention details.permissions.perm1

:+1:

ximenesuk:
Looks good to merge.

PR #1405, opened 2016-02-22
sysadmins: Document hard-link semantic change for in-place import

ghost:
For https://trello.com/c/QYIXKFHK/82-hardlink-permissions-on-centos-ubuntu

/cc @manics @joshmoore

### DIFF
--- a/omero/sysadmins/in-place-import.txt
+++ b/omero/sysadmins/in-place-import.txt
@@ -279,22 +279,21 @@ If you are unclear about how hard-linking works, please see the
 `Hard link <http://en.wikipedia.org/wiki/Hard_link>`_ article on
 Wikipedia.
 
-.. note:: The semantics of hard-linking have changed recently on Linux
-    systems with the "protected hardlinks" feature, which is enabled
-    by default and is in use on Ubuntu 14.04, CentOS 7 and other
-    contemporary systems.  Linux now requires that when you create a
-    hard-link to a file that you are either the *owner* of the file,
-    or that you have *read/write access* to the file.  Other Unix
-    systems, and older Linux systems, allow a hard link to be made if
-    you have *search access* to the file (i.e. you have read/execute
-    permission on the directory containing the file), but do not check
-    the file permissions themselves.  See the `kernel-hardening
-    mailing list post
-    <http://www.openwall.com/lists/kernel-hardening/2012/02/21/20>`__
-    which describes the change in more detail.  The implication for
-    in-place import is that the user performing the import must own or
-    have read-write permissions on the data files being imported
-    in-place.
+The semantics of hard-linking have changed recently on Linux systems
+with the "protected hardlinks" feature, which is enabled by default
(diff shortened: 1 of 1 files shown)


hflynn:
This is quite a lot of content for a note box, I'm wondering if it would be better to just have it as a new paragraph of text.

omero/sysadmins/in-place-import.txt
```diff
@@ -279,6 +279,23 @@ If you are unclear about how hard-linking works, please see the
 `Hard link <http://en.wikipedia.org/wiki/Hard_link>`_ article on
 Wikipedia.
 
+.. note:: The semantics of hard-linking have changed recently on Linux
+    systems with the "protected hardlinks" feature, which is enabled
+    by default and is in use on Ubuntu 14.04, CentOS 7 and other
+    contemporary systems.  Linux now requires that when you create a
+    hard-link to a file that you are either the *owner* of the file,
```
mtbc: I'm doubtful about the double "that"; maybe start with the "when"?

omero/sysadmins/in-place-import.txt
```diff
@@ -279,6 +279,23 @@ If you are unclear about how hard-linking works, please see the
...
+    or that you have *read/write access* to the file.  Other Unix
+    systems, and older Linux systems, allow a hard link to be made if
```
mtbc: hard-link as a noun had a hyphen just above

omero/sysadmins/in-place-import.txt
```diff
@@ -279,6 +279,23 @@ If you are unclear about how hard-linking works, please see the
...
+    you have *search access* to the file (i.e. you have read/execute
+    permission on the directory containing the file), but do not check
+    the file permissions themselves.  See the `kernel-hardening
+    mailing list post
+    <http://www.openwall.com/lists/kernel-hardening/2012/02/21/20>`__
+    which describes the change in more detail.  The implication for
+    in-place import is that the user performing the import must own or
+    have read-write permissions on the data files being imported
```
mtbc: it was "read/write access" above and is "read-write permissions" here, not sure if that's a useful change

mtbc:
I'm not sure if they could possibly be attempting in-place from a system where client-side it _looks_ (due to network mount options) that they have the required permissions but in fact on the underlying filesystem they do not; can that situation arise for in-place? It's too early in the morning for me to know. But, even so, perhaps that's an awkward detail not worth adding. Also, I don't know if it's worth mentioning that symlinks should still work.

ghost:
The suggested corrections have been made.

@mtbc Regarding systems where the client-side looks correct, I think that's a bit out of scope--we can't do much if the system is behaving in a non-standard way.  They will need to determine for themselves if hardlinks work or not.

mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/in-place-import.html.

mtbc:
Thank you, good to merge.

PR #1413, opened 2016-02-25
Add basic import template page

ximenesuk:
This PR adds a basic page covering the recently added import templates, see https://github.com/openmicroscopy/openmicroscopy/pull/4393

### DIFF
--- a/omero/users/cli/import-target.txt
+++ b/omero/users/cli/import-target.txt
@@ -51,7 +51,7 @@ For example::
 The first case is equivalent to the previous example, the most recent Dataset
 will be used. In the second case the oldest Dataset will be used. In the third
 case the import should fail if multiple datasets with that name exist. In the
-first three cases a new Dataset will be created is none exists. In the last
+first three cases a new Dataset will be created if none exists. In the last
 case a new Dataset should be created even if one or more already exist.
 
 If the name contains spaces or other characters that cannot be used on the
@@ -77,7 +77,7 @@ expression. For example::
     $ bin/omero import ~/images/dv/SMN10ul03_R3D_D3D.dv -T "regex:^.*images/(?<Container1>.*?)"
 
 would use a Dataset with name being the path following ``images/``,
-ie just ``dv``.
+in this case ``dv``.
 
 The name discriminator can be explicitly used and, as in the previous section,
 also qualified::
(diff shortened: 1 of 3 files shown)


jburel:
`a new Dataset will be created is none exists`=> should be `if` none...

hflynn:
You need to add the new page to the toctree @ximenesuk 
(i.e. http://www.openmicroscopy.org/site/support/omero5.2/users/cli/index.html needs this file adding, I suggest between 'Import images' and 'Manage sessions')

jburel:
`ie` => `i.e.`

mtbc:
These question marks are interesting: it might be worth giving a couple more regex examples and/or telling us what kind of regexes are permitted. Basic POSIX.2 ones? Extended? Perl-compatible? (For instance, do I write `\S` or `[^[:space:]]` or what?)

ximenesuk:
Hmm, @mtbc I didn't really want to get the page bogged down in regex examples or detail. I did wonder if there was a suitable link that could be provided for the sort of user who might want to use this feature., i.e. to the type of regex allowed, which is effectively that used in `java.util.regex.*` /cc @joshmoore

mtbc:
Ah, it's the Java ones! Yes, even just some link to http://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html or suchlike might be useful, agreed that you don't want to teach people regexes on this page. (Or if Java happens to provide a superset of something sufficient and more portable, could just name that instead, then maybe we can switch engine without breaking the API.)

hflynn:
This is rebuilt on staging if you could confirm whether it's good to merge now @mtbc

mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/users/cli/import-target.html

(I suppose we need a major-release-template card for seeing if links with stuff like `javase/7/` need bumping when we bump minimum Java version.)

jburel:
@mtbc: Ok to merge it as it is?
so I can start the various builds tomorrow.

mtbc:
I don't know anything about import templates but @ximenesuk certainly addressed my concerns in his latest commit so I've no objections to merging.

ximenesuk:
The imports used here were effectively tested by @pwalczysko in the corresponding code PR, they are lifted from the testing instructions of that PR. So the code examples have been tested indirectly.

joshmoore:
Very nice. If we had inline manual support like `git cmd --help` I'd highly suggest putting this as part of the CLI itself. Until then a link to this page from the CLI might be useful.

NB: For the GUI importer, everything here is re-usable via the `ImportTarget` interface.

jburel:
Comment added to https://trello.com/c/ph9eC6QU/9-template-import-page-follow-up

Merging

PR #1415, opened 2016-02-26
Adding downloads refs to troubleshooting

hflynn:
This adds reference to the new experimental web configurations to the troubleshooting page since people upgrading are unlikely to revisit the installation docs.

The upgrade page probably needs some work too but that isn't going to happen in time for 5.2.2 unless there is a blocker.


hflynn:
2nd commit uses text from https://github.com/aleksandra-tarkowska/ome-documentation/commit/75bc809bcbc07b6db95588239ba5591473798cae that Ola missed when moving the content to the main troubleshooting page.

atarkowska:
looks good thx

PR #1418, opened 2016-03-03
Removing VA & demo docs

hflynn:
See https://trello.com/c/gKMVgoQ8/26-consider-replacing-virtual-applicance-with-docker and https://trello.com/c/cIJvElPd/401-pages-to-delete-for-5-3-when-we-don-t-mind-breaking-urls
The VA has been retired so these docs can go completely for the new major version release and the Demo server content is now replicated on Help so isn't needed in the docs too (again, have been holding off removing the URL until a new major release).

Will be staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/users/index.html once built.


sbesson:
Should we remove the corresponding images as part of this PR? Also does this mean we should consider putting redirects into place for the `users/demo-server.html` pages to the help documentation?

hflynn:
Bah, I forgot about the images, will delete them now.
A redirect certainly wouldn't hurt, I think I have all the plone links pointing to help already but I'll check.

hflynn:
Ah no, actually, the first link on the products page is going to the users version - yes, a redirect would be good (although I'll update the link now anyway).

hflynn:
Images should be all gone now too.

sbesson:
https://ci.openmicroscopy.org/job/OMERO-REGIONS-merge-docs/8/ works as expected and extra images are removed.

PR #1420, opened 2016-03-03
Remove outdated insight contributing page

hflynn:
See https://trello.com/c/cIJvElPd/401-pages-to-delete-for-5-3-when-we-don-t-mind-breaking-urls decision had already been made to remove this page for the next major release (when we don't mind breaking URLs) as it's a legacy from when insight was separate from the server code base and isn't needed any more.


hflynn:
https://www.openmicroscopy.org/site/support/omero5.3-staging/developers/index.html#insight

mtbc:
Good to merge. :+1:

The "our source is all Unicode" (both Java and Python actually) bit in the "Running code" section should probably migrate to some more general contributing docs; I could create a card.

hflynn:
I'll have a look for somewhere to stick it later, thanks.

PR #1417, opened 2016-03-02
upgrade dependences

atarkowska:
upgrading omero needs to remind about upgrading dependencies. This will make sure people install the latest Django version

PR #1421, opened 2016-03-03
Adding note re:web deps upgrade

hflynn:
See @jburel 's comment on https://github.com/openmicroscopy/ome-documentation/pull/1417

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -84,6 +84,9 @@ up to date to ensure that security updates are applied.
 
      $ pip install --upgrade -r share/web/requirements-py27-win.txt
 
+.. note:: This is mainly for new major releases. If carrying out a minor
+    release upgrade, you should check that updates will not break your set-up
+    before performing this on your production system.
 
 Web plugin updates
 ^^^^^^^^^^^^^^^^^^


jburel:
Thanks for adding the note, 
Better to be on safe side, I will also add a note the release template since we should check that the steps are fully tested/mentioned in announcement

hflynn:
https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/server-upgrade.html#omero-web-dependencies

atarkowska:
@jburel we test daily upgrade. I am guessing we should pick up any breaking changes straight away.

hflynn:
Up to you whether you want to merge this or not @jburel - I certainly don't see the harm

jburel:
As I indicated, it is better to be on the safe side. no harm adding the note in my opinion.
As we have seen with pillow last time (cap to 2.9), we did not catch the problem until we deployed on nightshade even if it is built and deployed daily.

atarkowska:
@jburel I am saying we are testing the latest version of web dependencies on ci, everyday virtualenv is updated with the latest version. if you wish I can add pillow, numpy and more... Then we will pick that up in the future

atarkowska:
My previous comment is not about the note block, just general comment. Note is definitely good idea.

jburel:
@aleksandra-tarkowska: ok I understood your comment was related to the note, we certainly need to discuss the upgrade strategy so we do not miss anything.
Thanks for taking care of it in the first place. 
Merging

PR #1433, opened 2016-03-29
update graphs migration documentation for OMERO 5.3

mtbc:
The legacy graphs requests are to be removed and the Java gateway methods have changed. Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/developers/Server/GraphsMigration.html.

--no-rebase

### DIFF
--- a/omero/developers/Server/GraphsMigration.txt
+++ b/omero/developers/Server/GraphsMigration.txt
@@ -1,5 +1,5 @@
-Using the new 5.1 graph requests
-================================
+Using graph requests
+====================
 
 Migration is required
 ---------------------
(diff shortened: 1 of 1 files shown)


hflynn:
While you're editing this page anyway can we have a title change? 'new 5.1' seems rather out of date now - just 'Using graph requests'?

hflynn:
Changes all look fine formatting and rendering-wise, I leave someone else to confirm the technical correctness.

I do wonder though if 'becomes' is still the right terminology in the examples - would it be clearer to new devs to clearly state 'where xxx was used prior to 5.1, now use' ?

dominikl:
The Java examples look fine. Maybe someone else wants to have a look at the Python ones?

mtbc:
Thank you. Python: no point, they're unchanged by this PR and unaffected by the Java gateway.

atarkowska:
Python examples match integration tests. Looks good to me.

PR #1437, opened 2016-03-30
Add windows deprecation warnings

hflynn:
See https://trello.com/c/tVSjHnee/432-add-warnings-for-windows-end-of-support-to-5-2-3-sysadmin-docs - as pointed out by @aleksandra-tarkowska - as well as updating the version requirements page, we should warn new sysadmins starting with 5.2 that Windows is best avoided.

I went with these pages on the assumption that would be enough for people to notice, but I can add the warning to all the windows installation folder pages if anyone thinks that is necessary.

### DIFF
--- a/omero/sysadmins/windows/install-web.txt
+++ b/omero/sysadmins/windows/install-web.txt
@@ -4,6 +4,12 @@ OMERO.web deployment
 .. Warning:: The guidance below applies to OMERO **5.2**. Code fixes for
     OMERO.web deployment on Windows have not been back-ported to the 5.1 line
     and this guide cannot be used to deploy OMERO.web 5.1.x on Windows.
+    
+    OMERO 5.2 is the last major version which will feature Windows
+    support for OMERO.web deployment. See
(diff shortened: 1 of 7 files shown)


mtbc:
Good to merge.

omero/sysadmins/windows/install-web.txt
```diff
@@ -4,6 +4,12 @@ OMERO.web deployment
 .. Warning:: The guidance below applies to OMERO **5.2**. Code fixes for
     OMERO.web deployment on Windows have not been back-ported to the 5.1 line
     and this guide cannot be used to deploy OMERO.web 5.1.x on Windows.
+
```
atarkowska, line 5: does it make sense to split warning on http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/windows/install-web.html and make windows removal first ?

atarkowska:
shouldn't that warning be on each http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/windows/index.html

omero/sysadmins/windows/install-web.txt
atarkowska, line 5: or at least make removal first

hflynn:
Should be sorted now @aleksandra-tarkowska

PR #1438, opened 2016-03-31
server-security: client default is to drop encryption after authentication

carandraug:
Current documentation suggests that the default client behaviour is to keep encryption after the authentication with the possibility of having disabling it for speed up. It's actually the opposite (although this is not straightforward since it's also dependent on server configuration).


hflynn:
Staged on https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/server-security.html ready for review by the team. Thanks @carandraug

atarkowska:
Note that we will remove the lock in 5.3 https://trello.com/c/sc79fkLw/58-drop-lock-icon-in-web

joshmoore:
:+1:

PR #1439, opened 2016-03-31
nginx version

atarkowska:
update nginx version requirements

cc @rleigh-dundee 

For review: https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/version-requirements.html#nginx

### DIFF
--- a/omero/sysadmins/version-requirements.txt
+++ b/omero/sysadmins/version-requirements.txt
@@ -187,8 +187,8 @@ The planned changes for 5.2 depend upon our support of
 
 * nginx
 
-  * [5.2] 1.0 dropped, 1.2 deprecated, 1.4 deprecated, 1.6 supported, 1.8 or 1.9 recommemded
-  * [5.3] 1.2 dropped, 1.4 dropped, 1.6 supported, 1.8 or 1.9 recommemded
+  * [5.2] 1.0 dropped, 1.2 deprecated, 1.4 deprecated, 1.6 supported, 1.8 or 1.9 recommended
+  * [5.3] 1.2 dropped, 1.4 dropped, 1.6 supported, 1.8 or 1.9 recommended
   * Rationale: 1.4 is the Ubuntu 14.04 and Debian 8 version.
     Removed for 5.3 accordingly.
 
--- a/omero/sysadmins/version-requirements.txt
+++ b/omero/sysadmins/version-requirements.txt
@@ -187,10 +187,10 @@ The planned changes for 5.2 depend upon our support of
 
 * nginx
 
-  * [5.2] 1.0 dropped, 1.2 deprecated, 1.4 supported, 1.6 recommemded
-  * [5.3] 1.2 dropped, 1.4 deprecated, 1.6 recommemded
+  * [5.2] 1.0 dropped, 1.2 deprecated, 1.4 deprecated, 1.6 supported, 1.8 or 1.9 recommemded
+  * [5.3] 1.2 dropped, 1.4 dropped, 1.6 supported, 1.8 or 1.9 recommemded
   * Rationale: 1.4 is the Ubuntu 14.04 and Debian 8 version.
-    Deprecate for 5.3 accordingly.
+    Removed for 5.3 accordingly.
 
 Support levels
 ==============
@@ -1728,13 +1728,6 @@ Distribution support
       - Debian
       - Homebrew
       - FreeBSD Ports
-    * - 1.0
-      - N/A
-      - N/A
-      - 12.04 (1.1)
-      - N/A
-      - N/A
-      - N/A
     * - 1.2
       - N/A
       - N/A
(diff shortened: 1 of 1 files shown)


jburel:
Just a heads up, we will need to check the version installed in `omero-install`. so we do not end up having a mismatch between version dropped and what we have in the installed doc e.g. postgres on ubuntu.

atarkowska:
@jburel I think we install distribution packages/download from nginx website, version is provided by them, do we hardcode any versions?

jburel:
We don't hardcode version but in the case of postgres we had 9.3 dropped in our list and 9.3 is the version installed  when using the distribution package on ubuntu

jburel:
typo: recommemded => recommended

jburel:
Adjustment made to https://github.com/ome/omero-install/pull/94 to match changes in the doc.
cc @hflynn

omero/sysadmins/version-requirements.txt
```diff
@@ -1750,12 +1743,26 @@ Distribution support
...
+      - N/A
+      - N/A
+      - N/A
+      - N/A
+      - Yes
+      - Yes
+    * - 1.9
+      - N/A
+      - N/A
+      - 14.10
```
ghost, line 1762: Are the Debian and Ubuntu versions here correct?

hflynn, line 1762: Who can answer this?

joshmoore, line 1762: Is the problem _which_ Ubuntu is being looked at? @aleksandra-tarkowska : what did you check against?

atarkowska, line 1762: on the bottom of the table https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/version-requirements.html#nginx there are REF links, see there

atarkowska, line 1762: sorry I think I am not getting this table right, 1.9 in general should not be there, as it is develop

aleksandra-tarkowska: do we need to show 1.4 and 1.2 ? 1.6 is legacy version already and is the lowest installed by default

atarkowska, line 1762: do we need to show 1.2 ? 1.4/1.6 is legacy version already and is the lowest installed by default

jburel, line 1762: note that we have now installed 1.8 on all OS in omero-install
For Ubuntu14.04, the "default" version installed is 1.4
if we wish to support < 1.8 we will have to introduce a new parameter

hflynn:
Does this PR need further work (e.g. in light of https://trello.com/c/FXt1rXQl/1-scoping-team) or is it ready to be relisted for review?

jburel:
the omero-install fixing the version has already been merged. In our doc 1.6 installation will show up instead of 1.4 for ubuntu.

hflynn:
Looks like we have a mismatch between the versions listed in the 2 tables under https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/version-requirements.html#nginx - first table version 1.0->1.6 and 2nd 1.2->1.9. Looks like we could at least remove 1.0 from the first table and add 1.8 there if not 1.9 as well.

jburel:
@aleksandra-tarkowska: could you check the version mismatch?

atarkowska:
Do we really need that table? as @kennethgillen just pointed 1.10 was released, I think they are changing too fast

atarkowska:
I would only keep 

```
[5.2] 1.0 dropped, 1.2 deprecated, 1.4 deprecated, 1.6 supported, 1.8 or 1.10 recommended
[5.3] 1.2 dropped, 1.4 dropped, 1.6 supported, 1.8 or 1.10 recommended
```

or even 1.8+

hflynn:
Removing the last table would be out of keeping with the rest of the page, I would just cut it down to show 1.6+. i'd also favour using 1.8+ as the recommended version, we don't want to start referring to 1.10 as recommended when we haven't tested it - we can always update again when 5.3 is closer if we've tested it by then.

atarkowska:
but 1.6+ (1.8+) means 1.10, sorry I am confused now

atarkowska:
please also keep in mind that omero-install install now 1.10

https://travis-ci.org/ome/omero-install/jobs/126070807#L4216

hflynn:
Ok, i didn't know that.

I would make the last table 1.6, 1.8, 1.9 and 1.10 in that case. And update the table above it as you suggested in your comment before mine.

atarkowska:
1.9 is not what you want as this is develop

hflynn:
ok, 1.6, 1.8 and 1.10 then

atarkowska:
my point was only that the way we recommend to install nginx in our doc (based on omero-isntall) it will install the latest. This table is out of date once new release comes out If this has to stay I would just make 1 line 1.6+ not 3 for each version

hflynn:
Ok, do that then.

omero/sysadmins/version-requirements.txt
ghost, line 1762: @aleksandra-tarkowska All the versions should be there; this table is firstly a list of the upstream releases of a project, and secondly showing our level of support for each version.  For other components on this page, we have dropped older irrelevant releases, and for this component everything before 1.6 sounds fine.  But having all the later ones there is necessary, and given that the latest Ubuntu LTS is using 1.9, it needs to be there as well.

ghost:
@aleksandra-tarkowska 1.6+ is not an appropriate way to represent things for these tables.  They document the start and end of life of each upstream release, and our support for each major release and condensing the information in this way would defeat the purpose of having the information in this form to begin with; please do look at the other projects on the page.

hflynn:
Can we get the information for nginx correct for the 5.2.3 release and then discuss the future of this page via https://trello.com/c/FXt1rXQl/1-scoping-team please?

hflynn:
Closing in favour of #1460

PR #1440, opened 2016-03-31
DropBox Unicode configuration section

ximenesuk:
This PR adds a short section to the docs advising on how to customise the PYthon configuration to deal with Unicode charactewrs in path and file names.
See: https://trello.com/c/lFY8DI8Z/20-unicode-in-upgrade

### DIFF
--- a/omero/sysadmins/dropbox.txt
+++ b/omero/sysadmins/dropbox.txt
@@ -146,6 +146,10 @@ For more details on using customization files in Python see:
 For more discussion on this issue within OMERO see the forum post:
 `Dropbox halts on certain unicode characters <https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7810#p15910>`_.
 
+.. note::
+    If a customization file is used and the OMERO server is upgraded please
+    ensure the file is still available to DropBox after the upgrade.
+
 Advanced use
 ------------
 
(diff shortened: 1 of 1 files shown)


mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/dropbox.html#unicode-path-and-file-names.

ghost:
The text and HTML page look fine and good to merge.

The only question I have here is why this isn't the default behaviour?  It seems like having it able to process Unicode filenames is an expected default, not a special-case customisation.

ximenesuk:
@rleigh-dundee It isn't the default because @joshmoore and I battled long and hard trying various solutions without success. This local customisation was the only thing that worked without breaking other things. An eventual move to Python 3 would remove the problem but for now this seemed to be the best fix without major changes to the code.

ximenesuk:
@joshmoore are you okay with this being merged?

joshmoore:
Happy for it to be merged. Thinking back, I already can't remember the various conditions under which it happens/happened. If you remember something like "when locale=X,..." that may be useful to include so that users don't assume they _always_ have to make this setting.

ximenesuk:
@joshmoore I've been unable to find anything obvious in the various forum threads and cards that might help here.

joshmoore:
Fair enough.

PR #1441, opened 2016-04-01
add a graph requests overview for client developers

mtbc:
This page offers a brief overview of the full set of graphs requests included for the OMERO 5.3 branch. It should eventually be expanded with examples, perhaps inheriting some from the  graphs migration page when that is retired, but at least it is a start. Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/developers/Server/GraphRequests.html.

### DIFF
--- a/formats/developers/using-ome-xml.txt
+++ b/formats/developers/using-ome-xml.txt
@@ -72,7 +72,7 @@ Things to avoid
    Direct copying of our structure is not necessary as XML already has a
    method of including items from another schema, **see below**. From
    our point of view this is important as it allows us to control which
-   parts of our model can be used as stand alone objects in others' work.
+   parts of our model can be used as stand-alone objects in others' work.
 
 Correct approaches
 ------------------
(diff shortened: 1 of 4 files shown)


mtbc:
Needs the 5.3 slice docs staged to fill in the broken links, cc: @sbesson who probably knows what to do. (Is there a magic "ignore this broken link" rune I should add?)

sbesson:
@mtbc: we could ignore the broken link by extending the `linkcheck_ignore` list in `conf.py`. Since we have latest 5.3 API redirects in place, we could also knowingly "break" the build as we have done in other occasions /cc @hflynn. Either way, we'll want to make sure these links are fixed when we tag OMERO 5.3.0-m2 as the slice API documentation should be fixed. I will open a PR to point to the latest omero5.3 redirects shortly.

sbesson:
Briefly discussed with @mtbc who suggested a better workaround: opening a PR adding these API HTML links to `linkcheck_ignore` and keeping this PR open and `ON_HOLD` until 5.3.0-m2 is tagged and the new API is copied then close it and check the build remains green.

joshmoore:
Looks good. I have some slight concern that the listing here will get out of sync with the javadocs, but until we feel comfortable with the javadocs as the/an official landing page, :+1:

PR #1442, opened 2016-04-01
Fix prolog and use correct latest OMERO redirects

sbesson:
This should allow the staging OMERO 5.3 documentation pages to link to the OMERO 5.3 milestones. /cc @mtbc

### DIFF
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -46,8 +46,8 @@ else:
     release = 'UNKNOWN'
 
 rst_prolog = """
-.. note:: **This documentation is for the new OMERO 5.2 version.** See the `latest OMERO 5.1.x
-    version <http://www.openmicroscopy.org/site/support/omero5.1/>`_ or the
+.. note:: **This documentation is for the new OMERO 5.3 version.** See the `latest OMERO 5.2.x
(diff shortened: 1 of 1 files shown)


mtbc:
Ah, this does seem to have changed my links to now point at http://downloads.openmicroscopy.org/omero/5.3.0-m1/api/slice2html/omero/cmd/GraphQuery.html and suchlike.

mtbc:
In the longer term I am not sure if we need to note to revert some of this or if we should use some variable that everywhere becomes `5.3` for this branch.

mtbc:
Has the intended effect. :+1: Planning to merge this or hold it open? I'm good either way.

sbesson:
There are some remaining `5.2` instances across the documentation which are meant to be replaced by `5.3` but  https://trello.com/c/PEvlrmHX/36-document-breaking-api-changes-for-devs-add-new-examples and similar cards should force us to review their content. Can leave this PR open for now and have the discussion about the variable substitution.

sbesson:
Since https://github.com/openmicroscopy/ome-documentation/pull/1443 who was included earlier depended on this PR, merging this.

PR #1443, opened 2016-04-04
avoid link checking for Blitz API requests

mtbc:
Allow docs to refer to Blitz API docs not yet staged; see https://github.com/openmicroscopy/ome-documentation/pull/1441#issuecomment-205184417. To be closed once `5.3.0-m2` is tagged.

PR #1445, opened 2016-04-05
adjust for DiskUsage being renamed to DiskUsage2

mtbc:
trivial fix, though link will be broken until m2


joshmoore:
as https://github.com/openmicroscopy/openmicroscopy/pull/4551, :thumbsup:

sbesson:
Should we follow the same strategy as in https://github.com/openmicroscopy/ome-documentation/pull/1443 and include another commit that needs to be reverted post m2 tagging?

mtbc:
This one should already be covered by the regular expression in #1443.

PR #1446, opened 2016-04-06
rewrite how to access information on Quartz cron expressions

mtbc:
Unfortunately the version of Quartz we use is so ancient that one must now dig somewhat to find documentation. Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/config.html.


dominikl:
Good to merge :+1:

PR #1450, opened 2016-04-12
Delete old unused images

hflynn:
When deleting the windows section I noticed that the image folder looked like it needed some cleaning up. This PR deletes all the image files that weren't referenced on any of the doc pages when I grepped for them. Review just requires checking builds are green.


sbesson:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/306 passed without warnings/errors.

PR #1452, opened 2016-04-20
remove misleading regex metacharacters from linkcheck_ignore

mtbc:
Removing the leading `^` so as not to give the impression that partial string matches could work. `linkcheck_ignore` does whole-string matching so this PR should not change any behavior.


hflynn:
Is there anything to test here aside from builds being green?

hflynn:
https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-merge-docs/311/ green with this included

hflynn:
https://ci.openmicroscopy.org/view/Docs/job/CONTRIBUTING-merge-docs/746/ also green

mtbc:
Thank you.

PR #1455, opened 2016-04-20
Pillow version

jburel:
- Remove pillow version in text
- Do not remove it from files in `/unix/walkthrough` 
  This is taken care of by omero-install


hflynn:
Looks good thanks

PR #1458, opened 2016-04-25
add clarification about multi-major-version upgrade workflow

mtbc:
See https://trello.com/c/YsWiFNmq/70-skipping-major-versions -

> Working under the assumption that the runtime tweaks will always be reflected in the upgrade scripts.

Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/server-upgrade.html#run-the-upgrade-script.

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -241,6 +241,11 @@ to run.
     (1 row)
 
 
+If you are upgrading from a server earlier than |previousversion| then
+it suffices to run the earlier upgrade scripts in sequence before the
+one above. There is no need to download and run the server from an
+intermediate major release.
+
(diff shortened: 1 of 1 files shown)


hflynn:
👍 for the use of the previous version auto-replacement. 

Renders fine on staging, I just wonder whether it might be good to signpost this a bit more? Although I guess we need to trust people to read this page properly anyway.

mtbc:
I think it's probably something many people assume to be true anyway and if they assume the opposite nothing will break, they'll just waste some time.

hflynn:
Ah ok, good to merge then I think, unless @joshmoore has any objections.

joshmoore:
My guess is that people will miss it, but we now have something to point to. :+1:

PR #1460, opened 2016-04-27
Nginx version

hflynn:
This replaces #1439 as Ola is too busy to follow up. Staged at https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/version-requirements.html#nginx

I've hopefully followed the comments re: what we support for 5.2.3 now. I couldn't make head nor tail of the Ubuntu and Debian support refs so @rleigh-dundee perhaps you'll check and let me know if the distribution support table needs updating.

### DIFF
--- a/omero/sysadmins/version-requirements.txt
+++ b/omero/sysadmins/version-requirements.txt
@@ -187,8 +187,8 @@ The planned changes for 5.2 depend upon our support of
 
 * nginx
 
-  * [5.2] 1.0 dropped, 1.2 deprecated, 1.4 deprecated, 1.6 supported, 1.8 or 1.9 recommemded
-  * [5.3] 1.2 dropped, 1.4 dropped, 1.6 supported, 1.8 or 1.9 recommemded
+  * [5.2] 1.0 dropped, 1.2 deprecated, 1.4 deprecated, 1.6 supported, 1.8 or 1.9 recommended
+  * [5.3] 1.2 dropped, 1.4 dropped, 1.6 supported, 1.8 or 1.9 recommended
   * Rationale: 1.4 is the Ubuntu 14.04 and Debian 8 version.
     Removed for 5.3 accordingly.
 
--- a/omero/sysadmins/version-requirements.txt
+++ b/omero/sysadmins/version-requirements.txt
@@ -187,10 +187,8 @@ The planned changes for 5.2 depend upon our support of
 
 * nginx
 
-  * [5.2] 1.0 dropped, 1.2 deprecated, 1.4 deprecated, 1.6 supported, 1.8 or 1.9 recommended
-  * [5.3] 1.2 dropped, 1.4 dropped, 1.6 supported, 1.8 or 1.9 recommended
-  * Rationale: 1.4 is the Ubuntu 14.04 and Debian 8 version.
-    Removed for 5.3 accordingly.
+  * [5.2] 1.2 deprecated, 1.4 deprecated, 1.6 supported, 1.8+ recommended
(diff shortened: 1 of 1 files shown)


hflynn:
I notice now that 5.1 has been removed from the summary sections but kept in the detailed tables, I might remove it from all the tables tomorrow if I get time.

ghost:
@hflynn It's a historical record for showing the change in support status over time, which is useful for a sysadmin when planning an upgrade.  It shows what changed.  Is having it there a problem?

The support status lines are incorrect.  They should be:
5.1: Supported/Recommended/Unsupported/Unsupported
5.2: Deprecated/Supported/Recommended/Unsupported
5.3: Dropped/Supported/Supported/Recommended

This is because
- we should only have one recommended version for each release
- we can't retrospectively make something recommended a year or two after its release--it's documenting what was supported at that time, so e.g. 5.1 can't have 1.8 and 1.10 recommended since both 1.8 and 1.10 _didn't even exist_ at that point in time.

In the distribution support table:
1.10 should have Ubuntu (16.04) adding
FreeBSD provides 1.8; 1.6 should be N/A
Homebrew provides 1.8; 1.6 should be N/A

hflynn:
Fair enough re: 5.1 but Ola told me that omero-install is now using 1.10 for new 5.2 installs so how can we say 1.10 is unsupported for 5.2?

ghost:
We can certainly do so if we do support it.  It's definitely not something which should be "Recommended" though--it's been out for a grand total of _two days_, so has not had any serious testing.  We can recommend it for 5.3 once it's had more testing.  I'd be inclined to keep 1.8 recommended for both for the moment though.

hflynn:
The 5.3 items are our current plans so I think it's fine to say we expect to be recommending 1.10 at this point. I'll fix the rest as you suggest though.

hflynn:
Hopefully fixed as far as we are able for the 5.2.3 release. I propose we update this page following the outcome of https://trello.com/c/FXt1rXQl/1-scoping-team-req and push a patch release of the docs to update the community then so they can prepare in advance of the 5.3 release.

jburel:
cf. https://github.com/openmicroscopy/ome-documentation/pull/1460#issuecomment-215362262
1.10: I would like to double-check the ubuntu case. I don't think this is the case

jburel:
on ubuntu 14.04 nginx 1.8 will be installed.

hflynn:
The link Ola gave was Centos7 installing 1.10

hflynn:
But I can change the 5.2 status to 1.10 unsupported if we are concerned.

ghost:
@jburel See http://packages.ubuntu.com/search?keywords=nginx&searchon=names&suite=all&section=all (trusty).  It's 1.4.6 as noted in the table for 14.04.

hflynn:
@rleigh-dundee JM is referring to what version our installation walkthrough for Ubuntu installs not what the default is for that distribution.

ghost:
@hflynn OK.  But "what's in our installation scripts" is a rather different question than "what's provided by the distributions", and "what are the minimum requirements", and is not what this page is about.

hflynn:
My point is that our recommended versions and what we are installing via our scripts should match. The distribution support documented here is as you recommended.

ghost:
OK.  I would however want to flag up why we are deliberately deviating from the version provided by the distribution though, including its security support.  I would hope what we are installing is properly supported and there's a good reason for deviating.

hflynn:
That is a discussion for the web team

hflynn:
Taken a middle ground and changed 1.10 status to 'upcoming' for 5.2. Can we all agree these changes are ok for the 5.2.3 release now?

jburel:
sounds reasonable thanks @hflynn

PR #1464, opened 2016-05-03
Add item

jburel:
This will have to be reviewed later on.cc @hflynn 
This should fix the issue


hflynn:
Ok, now you've changed the script, it looks like it's not looking for the requirements file after all so it's now failing because you've added it to a toctree! Can't win! I'll try with just the script change and see if that works

PR #1473, opened 2016-05-06
Merge develop into regions

hflynn:
Merge develop into regions to fix conflicts


sbesson:
Works as expected. Will merge once Travis is green. Thanks

PR #1482, opened 2016-05-16
Manual link fix

hflynn:
This would usually be fixed on the code side but there's no infrastructure set up to do so given we are hoping not to have another 5.2.x release so I've just manually fixed it in the docs page instead.

Staged on http://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/config.html#scripts


hflynn:
https://ci.openmicroscopy.org/view/Docs/job/OMERO-5.2-merge-docs/4/ is green

PR #1483, opened 2016-05-17
Move more content to sysadmin main index page

hflynn:
See https://trello.com/c/kdJKkKy0/431-restructure-sysadmin-installation-docs - this takes advantage of the fact we now only have one set of installation docs to re-organize the sysadmin main index page and move more of the content here rather than buried under sub-index pages.

I decided not to get rid of the unix subdirectory in the end because it would require a load of redirects for the sake of removing 5 characters from the URL. I just ditched the index page because AFAIK aware external links were generally going to the main server installation page anyway.

Might still ditch the install-web subdirectory but waiting to see outcome of scoping.

Staged on https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/index.html


manics:
There's no harm in having subdirectories, no-one ever types them in directly and it makes it easier if we add more sections in future.

ghost:
This looks fine.  Please merge.

PR #1485, opened 2016-05-18
Add min versions for Java & Postgres & a bunch of sysadmin doc updates

hflynn:
This addresses https://trello.com/c/FmCsmVP7/208-pixelsservice-resolution-order-first-pixels-then https://trello.com/c/xuN70hl2/430-sysadmin-docs-should-cover-ulimits-for-files-and-processes and https://trello.com/c/uripejjj/394-add-min-version-for-postgres-etc-to-conf-py to clean up a bunch of sysadmin docs tasks that have been hanging around too long. Also a bit of clean up on the users index page. 

Staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/users/index.html and https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/index.html and various pages linked.

### DIFF
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -71,6 +71,7 @@ rst_epilog += """
 .. |current_dbver|  replace:: %s
 .. |previous_dbver|  replace:: %s
 .. |iceversion| replace:: 3.5.1
+.. |postgresversion| replace:: 9.4
 """ % (previousversion, conf_autogen.current_dbver,
        conf_autogen.previous_dbver)
 
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -72,6 +72,7 @@ rst_epilog += """
 .. |previous_dbver|  replace:: %s
 .. |iceversion| replace:: 3.5.1
 .. |postgresversion| replace:: 9.4
+.. |javaversion| replace:: 1.7
 """ % (previousversion, conf_autogen.current_dbver,
        conf_autogen.previous_dbver)
 
--- a/omero/sysadmins/troubleshooting.txt
+++ b/omero/sysadmins/troubleshooting.txt
@@ -456,3 +456,8 @@ http://bugs.sun.com/bugdatabase/view\_bug.do?bug\_id=4751177 or this
 :ome-users:`ome-users thread <2009-March/001465.html>` on our mailing list for
 more information.
 
+Data corruption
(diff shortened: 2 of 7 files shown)


mtbc:
If PostgreSQL version 9.4 or later is required then also update `sysadmins/version-requirements.txt`?

hflynn:
We have 9.3 marked as deprecated already @mtbc it was the postgres page that hadn't been updated not version-requirements

mtbc:
Surely "9.3 deprecated, 9.4 recommended" is inconsistent with "9.4 required"? I may simply be just completely misunderstanding what's going on here though!

hflynn:
We already had this state of affairs after @jburel 's update the postgres page for 5.2.3, I can amend the wording to 9.4 is recommended if you prefer though, to my knowledge, 9.3 isn't actually broken.

manics:
If anything I vote for dropping 9.3, especially since 9.5 is out

mtbc:
@hflynn: would much prefer "recommended", thanks, unless @manics manages to get 9.3 dropped!

hflynn:
Happy to make 9.3 dropped on the requirements page instead but that's not a docs decision - I'm adding a comment to the postgres card on the requirements scoping list -https://trello.com/c/A4nhbaE8/8-have-sql-upgrade-scripts-require-pg-9-3

jburel:
technically it works with pg 9.3. This is the default version installed on ubuntu 14.04
As @hflynn pointed out during 5.2.3, we have adjusted omero-install and the generated walkthrough so they, at least, match what was mentioned as recommended/dropped etc. (omero-install scripts only install 9.4 and 9.5)
The page is confusing for many people (topic for another day)

The changes made in this PR look fine

hflynn:
One last commit to make clear that Java 7 is **required** but Postgres 9.4 is only recommended @mtbc

mtbc:
Thank you. :smiley:

hflynn:
Staging rebuilt and everything green, going to merge tomorrow to get this out of the way unless there are any vetoes by then.

hflynn:
No vetoes received, merging. We will no doubt be revisiting version issues before 5.3.0 anyway, I still need to implement splitting up the page and simplifying it.

hflynn:
--rebased-to #1504

PR #1488, opened 2016-05-26
Define schema documentation alias

sbesson:
As part of the schema release process review, a proposal is to have the oXygen generated documentation served under a more canonical URL. This PR updates the Model and Formats documentation set to define an extlink alias for the schema documentation simplifying future updates.

To test this PR, check the build passes and the link checking is passing

Update: as this discussion turned into a debate about documentation URLs, I proposed to stick to the old URL for testing the [schema release workflow](https://trello.com/c/zkbHhnNO/31-schema-release-process), postpone the new URL decision and give us more time to include all relevant topics including [OME FIles documentation](https://trello.com/c/01I3lWQa/3-discuss-ome-files-docs) and plone migration.

### DIFF
--- a/formats/conf.py
+++ b/formats/conf.py
@@ -42,6 +42,7 @@ model_extlinks = {
     # Doc links
     'omero_doc' : (oo_site_root + '/support/omero/%s', ''),
     'bf_doc' : (oo_site_root + '/support/bio-formats/%s', ''),
+    'schema_doc' : (oo_root + '/schema_doc/' + release + '/%s', ''),
     # Downloads
     'bf_downloads' : (downloads_root + '/latest/bio-formats/%s', ''),
     'image_downloads' : (downloads_root + '/images/%s', ''),
--- a/formats/conf.py
+++ b/formats/conf.py
@@ -42,7 +42,7 @@ model_extlinks = {
     # Doc links
     'omero_doc' : (oo_site_root + '/support/omero/%s', ''),
     'bf_doc' : (oo_site_root + '/support/bio-formats/%s', ''),
-    'schema_doc' : (oo_root + '/schema_doc/' + release + '/%s', ''),
+    'schema_doc' : (oo_root + '/Schemas/Documentation/Generated/OME-' + release + '/%s', ''),
     # Downloads
     'bf_downloads' : (downloads_root + '/latest/bio-formats/%s', ''),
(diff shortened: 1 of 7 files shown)


ghost:
Why the underscore rather than a hyphen?  And why not `doc/schema/year` so the path can be used for other docs, or even `/site/support/schema/year` to fit in with the other docs?

sbesson:
I mostly used `schema_doc` as I browsed what others had done and found e.g. http://www.crossref.org/help/schema_doc/4.3.6/AccessIndicators_xsd.html but this PR is completely open to alternate suggestions. I think `/site/support/*` is too much related to the plone structure of our website and I do not see us fitting this documentation as part of it. For `doc/schema/year`, you would see e.g. http://www.openmicroscopy.org/doc/schema/2015-01/ome.html right? My only question about that is whether anything else is meant to be under `doc/`.

ghost:
Yes, `/doc/schema/2015-01/ome.html` was the pattern I was thinking of.

With regard to `/site/support/`, I've always hated the mixing of the static plone content and generated sphinx content.  I'd be strongly in favour of moving all the non-plone documentation under /doc if we had the option of doing so.

With regard to this PR, the changes look fine bar any outstanding naming considerations.

hflynn:
I think moving everything is only going to happen when we ditch plone @rleigh-dundee 
Given we don't have any other content under /doc/ I'd prefer the /schema_doc/
As far as /support/ goes, we already have http://www.openmicroscopy.org/site/support/ome-model/schemas/ so i think /support/schema/ would get confusing.

ghost:
Why the preference for `schema_doc` when `doc` would provide flexibility for the future though?

sbesson:
Another conceivable proposal along the same lines as @rleigh-dundee's would be to start using a dedicated domain for versioned documentation sets, i.e. http://docs.openmicroscopy.org/schemas/2015-01/ (and eventually http://docs.openmicroscopy.org/software/version/ mirroring http://downloads.openmicroscopy.org/software.version)

hflynn:
A dedicated domain is probably the way to go in the future but I'd rather we did it when we can move everything there at once rather than piecemeal. I'm also loathe to start deciding future URLs for everything when we haven't decided how to structure the site post-plone yet.

hflynn:
Merging this for now as I'm about to start further work on these docs. Now the alias is properly defined, we can just amend the URL in the config when we make a decision.

PR #1491, opened 2016-06-08
Adding June 2016 changes

hflynn:
See https://trello.com/c/q7EZXzW8/109-model-documentation
This is adding the June 2016 changes to the sphinx schema documentation. I've also tidied up some out-of-date info on the developers index page.

Staged at https://www.openmicroscopy.org/site/support/ome-model-staging/developers/index.html and https://www.openmicroscopy.org/site/support/ome-model-staging/schemas/index.html

### DIFF
--- a/formats/schemas/june-2016.txt
+++ b/formats/schemas/june-2016.txt
@@ -46,8 +46,11 @@ Details
   rendering framework, making it easier to implement across clients (see this
   `Design issue <https://github.com/openmicroscopy/design/issues/19>`_ for an
   example discussion).
-- The (Poly)Line Markers ``Circle`` and ``Square`` have also been dropped for
-  similar reasons to the ROI properties.
+- The ``Marker`` enumeration has been reduced to ``Arrow`` only, dropping
+  ``Circle`` and ``Square``. This affects the ``Line.MarkerStart``,
+  ``Line.MarkerEnd``, ``Polyline.MarkerStart`` and ``Polyline.MarkerEnd``
+  attributes. As with the ROI property changes, these are also intended to
+  simplify the graphical aspects of the data model.
 
 The following are simplifications for code generation purposes and have
 no functional effects:
--- a/formats/schemas/june-2016.txt
+++ b/formats/schemas/june-2016.txt
@@ -1,0 +1,66 @@
+Changes for June 2016
+=====================
+
+
+The list of the key changes for the June 2016 major release of the
+OME-XML data model. This schema release will tie in with the Bio-Formats
+5.2 release.
+
+The new major release of the schema has a new namespace and all version
+numbers are reset to 1. As a major release, any file that validated
+correctly using the last major release will probably not validate
+correctly using this new release. Some files that failed to validate
+before will now be valid. It is important to update any file readers and
+writers to understand the changes.
+
+The version number of all schema files is now 1.
+
+This schema uses the new namespace:
+
+::
+
+    http://www.openmicroscopy.org/Schemas/OME/2016-06/
+
+All the schema files are now located at this same namespace.
+
+Overview of changes
+-------------------
+
+- This release will be implemented in Bio-Formats 5.2 and OMERO 5.3.
+- This release introduces the concept of Folders, a new Model Object which may
+  contain Images, ROIs and other Folders, and which has a strict tree
+  hierarchy.
+
+
+Details
+^^^^^^^
+
+- The model element ``Folder`` was added. A Folder specifies a possibly
+  heterogeneous collection of data and may contain other Folders, Images and
+  ROIs. Data may be in multiple Folders but a Folder may not be in more than
+  one other Folder, giving a strict tree hierarchy.
+- The ROI properties ``ROI.Namespace``, ``Shape.Linecap`` and
+  ``Shape.Visible`` have been dropped.
+- The (Poly)Line Markers ``Circle`` and ``Square`` have been dropped.
+
+The following are simplifications for code generation purposes only and have
+no functional effects:
+
+- The model element ``LightSource`` has become an abstract class, meaning it
+  is now a complexType in the .xsd.
+- The model element ``Shape`` has become an abstract class, meaning it
+  is now a complexType in the .xsd.
+- The MapPairs model element has been removed and the Map model element has
+  been made a complexType.
+- xsd:appinfo has been extended to provide more detail for enumeration code
+  generation, in particular for units.
+
+Upgrading and Downgrading
+-------------------------
+
+The XSLT transforms between January 2015 and june 2016 versions are
+available here:
+
+`http://www.openmicroscopy.org/Schemas/Transforms/2015-01-to-2016-06.xsl <http://www.openmicroscopy.org/Schemas/Transforms/2015-01-to-2016-06.xsl>`_
+
+`http://www.openmicroscopy.org/Schemas/Transforms/2016-06-to-2015-01.xsl <http://www.openmicroscopy.org/Schemas/Transforms/2016-06-to-2015-01.xsl>`_
--- a/formats/schemas/june-2016.txt
+++ b/formats/schemas/june-2016.txt
@@ -13,15 +13,14 @@ correctly using this new release. Some files that failed to validate
 before will now be valid. It is important to update any file readers and
 writers to understand the changes.
 
-The version number of all schema files is now 1.
-
 This schema uses the new namespace:
 
 ::
 
     http://www.openmicroscopy.org/Schemas/OME/2016-06/
 
-All the schema files are now located at this same namespace.
+All the schema files have been unified into a single ome.xsd file located at
+this namespace. The version number of this schema file is 1.
 
 Overview of changes
 -------------------
--- a/formats/schemas/june-2016.txt
+++ b/formats/schemas/june-2016.txt
@@ -57,7 +57,7 @@ no functional effects:
 
 - The model elements ``LightSource`` and ``Shape`` have become abstract
   classes, meaning they are now complexTypes in the :file:`.xsd`. This
-  provides a more intuitive object oriented design (e.g. Rectangle is a type
+  provides a more intuitive object-oriented design (e.g. Rectangle is a type
   of Shape rather than Shape containing a Rectangle) and simplifies the data
   representation (schema) as well as the code generation because two different
   ways to model inheritance and polymorphism have been changed to one.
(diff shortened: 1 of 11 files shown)


formats/schemas/june-2016.txt
```diff
@@ -0,0 +1,65 @@
...
+before will now be valid. It is important to update any file readers and
+writers to understand the changes.
+
+This schema uses the new namespace:
+
+::
+
+    http://www.openmicroscopy.org/Schemas/OME/2016-06/
+
+All the schema files have been unified into a single ome.xsd file located at
```
mtbc: maybe some kind of `:file:` markup

formats/schemas/june-2016.txt
```diff
@@ -0,0 +1,65 @@
...
+- The (Poly)Line Markers ``Circle`` and ``Square`` have been dropped.
+
+The following are simplifications for code generation purposes only and have
+no functional effects:
+
+- The model element ``LightSource`` has become an abstract class, meaning it
+  is now a complexType in the .xsd.
+- The model element ``Shape`` has become an abstract class, meaning it
+  is now a complexType in the .xsd.
+- The MapPairs model element has been removed and the Map model element has
```
mtbc: don't know if this model element also deserves backticks

formats/schemas/june-2016.txt
```diff
@@ -0,0 +1,65 @@
...
+  been made a complexType.
+- xsd:appinfo has been extended to provide more detail for enumeration code
+  generation, in particular for units.
+
+Upgrading and Downgrading
+-------------------------
+
+The XSLT transforms between January 2015 and june 2016 versions are
```
mtbc: June

mtbc:
Looking pretty good, thank you!

ghost:
@sbesson It would be good to provide a _rationale_ for these changes, in addition to the change itself to describe the reasons and benefits for making them.  Do we have a concise one for Folders?  And the ROI property removals?  The rest are purely for the benefit of code generation as far as I can see, and don't affect the data at all.

sbesson:
@rleigh-dundee: thanks for raising this point. Adding more context on the decision makes sense. For the folders, I would actually be inclinded to refer to the various blog posts we already published about it, most importantly @mtbc's http://blog.openmicroscopy.org/data-model/future-plans/2016/05/23/folders-upcoming/.

On the ROI property removal, we could also refer the corresponding design issues where the discussion happened and maybe briefly summarize them.

As for the abstract proprietary improvements, is their something to be said about their impact on the OME-XML (in the sense of simplification)? See the samples opened by @dgault https://github.com/openmicroscopy/bioformats/pull/2447 versus their 2015-01 counterpart (`ROI.ome.xml`, `instrument.ome.xml` notably)

ghost:
@sbesson Links would definitely be good, but a single sentence summary for each here would I think be useful for anyone who wants a concise summary of the changes.

sbesson:
As @mtbc is unavailable, I can certainly have a go at a one-line rationale sentence for `Folder` and `ROI` changes (unless @hflynn already has something). Can you provide the same for the C++ related changes (`MapPairs` notably)?

formats/schemas/june-2016.txt
```diff
@@ -0,0 +1,65 @@
...
+correctly using the last major release will probably not validate
+correctly using this new release. Some files that failed to validate
+before will now be valid. It is important to update any file readers and
+writers to understand the changes.
+
+This schema uses the new namespace:
+
+::
+
+    http://www.openmicroscopy.org/Schemas/OME/2016-06/
```
dgault: Im assuming this should be http://www.openmicroscopy.org/Schemas/OME/2016-01/

formats/schemas/june-2016.txt
```diff
@@ -0,0 +1,65 @@
...
+
+Upgrading and Downgrading
+-------------------------
+
+The XSLT transforms between January 2015 and june 2016 versions are
+available here:
+
+`http://www.openmicroscopy.org/Schemas/Transforms/2015-01-to-2016-06.xsl <http://www.openmicroscopy.org/Schemas/Transforms/2015-01-to-2016-06.xsl>`_
+
+`http://www.openmicroscopy.org/Schemas/Transforms/2016-06-to-2015-01.xsl <http://www.openmicroscopy.org/Schemas/Transforms/2016-06-to-2015-01.xsl>`_
```
dgault: Again the 2016-06 should be 2016-01

dgault:
As for the impact or rationale of the abstract changes to Shape and LightSource, other than for code generation purposes, one benefit would perhaps be that it provides a more intuitive object oriented design for the model objects (ie Rectangle is a type of Shape rather than Shape contains a Rectangle) and does simply the OME-XML some what as well.

ghost:
That's definitely the rationale for Shape and LightSource.  We change from having two different ways to model inheritance and polymorphism to one, which simplifies both the schema (data representation) and the code generator, which now only needs to cope with a single method.

Map is a model object which was never used directly, only as a base, and so making it a complexType allows for the removal of the redundant model object.  MapPairs was an unnecessary container which is superseded by using the maps directly in the generated code.

sbesson:
For the other topics, a first proposal would be along the lines of:
- Folders have been primarily introduced to allow the organization of ROI elements into hierarchical (tree?) structures as opposed to flat structures - see http://blog.openmicroscopy.org/data-model/future-plans/2016/05/23/folders-upcoming/ for more details
- the dropping of some ROI properties aims at simplifying various graphical aspects in the data model in favor of more generic enumerated values independent of the rendering framework and easier to implement across clients e.g. OMERO - see discussion on https://github.com/openmicroscopy/design/issues/19

hflynn:
Had an initial stab at adding the rationales from comments. @sbesson I added that design link to the ROI properties point but was a bit confused by the fact it discusses font properties and not any of the properties mentioned.

sbesson:
@hflynn: good point, the issue I linked at effectively discuss properties removed from OMERO with no effect to the OME-XML model. The reason I used it is because this is where I found a largest number of comments from @mtbc, @rleigh-dundee, @will-moore and @joshmoore arguing for a simplification of the rendering/graphical aspects of the model.

hflynn:
Ah! In that case, my link text probably describes the context about right then.

formats/schemas/june-2016.txt
```diff
@@ -0,0 +1,77 @@
...
+  this addition is to allow the organization of ROI elements into hierarchical
+  structures (see this `Folders blog post <http://blog.openmicroscopy.org/data-model/future-plans/2016/05/23/folders-upcoming/>`_ for further discussion).
+- The ROI properties ``ROI.Namespace``, ``Shape.Linecap`` and
+  ``Shape.Visible`` have been dropped. This simplifies graphical aspects in
+  the data model in favor of more generic enumerated values independent of the
+  rendering framework, making it easier to implement across clients (see this
+  `Design issue <https://github.com/openmicroscopy/design/issues/19>`_ for an
+  example discussion).
+- The (Poly)Line Markers ``Circle`` and ``Square`` have also been dropped for
+  similar reasons to the ROI properties.
```
sbesson: Went over the details of this page with @chris-allan @emilroz and @knabar.  This line should be amended to be more specific and explain that: 
1. the `Marker` enumeration is reduced to `Arrow` only
2. the exact name of the affected attributes `Line.MarkerStart`, `Line.MarkerEnd`, `Polyline.MarkerStart` and `Polyline.MarkerEnd`

hflynn: Fix pushed, thanks!

hflynn:
@mtbc can you have another look at this now you're back please? I don't think the last couple of commits have been reviewed as yet

formats/schemas/june-2016.txt
```diff
@@ -0,0 +1,80 @@
...
+  ``Line.MarkerEnd``, ``Polyline.MarkerStart`` and ``Polyline.MarkerEnd``
+  attributes. As with the ROI property changes, these are also intended to
+  simplify the graphical aspects of the data model.
+
+The following are simplifications for code generation purposes and have
+no functional effects:
+
+- The model elements ``LightSource`` and ``Shape`` have become abstract
+  classes, meaning they are now complexTypes in the :file:`.xsd`. This
+  provides a more intuitive object oriented design (e.g. Rectangle is a type
```
mtbc: hyphenate object-oriented

mtbc:
Minor quibble aside, this looks great, thank you. Good to merge.

mtbc:
:+1:

hflynn:
Leaving this open for now as we haven't run the staging job with the new version set yet which may throw up further issues to fix.

sbesson:
Note the version number is now defined in `formats/conf.py` and could be bumped to assess the number of failing links?

hflynn:
Good point, will test locally

hflynn:
This will probably make the staging build yellow again now because the samples aren't up yet but at least we'll know what still needs doing. Think I've caught the bits where the old version number was hardcoded into link text but we'll want to doublecheck. Other remaining to dos is updating the XML examples but this is already on a card.

hflynn:
Also, when we replace the ROI diagram, it needs its own heading rather than being under '3D ROI'

sbesson:
@hflynn: while waiting for the remaining bits (samples, diagrams) to be sorted out here, an alternative would be to remove the last commit bumping the version number and get this PR in with the latest model additions in.

hflynn:
My favoured option is to remove the out-of-date diagrams and replace them with a note that they'll be published asap and then merge and release as soon as the samples are live. I will push another commit once we have the list of diagrams sorted.

hflynn:
Last commit takes a halfway house approach re: diagrams - on the overview page I've commented out all the out-of-date ones as the files specifically says _current data model_. For the more detailed pages, I have also commented out the ROI diagram because it needs updating but the other ones only need Folders adding so I've noted this on the page rather than removing them altogether. 
Arguably, the second annotation diagram should also be removed as the changes needed are not just adding one box - if people think that's inconsistent with the way I've treated the ROI diagram, I can comment out that one too.

Overall, I thought it was better to leave the image links and files in the repo so we can replace like with like when the new ones are ready.

hflynn:
https://www.openmicroscopy.org/site/support/ome-model-staging/developers/model-overview.html etc updated now

sbesson:
:+1: all ready to go for me. @mtbc: are you okay with the last version?

mtbc:
Yes, fine with me.

hflynn:
Ok, I will merge. I want to do a fresh PR to update the OME-TIFF/XML snippets before we publish this though. Will do that asap.

dgault:
Everything looks good for me too 👍

PR #1492, opened 2016-06-15
Replace non-functional example with a working one

pwalczysko:
Just fixing an example in http://www.openmicroscopy.org/site/support/omero5.2/developers/testing.html#individual-test-class-methods. The status at present is that this example does not execute any tests because there are none of the specified name. This PR is replacing the example with more present test, which executes fine with a reasonable output.

Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/developers/testing.html#individual-test-class-methods 

cc @ximenesuk @mtbc

### DIFF
--- a/omero/developers/testing.txt
+++ b/omero/developers/testing.txt
@@ -97,7 +97,7 @@ working on. This can be done using the ``test`` target. For example:
 
 ::
 
-    ./build.py -f components/tools/OmeroJava/build.xml test -DTEST=integration/AdminTest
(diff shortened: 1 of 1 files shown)


sbesson:
:+1:

hflynn:
Actually staged on http://www.openmicroscopy.org/site/support/omero5.3-staging/developers/testing.html#individual-test-class-methods btw

hflynn:
--rebased-to #1502

PR #1493, opened 2016-06-22
Enforce UTF-8 encoding of the XML string in the OME-TIFF specification

sbesson:
See https://trello.com/c/oICUieGO/98-file-formats-xml-restrictions


ghost:
Can I mark this +1000 ?

Looks fine and good to merge.

PR #1496, opened 2016-06-30
Add note that test-compile is needed before running tests

manics:
`./build.py` is insufficient for running `openmicroscopy` tests.


sbesson:
:+1:

hflynn:
--rebased-to #1503

PR #1498, opened 2016-07-01
document extra SQL scripts for OMERO 5.3

mtbc:
https://github.com/openmicroscopy/openmicroscopy/pull/4732 introduces and describes a couple of SQL utility scripts for OMERO server admins to use. Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/server-upgrade.html.

cc: @pwalczysko

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -134,12 +134,14 @@ below. Please refer to each section for additional details.
 Check ahead for upgrade issues
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-There *may* be a "precheck" SQL script provided that performs various
+There is a "precheck" SQL script provided that performs various
 database checks to verify readiness for upgrade. The precheck script
 works even with the OMERO server running so it may be used before
 downtime for the actual upgrade is scheduled. Issues that the script
 reports will need to be resolved before the upgrade may proceed. The
-precheck script will *not* make any changes to the database.
+precheck script will *not* make any changes to the database: it merely
+performs various precautionary checks also done by the actual upgrade
+script.
 
 .. parsed-literal::
 
@@ -158,9 +160,6 @@ precheck script will *not* make any changes to the database.
                                                                         +
 
     (1 row)
-
-If a precheck script is not included in the OMERO.server distribution
-then no such check need be performed.
 
 
 .. _back-up-the-db:
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -284,7 +284,7 @@ Delete certain annotations (optional)
 For various reasons production databases may accumulate non-sharable
 annotations that are orphaned. These are :doc:`structured annotations
 </developers/Model/StructuredAnnotations>` that are 'basic' (`Boolean`,
-`TimeStamp`, `Term`), 'numeric' (`Double`, `Long`), or `Comment`, and
+`Timestamp`, `Term`), 'numeric' (`Double`, `Long`), or `Comment`, and
 that are *not* annotating an object. An illustrative example is that
 deleting a rating in OMERO.insight 5.2 may have left behind the
 corresponding `Long` annotation that captured the rating's number of
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -288,7 +288,10 @@ annotations that are orphaned. These are :doc:`structured annotations
 that are *not* annotating an object. An illustrative example is that
 deleting a rating in OMERO.insight 5.2 may have left behind the
 corresponding `Long` annotation that captured the rating's number of
-stars. The script does *not* delete annotations that have a
+stars. Non-sharable annotations, like comments and ratings, cannot be
+viewed from OMERO.insight or OMERO.web after they have been orphaned
+because they are no longer associated with any model object such as an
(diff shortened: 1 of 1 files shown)


omero/sysadmins/server-upgrade.txt
```diff
@@ -131,6 +131,38 @@ below. Please refer to each section for additional details.
     :local:
     :depth: 1
 
+Check ahead for upgrade issues
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There *may* be a "precheck" SQL script provided that performs various
```
pwalczysko: `May` - this is not giving me enough comfort as a user I would say. I think it would be good to tell me right here whether or not I should worry about this script (see the comment below, combine with the sentence `If the script is not here, then you do not need to worry.`) If I get the explanation right here, I might be lucky enough to be able to skip this paragraph.

omero/sysadmins/server-upgrade.txt
```diff
@@ -131,6 +131,38 @@ below. Please refer to each section for additional details.
...
+                                                                        +
+                                                                        +
+    YOUR DATABASE IS READY FOR UPGRADE TO VERSION |current_dbver|           +
+                                                                        +
+                                                                        +
+
+    (1 row)
+
+If a precheck script is not included in the OMERO.server distribution
+then no such check need be performed.
```
pwalczysko: I think this info should be moved up, see previous comment on line 137

omero/sysadmins/server-upgrade.txt
sbesson: An alternative here might to be to _always_ provide a `precheck` script from 5.3.0 and onwards but have the default script be a no-op.

omero/sysadmins/server-upgrade.txt
pwalczysko: As for this paragraph, I am missing a context. Could there be a link or other reference to the paragraph which explains what `precheck.sql` actually does ? I suppose this is the paragraph drafted in the second commit here.

Edit: This paragraph actually does say what the script does. The thing I am missing here (and that is why I would like the link to be included) is "What happens when this script fails, how do I proceed ?"

omero/sysadmins/server-upgrade.txt
```diff
@@ -280,6 +280,31 @@ intermediate major release.
...
    :forum:`this forum thread <viewtopic.php?f=5&t=7778>` for more information.
 
+Delete certain annotations (optional)
+"""""""""""""""""""""""""""""""""""""
+
+For various reasons production databases may accumulate non-sharable
+annotations that are orphaned. These are :doc:`structured annotations
+</developers/Model/StructuredAnnotations>` that are 'basic' (`Boolean`,
+`TimeStamp`, `Term`), 'numeric' (`Double`, `Long`), or `Comment`, and
+that are *not* annotating an object. An illustrative example is that
```
pwalczysko, line 288: `...annotating an object.` Reformulate : ``Non-sharable orphaned annotations are :doc:`structured annotations
+</developers/Model/StructuredAnnotations>`that are 'basic' (`Boolean`,`TimeStamp`,`Term`), 'numeric' (`Double`,`Long`), or`Comment`, which are _not_ annotating an object.

omero/sysadmins/server-upgrade.txt
```diff
@@ -280,6 +280,31 @@ intermediate major release.
...
+deleting a rating in OMERO.insight 5.2 may have left behind the
+corresponding `Long` annotation that captured the rating's number of
+stars. The script does *not* delete annotations that have a
+custom/non-OME namespace (ns) set.
```
pwalczysko, line 295: I think it would be worth to mention in this rationale that we are talking aoubt ``annotations which cannot be accessed using OMERO.clients anymore, and thus might be an artifact of the workflow rather than useful data.

pwalczysko:
Couple of comments.

omero/sysadmins/server-upgrade.txt
mtbc: Okay to go ahead and adopt @sbesson's idea here?

omero/sysadmins/server-upgrade.txt
mtbc, line 288: How to reformulate it: what needs to change, or what's missing?

omero/sysadmins/server-upgrade.txt
mtbc: Unfortunately, the answer is: "it depends". If it's not obvious enough from the error it generates then people may need to chat to us via Community. If the answer were an easy universal one then the upgrade script would already be doing it for them. Fortunately in the typical case we expect the checks to pass.

omero/sysadmins/server-upgrade.txt
manics: At the very least the "default" script could check the PG version instead of a noop.

pwalczysko:
Relisting for tomorrow, sorry, with the workshop rehearsals did not find time to review it.

mtbc:
No problem, it sounds like you have your priorities correctly ordered! :+1:

pwalczysko:
The changes look good. The staging page has built okay. 
Ready to merge.

omero/sysadmins/server-upgrade.txt
pwalczysko, line 288: Sorry, missed your answer. I would suggest best leave it alone, the additional explanation about what is "Orphaned" and "non-sharable" was added by you and with this it is understandable.

PR #1499, opened 2016-07-11
Update Model diagrams & ref text

hflynn:
As a somewhat temporary solution since we haven't had time to properly redesign these, Gus has added Folders & updated the current ROI model overview.

Staged at https://www.openmicroscopy.org/site/support/ome-model-staging/developers/model-overview.html and following pages.

Also includes some tweaks to the ROI page I spotted when going through the diagram changes needed.


joshmoore:
Does `Folder` need a `ROIRef` in https://github.com/hflynn/ome-documentation/blob/fba372747e5c522df8c3ed2cc87c437d01416565/formats/images/Annotation-Points.png ?

hflynn:
@mtbc can you answer Josh's question?

(Further commit with smaller diagrams & black lines removed coming btw but current state is now on staging)

mtbc:
I don't think so because that's all about annotations. As I noted on https://trello.com/c/q7EZXzW8/109-model-documentation we _do_ need those refs on `Project-Dataset-Top.png` though (and I wonder if Folder should be further up the list with the other containers).

mtbc:
Why are we making all these images executable?

gusferguson:
@mtbc - I don't think it would fit in higher up because your containers are all spread out anyway. I would suggest sticking with it where it is now - it makes life a whole lot easier - and it can be re-visited when we do the complete overhaul.
What do you mean by executable?

mtbc:
@gusferguson: Executable: discussed with @hflynn how to fix. (All the 100644 → 100755 in the diff.)

mtbc:
Ah, yes, we have project above dataset, screen below plate, etc., it's all a bit of a mess anyway, fair enough! Can wait for overhaul.

mtbc:
Good to merge.

PR #1502, opened 2016-07-13
Replace non-functional example with a working one (rebased onto dev_5_2)

hflynn:
This is the same as gh-1492 but rebased onto dev_5_2.

---

Just fixing an example in http://www.openmicroscopy.org/site/support/omero5.2/developers/testing.html#individual-test-class-methods. The status at present is that this example does not execute any tests because there are none of the specified name. This PR is replacing the example with more present test, which executes fine with a reasonable output.

Staged at http://www.openmicroscopy.org/site/support/omero5.2-staging/developers/testing.html#individual-test-class-methods 

cc @ximenesuk @mtbc

### DIFF
--- a/omero/developers/testing.txt
+++ b/omero/developers/testing.txt
@@ -97,7 +97,7 @@ working on. This can be done using the ``test`` target. For example:
 
 ::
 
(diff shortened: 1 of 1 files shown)


hflynn:
--rebased-from #1492

mtbc:
Good to merge.

PR #1503, opened 2016-07-13
Add note that test-compile is needed before running tests (rebased onto dev_5_2)

hflynn:
This is the same as gh-1496 but rebased onto dev_5_2.

---

`./build.py` is insufficient for running `openmicroscopy` tests.


hflynn:
--rebased-from #1496

simleo:
Docs build correctly and changes look good. OK to merge.

PR #1507, opened 2016-07-28
Adding private repo cloning address

hflynn:
Because I always forget this!

Staged at http://www.openmicroscopy.org/site/support/contributing-staging/using-git.html


sbesson:
:+1:

PR #1512, opened 2016-08-19
Update BF C++ ref to OME Files

hflynn:
See https://ci.openmicroscopy.org/view/Docs/job/FORMATS-merge-docs/1034/warnings3Result/
Although this URL is now being redirected, the reference in the docs was still described as Bio-Formats C++ so I've updated it anyway.

Once we have the web navigation etc properly sorted for the new docs, this can get a proper `:ome-files:` ref link in both these docs and the BF docs.

### DIFF
--- a/common/conf.py
+++ b/common/conf.py
@@ -161,6 +161,7 @@ extlinks = {
     'secvuln' : (oo_root + '/info/vulnerabilities/%s', ''),
     'bf_plone' : (oo_site_root + '/products/bio-formats/%s', ''),
     'partner_plone' : (oo_site_root + '/products/partner/%s', ''),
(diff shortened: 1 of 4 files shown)


hflynn:
Staged at http://www.openmicroscopy.org/site/support/ome-model-staging/ome-tiff/tools.html

sbesson:
Looks good  to me. While advertising OME Files C++, can we update the bullet points in the landing page of this documentation set to also mention it alongside Bio-Formats and OMERO?

hflynn:
Will do

hflynn:
@sbesson pushed - the other links on the landing page are to product pages so I've copied that and also added in extlinks for the cpp plone page & downloads since these URLs are already established.

hflynn:
Now live

PR #1514, opened 2016-08-25
Replace GH help link for PRs

hflynn:
This link is broken in both the OMERO & contributing docs and when I poked around, it seems GH now have a bunch of different entries mentioning PRs rather than one definitive article, so I've replaced the link with the relevant help category instead.

Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/ (bottom of landing page) and  http://www.openmicroscopy.org/site/support/contributing-staging/using-git.html

### DIFF
--- a/contributing/using-git.txt
+++ b/contributing/using-git.txt
@@ -831,7 +831,7 @@ then the target user need only click on a button. If not, then there may
 be some back-and-forth on the work done, similar to the code reviews of
 a deliverable branch. For background, see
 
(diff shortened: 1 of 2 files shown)


sbesson:
Looks good, any reason not to use https://help.github.com/articles/about-pull-requests/ which seems to be the new landing page for Pull Requests on the GitHub help?

hflynn:
Ah! No, I just didn't find that via their search option. I'll change it now.

hflynn:
Pushed, thanks @sbesson

PR #1515, opened 2016-08-25
Update UnitsofMeasurement homepage

hflynn:
See https://ci.openmicroscopy.org/view/Docs/job/FORMATS-merge-docs/1040/warnings3Result/
I'm assuming this is the replacement link from google search

Staged at http://www.openmicroscopy.org/site/support/ome-model-staging/developers/ome-units.html

### DIFF
--- a/formats/developers/ome-units.txt
+++ b/formats/developers/ome-units.txt
@@ -14,8 +14,7 @@ attribute pairs into a single unit aware attribute in the output java classes.
 The java classes make use of the types from the new `ome.units.*` classes.
 
 The classes in `ome.units` are inspired by but not a direct implementation of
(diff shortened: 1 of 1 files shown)


sbesson:
Looks good and in agreement with https://jcp.org/en/jsr/detail?id=363. There are really three layers at which we could link http://unitsofmeasurement.github.io/, http://unitsofmeasurement.github.io/unit-api/ and http://unitsofmeasurement.github.io/unit-api/site but I assume the latter makes sense as we refer to the API directly in the documentation.

hflynn:
Top level is essentially a blog/newsfeed, second level is mostly actually info about GH pages because they haven't put their content there so I figured the 3rd level was most actually useful and relevant.

PR #1516, opened 2016-08-25
Web deployment

hflynn:
This adds a new doc for deploying OMERO.web separately to the server and makes the apache deployment as deprecated.

Not sure if we really want to replicate info from the existing nginx page but I think the plan might be for this to replace that page in the future (especially when the parallel apache page is removed).

Staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/index.html


omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+
+
+::
+
+   $ pip install omego
+   $ omego download py
+   2016-08-24 16:20:14,252 [omego.artifa] INFO  Checking http://downloads.openmicroscopy.org/omero/5.2.5/./artifacts/OMERO.py-5.3.0-ice36-b29.zip
+   2016-08-24 16:20:14,252 [omego.fileut] INFO  Downloading http://downloads.openmicroscopy.org/omero/5.2.5/./artifacts/OMERO.py-5.3.0-ice36-b29.zip
+   2016-08-24 16:20:14,514 [omego.artifa] INFO  Unzipping OMERO.py-5.3.0-ice36-b29.zip
+  $ pip install -r OMERO.py/web/requirements-py27-experimental.txt
```
atarkowska: this is now names requirements-py27-trial.txt

atarkowska:
Thank you @hflynn  for all your help. Looks great. Code PR in https://github.com/openmicroscopy/openmicroscopy/pull/4805

https://ci.openmicroscopy.org/job/WEB-DEV-trial-deploy/

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+   -  :pypi:`virtualenv` (optional) tool to create isolated Python environments
+
+   -  :pypi:`PyPI <>` a package management system used to install and manage
+      software packages written in Python. PyPI is already installed if
+      you are using Python 2 >=2.7.9
+
+   -  `omego <https://github.com/ome/omego>`_ OME package management system
+
+   -  `ZeroC`_ IcePy 3.6 (IcePy 3.5 is also supported although requires global
+      installation from RPM)
```
mtbc: maybe a more generic term here to include Ubuntu, etc. too? or do Ubuntu users have to install the RPM somehow too?

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+
+   -  `Pillow`_ <3.4
+
+   -  `NumPy <http://www.numpy.org>`_ >=1.9 
+
+-  A `WSGI <http://wsgi.readthedocs.org>`_ capable web server such as
```
mtbc: hyphenate WSGI-capable

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+   `nginx <http://nginx.org/>`_ and `gunicorn <http://docs.gunicorn.org/>`_,
+
+
+Gunicorn default configuration
+------------------------------
+
+Install above OMERO.web dependences using the package management tools:
```
mtbc: "_the_ above"

mtbc: dependencies

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+
+
+Additional settings can be configured by changing the following properties:
+
+- :property:`omero.web.application_server.max_requests` to 500
+
+- :property:`omero.web.wsgi_workers` to (2 x NUM_CORES) + 1
+
+  .. note::
+      **Do not** scale the number of workers to the number of clients
```
mtbc: With the bold "do not" I wonder if you want to consider making this a warning rather than a note.

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
+OMERO.web WSGI deployment (TRIAL)
+=================================
+
+OMERO.web is the web application component of the OMERO platform which
+allows for the management, visualization (in a fully multi-dimensional
+image viewer) and annotation of images from a web browser. It also
+includes the ability to manage users and groups.
+
+Deploying OMERO.web separately to OMERO.server is recommended as they
```
mtbc: we have "separately from" below -- you might be making a valid distinction here in how you use it, just wanted to make sure it was intentional

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+    :property:`omero.web.application_server.max_requests` should be set to 0
+
+
+Benchmark
+---------
+
+We run example benchmarks on rendering thumbnails and 512x512 pixels planes
+for 100 concurrent users making 5000 requests in total::
+
+    $ ab -n 5000 -c 100 https://server.openmicroscopy.org/omero/webclient/render_thumbnail/size/96/1234/
```
mtbc: :+1: for providing examples!

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+To run the WSGI server in debug mode, enable
+`Gunicorn logging <http://docs.gunicorn.org/en/stable/settings.html#logging>`_
+using :property:`omero.web.wsgi_args`:
+
+::
+
+    $ OMERO.py/bin/omero config set omero.web.wsgi_args -- "--log-level=DEBUG --error-logfile=/home/omero/OMERO.server/var/log/error.log".
+
+
+EXPERIMENTAL: Gunicorn advance configuration
```
mtbc: "advanced"?

omero/sysadmins/unix/install-web-trial.txt
atarkowska: @mtbc this is something I would like to discuss across the team. How sysadmin prefer to install. The problem with Ice 3.6 is that it requires gcc because it has to compile IcePy. As there is no other options. If gcc is not accepted then they need to go for `yum -y install ice-python`

My biggest problem is that packages installed from RPM are old

atarkowska: Would be useful to have 2 installation examples below, pip and yum/apt-get ?

mtbc: I think it's okay to expect admins to be able to find it for their distro, I was just suggesting that "RPM" is a bit distro-specific in how to say that: for instance, I installed Debian's `python-zeroc-ice` package which isn't "installation from RPM" (I'd hope 3.6 will hit sid soon actually, Ubuntu already has it). Even just something like "installation from your distribution's software package repository" or somesuch might do?

mtbc: > How sysadmin prefer to install. 

is an interesting question that certainly need not block this PR. For instance, I have Pillow <3 because I installed Debian 8's `python-pil` (as we use newer features in our playbooks, things like ansible and shade I think I pulled in from Debian's packages for what will be version 9). Making much use of multiple package systems on the one machine just leads to confusion -- in knowing even which system to ask why this file is here, keeping up to date with security fixes, sweeping away cruft, etc. -- fortunately OMERO requires popular enough stuff that I can largely keep everything managed by the one system. Maybe I'm unusual, but I much prefer to use my distribution's package manager rather than that _and_ an extra package manager for every language I need extra stuff for: pip, cpan, hackage, etc. It also makes it simpler to restore state to a fresh system. However, @rleigh-codelibre or @kennethgillen for instance may have a wholly different view.

mtbc: There's an interesting related issue that perhaps we might capture on a card or something: often for prerequisites (e.g. PostgreSQL) if some version (a) is the "obvious" natural one to install on my distro (e.g., what I get if I don't put a version number in the package name) and (b) does satisfy the prerequisites (at least minimum, or whatever) then I'll tend to install that rather than one the version _recommended_ for software that needs it partly because (a) different software will tend to recommend different versions anyway and if I can satisfy all with one I'll do that and (b) when I have issues with it (e.g., in distro upgrade) I'd rather have the version that most other users of the distro have because then I'm more likely to find faster help. (Of course, I make exceptions depending on if the recommendation for above minimum is for reasons more concrete than, "newer so probably shinier too".) Again, I don't know if I'm unusual in this, but if not then it might have implications for our playbooks.

omero/sysadmins/unix/install-web-trial.txt
hflynn: A warning box implies danger though, I'm not aware that's needed. This copies the existing formulation on the nginx page

mtbc: Aha, okay.

omero/sysadmins/unix/install-web-trial.txt
joshmoore: > Maybe I'm unusual, but I much prefer to use my distribution's package manager rather than that and an extra package manager for every language I need extra stuff for: pip, cpan, hackage, etc. It also makes it simpler to restore state to a fresh system.

A difference here, @mtbc, is that I think we're quite close to having a requirements.txt shipped with OMERO, so that on every upgrade we would say, "please run `venv/bin/pip install -r requirements.txt` to have the latest" which isn't necessarily something you'd want to do with your wider system.

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+   -  `Pillow`_ <3.4
+
+   -  `NumPy <http://www.numpy.org>`_ >=1.9 
+
+-  A `WSGI <http://wsgi.readthedocs.org>`_-capable web server such as
+   `nginx <http://nginx.org/>`_ and `gunicorn <http://docs.gunicorn.org/>`_,
+
+
+Gunicorn default configuration
+------------------------------
```
will-moore: It seems that the instructions following this heading are more about "Installing requirements" rather than "Gunicorn default configurations"? Would a different heading make more sense?

will-moore: Or maybe we don't need this heading, since we are already under "Prerequisites"?

atarkowska: Sorry that can be removed. as below is an example of installation. I am about to open omero-install PR to have walkthrough,

will-moore:
Testing...

```
$ virtualenv omeroweb
$ source omeroweb/bin/activate
(omeroweb) $ pip install omego
(omeroweb) $ omego download --ice 3.6 --branch OMERO-DEV-breaking-build py
(omeroweb) $ mv OMERO.py-5.3.0-m2-1083-a3ff718-ice36-b754 OMERO.py
(omeroweb) $ pip install -r OMERO.py/share/web/requirements-py27-trial.txt
(omeroweb) $ cd OMERO.py 
(omeroweb) $ bin/omero config set omero.web.application_server development
(omeroweb) $ bin/omero config set omero.web.debug True
(omeroweb) $ bin/omero web start
...
Django version 1.8.14, using settings 'omeroweb.settings'
Starting development server at http://127.0.0.1:4080/
```

UPDATE: Works fine (after removing other OMERO libs from PATH etc).

atarkowska:
@will-moore could you review your PYTHONPATH and how many OMERO you have there (looks like you have OMERO-ice3.5 build somewhere. We have that deployed on CI https://ci.openmicroscopy.org/job/WEB-DEV-trial-deploy/7/console and it works even then cowfish has Ice 3.5 installed globally

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+
+::
+
+   $ pip install omego
+   $ omego download py
+   2016-08-24 16:20:14,252 [omego.artifa] INFO  Checking http://downloads.openmicroscopy.org/omero/5.2.5/./artifacts/OMERO.py-5.3.0-ice36-b29.zip
+   2016-08-24 16:20:14,252 [omego.fileut] INFO  Downloading http://downloads.openmicroscopy.org/omero/5.2.5/./artifacts/OMERO.py-5.3.0-ice36-b29.zip
+   2016-08-24 16:20:14,514 [omego.artifa] INFO  Unzipping OMERO.py-5.3.0-ice36-b29.zip
+  $ pip install -r OMERO.py/web/requirements-py27-trial.txt
+  $ ln -s `pwd`/OMERO.py-5.2.5-ice36-b28 `pwd`/OMERO.py
```
will-moore: I think this needs to go before the previous line

atarkowska: yes that should be 

```
$ ln -s `pwd`/OMERO.py-5.2.5-ice36-b28 `pwd`/OMERO.py
$ pip install -r OMERO.py/share/web/requirements-py27-trial.txt
```

omero/sysadmins/unix/install-web-trial.txt
```diff
@@ -0,0 +1,348 @@
...
+
+
+::
+
+   $ pip install omego
+   $ omego download py
+   2016-08-24 16:20:14,252 [omego.artifa] INFO  Checking http://downloads.openmicroscopy.org/omero/5.2.5/./artifacts/OMERO.py-5.3.0-ice36-b29.zip
+   2016-08-24 16:20:14,252 [omego.fileut] INFO  Downloading http://downloads.openmicroscopy.org/omero/5.2.5/./artifacts/OMERO.py-5.3.0-ice36-b29.zip
+   2016-08-24 16:20:14,514 [omego.artifa] INFO  Unzipping OMERO.py-5.3.0-ice36-b29.zip
+  $ pip install -r OMERO.py/web/requirements-py27-trial.txt
```
will-moore: Needs to be OMERO.py/share/web/requirements-py27-trial.txt

hflynn:
@aleksandra-tarkowska I had to manually adjust the indenting at the top of the osx-ice3.6 walkthrough to make the build happy so you'll need to adjust your autogen for that file next week.

joshmoore:
Leaving @aleksandra-tarkowska to open a PR against https://github.com/ome/omeroweb-install

atarkowska:
I fixed that before omeroweb-install was transferred to ome. This in only placeholder. We will test it once autogen job turns green.

joshmoore:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs-autogen-omeroweb/4/ :smile:

PR #1517, opened 2016-08-28
Update glossary for 5.3.0-m3 model

jburel:
Update model object glossary for 5.30-m3.
This was not updated for PR https://github.com/openmicroscopy/openmicroscopy/pull/4793. 
It was for the previous DB changes.
Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/developers/Model/EveryObject.html.

### DIFF
--- a/omero/developers/Model/EveryObject.txt
+++ b/omero/developers/Model/EveryObject.txt
@@ -4,14 +4,6 @@ Glossary of all OMERO Model Objects
 Overview
 --------
 
-In navigating the model objects used by the :doc:`OMERO API
-<../Modules/Api>` and in :omerocmd:`hql` it is often useful to look up
-the names of the object properties and the types of their values. This
-reference document lists every OMERO model object and their more useful
-properties, with an emphasis on enumerating every direct relationship
-among the objects.
-
-
 Reference
 ---------
 
@@ -212,7 +204,7 @@ Properties:
 ChannelBinding
 """"""""""""""
 
-Used by: :ref:`RenderingDef.waveRendering <OMERO model class RenderingDef>`
+Used by: :ref:`CodomainMapContext.channelBinding <OMERO model class CodomainMapContext>`, :ref:`RenderingDef.waveRendering <OMERO model class RenderingDef>`
 
 Properties:
   | active: ``boolean``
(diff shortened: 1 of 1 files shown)


omero/developers/Model/EveryObject.txt
```diff
@@ -4,14 +4,6 @@ Glossary of all OMERO Model Objects
 Overview
 --------
 
-In navigating the model objects used by the :doc:`OMERO API
```
mtbc, line 7: We probably do want this introductory paragraph. (Sorry, it's not autogenerated.)

jburel: I was surprised to see it gone, but It was too late for me to do further investigation

mtbc: Ideally we'll someday have code that, rather than just outputting a new file, optionally reads an existing one, copies that to the output, and just replaces the autogenerated segment of it.

omero/developers/Model/EveryObject.txt
```diff
@@ -5,13 +5,12 @@ Overview
 --------
 
 In navigating the model objects used by the :doc:`OMERO API
-<../Modules/Api>` and in :omerocmd:`hql` it is often useful to look up
-the names of the object properties and the types of their values. This
-reference document lists every OMERO model object and their more useful
-properties, with an emphasis on enumerating every direct relationship
+<../Modules/Api>` and in :omerocmd:`hql` it is often useful to look up
```
mtbc: seem to have added some trailing spaces

omero/developers/Model/EveryObject.txt
```diff
@@ -11,7 +11,6 @@ reference document lists every OMERO model object and their more useful
 properties, with an emphasis on enumerating every direct relationship
 among the objects.
 
-
```
mtbc, line 14: rendering seems unaffected by this

mtbc:
Good to merge.

omero/developers/Model/EveryObject.txt
sbesson: We could certainly make use of `literalinclude` if we want to preserve non-generated code in this documentation repository only

jburel: I will add an item to the board

mtbc: It's not easy to check out this repository and generate that model documentation which might be annoying. Still, could use includes to make this documentation update easier - that does seem like a good idea.

omero/developers/Model/EveryObject.txt
hflynn: As long as there's one blank line between text and headings, it just ignores additional ones

PR #1518, opened 2016-08-29
Remove broken javadoc links & old content from What's new for devs for m3

hflynn:
See https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-latest-docs/169/warnings3Result/ - the old graphs javadoc links needed removing from the doc as they are now gone from the code.

I've also removed the old content, leaving just some heading we will likely re-use, so we don't get 5.2 and 5.3 content mixed up when we come to finish updating this page. This is an interim fix to get the m3 build green


jburel:
Travis is green thanks @hflynn merging

PR #1525, opened 2016-09-12
move model object glossary overview paragraph to separate file

mtbc:
This PR should have no effect on the rendered output of http://www.openmicroscopy.org/site/support/omero5.3-staging/developers/Model/EveryObject.html as compared with http://www.openmicroscopy.org/site/support/omero5.3/developers/Model/EveryObject.html. Moving the overview paragraph out from the autogenerated remainder of the page makes it easier to rerun autogeneration or adjust the text without the two conflicting.

### DIFF
--- a/omero/developers/Model/EveryObject.txt
+++ b/omero/developers/Model/EveryObject.txt
@@ -4,12 +4,7 @@ Glossary of all OMERO Model Objects
 Overview
 --------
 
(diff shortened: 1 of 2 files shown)


hflynn:
Building staging now to check

hflynn:
Yep, can confirm staging looks the same, merging thanks

PR #1529, opened 2016-09-19
apache 2.4 command

jburel:
Fix command if apache 2.4 is installed


hflynn:
Staged at https://www.openmicroscopy.org/site/support/omero5.2-staging/sysadmins/unix/install-web/install-apache.html

hflynn:
Renders fine on staging and text update matches Ola's recommendation on the forum so good to merge. Do we want to add a note to the apache docs re:deprecation? I can open a new PR to do so or include it when I copy over the version history.

jburel:
@hflynn: we should add a note yes. So it is obvious.

PR #1534, opened 2016-09-21
Add ome-files-cpp staging URL to CI docs

hflynn:
As suggested on https://github.com/ome/ome-cmake-superbuild/pull/94 this PR adds the staging URL to the OME Files C++ CI docs.

Staged at https://www.openmicroscopy.org/site/support/contributing-staging/ci-ome-files.html

@rleigh-codelibre not sure how this deployment happens, can expand if needed


mtbc:
**Good to merge**, thank you.

(Though, now you've got me wondering if the job should be on https://www.openmicroscopy.org/site/support/contributing-staging/ci-docs.html which is where I might have actually looked!)

PR #1537, opened 2016-09-22
add notes about making minor model changes

mtbc:
Extends http://www.openmicroscopy.org/site/support/contributing-staging/data-model-schema.html to clarify how to adjust the process for "minor" model releases.

### DIFF
--- a/contributing/data-model-schema.txt
+++ b/contributing/data-model-schema.txt
@@ -137,14 +137,10 @@ Avoid changing:
 Testing
 ^^^^^^^
 
-.. TODO:
-   After the 5.1 release and redirect, the link below can become
-   :bf_doc:`C++ implementation <developers/cpp.html>`
-
 Once the above changes have been made and committed, it is time to test.
 This requires having various prerequisites installed for Bio-Formats
 development, including for the `C++ implementation
-<http://www.openmicroscopy.org/site/support/bio-formats5.1/developers/cpp/overview.html>`_.
+<http://downloads.openmicroscopy.org/latest/ome-files-cpp/>`_.
 Before each test, :ref:`clean the repository <clean_bf_repo>`::
 
   git clean -dfx
--- a/contributing/data-model-schema.txt
+++ b/contributing/data-model-schema.txt
@@ -312,6 +312,9 @@ generate the published schemas pages from the HEAD of Bio-Formats develop
 branch. When the job is promoted, the generated content is deployed under
 :file:`/var/www/html/www.openmicroscopy.org/specification/Schemas` on
 `web-prod` under using the :devs_doc:`scc deploy<scc.html>` script.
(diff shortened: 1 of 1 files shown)


hflynn:
Staging build running now

contributing/data-model-schema.txt
```diff
@@ -29,9 +29,32 @@ wastes your time. From the top-level :file:`bioformats` folder, while
 the most thorough is :command:`git clean -dfx`.
 
 
+Major or minor release?
+^^^^^^^^^^^^^^^^^^^^^^^
+
+A minor release of the OME model schema may suffice for changes like
```
sbesson, line 35: Thanks a lot for that. As we are using `may/must`, I would propose to add the usual reference to [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) for the exact interpretation of the terms.

mtbc: `git grep 2119` in `ome-documentation` gives me zilch; could you point me to an example of what I should say? Do you want the all-caps style too?

sbesson: Yes I do not think we have used it extensively across our documentation yet. An example of usage is at the top of https://github.com/openmicroscopy/design/issues/55. 

Happy to discuss if this is still premature but we could certainly promote a more general usage  /cc @hflynn @rleigh-codelibre

hflynn: I am currently working on a BF docs PR which will also reference it so happy to have this become more widely used (it isn't referenced as yet but I think Roger has already used it for the version requirements page wording - I certainly have https://trello.com/c/HSPvUKRK/474-best-practice-for-imperatives-key-word-usage on the docs board after he pointed it out to me)

mtbc: I'm not sure they add much in this case (English "must" is already unequivocal even without the emphasis I added); I also wonder if those definitions should be a more general intro to the docs than repeated at the top of most pages, then the pages can just put them in some special style (like RFCs' all-caps)? Could certainly add here if seems warranted but I think this might be worth a deeper think beyond this PR.

sbesson: Oh dear, didn't mean to open a can of worms :) Having this sentence as part of the introduction of the documentation would certainly be a sensible option especially if we reuse them more. Happy to use @hflynn's Trello card to channel this discussion.

sbesson:
Overall this reads well and describes most of the steps we have been through for 2016-06 version 2. Only minor question is whether adding a `seealso` note pointing at https://github.com/openmicroscopy/bioformats/pull/2553 for an example of minor schema release changes would be valuable here.

dgault:
These changes look good and certainly make sense and cover any minor release differences that I can think of. It all reads well and is easy enough to understand from my perspective anyway.

PR #1539, opened 2016-09-23
add examples of model change pull requests

mtbc:
In https://github.com/openmicroscopy/ome-documentation/pull/1537#issuecomment-249163837 @sbesson suggests a `seealso` note referencing past model changes. Staged at http://www.openmicroscopy.org/site/support/contributing-staging/data-model-schema.html#major-or-minor-release.


mtbc:
@sbesson: Help! CI is breaking the pull request links by setting `SOURCE_USER`. Suggestions?

sbesson:
I can see 2 solutions: either we define an extlink which does not inherit the configurable `omero_github_root` (e.g. see `omero_scc_branch` which points at an explicit fork) and always points at origin or we actively move away from the `SOURCE_USER` strategy across the board. The latter would be equivalent to treating GitHub source code links in the same way we treat e.g. API links where documentation needs to be updated after publication/tagging.

dgault:
The staged links are working correctly and linking to the PR's fine, formatting on the page also looks well. The merge docs build is now green again so should be good to merge.

PR #1540, opened 2016-09-27
Document the hyphen syntax for CLI graph commands

ximenesuk:
This PR adds some information about the hyphen syntax for the currently documented CLI graph commands, `delete` and `chgrp`.

### DIFF
--- a/omero/users/cli/chgrp.txt
+++ b/omero/users/cli/chgrp.txt
@@ -54,9 +54,8 @@ project 51 and datasets 53 and 54 to group 5.
     $ bin/omero chgrp 5 Dataset:54,53 Project:51
     $ bin/omero chgrp 5 Dataset:53 Project:51 Dataset:54
 
-To move a number of objects with sequentially numbered IDs an abbreviated
-form using a hyphen to specify an ID range can be used. This form can also be
-mixed with comma-separated IDs.
+To move a number of objects with sequentially numbered IDs a hyphen can be used
+to specify an ID range. This form can also be mixed with comma-separated IDs.
 ::
 
     $ bin/omero chgrp 5 Project:51 Dataset:53-56
--- a/omero/users/cli/chgrp.txt
+++ b/omero/users/cli/chgrp.txt
@@ -54,6 +54,20 @@ project 51 and datasets 53 and 54 to group 5.
     $ bin/omero chgrp 5 Dataset:54,53 Project:51
     $ bin/omero chgrp 5 Dataset:53 Project:51 Dataset:54
 
+To move a number of objects with sequentially numbered IDs an abbreviated
+form using a hyphen to specify an ID range can be used. This form can also be
+mixed with comma-separated IDs.
+::
+
+    $ bin/omero chgrp 5 Project:51 Dataset:53-56
+    $ bin/omero chgrp 5 Dataset:53-56,65,101-105,201,202
+
+.. note::
+    When moving multiple objects in a single command, if one object cannot
+    be moved then the whole command will fail and none of the specified
+    objects will be moved. The :option:`--dry-run` can be useful as a check
+    before trying to move large numbers of objects.
+
 Moving lower level objects
 ==========================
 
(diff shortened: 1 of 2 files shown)


omero/users/cli/delete.txt
```diff
@@ -54,6 +54,20 @@ project 51 and datasets 53 and 54.
     $ bin/omero delete Dataset:54,53 Project:51
     $ bin/omero delete Dataset:53 Project:51 Dataset:54
 
+To delete a number of objects with sequentially numbered IDs an abbreviated
+form using a hyphen to specify an ID range can be used. This form can also be
```
mtbc: I'm not fond of, "... using ... can be used."

omero/users/cli/delete.txt
```diff
@@ -54,6 +54,20 @@ project 51 and datasets 53 and 54.
...
+mixed with comma-separated IDs.
+::
+
+    $ bin/omero delete Project:51 Dataset:53-56
+    $ bin/omero delete Dataset:53-56,65,101-105,201,202
+
+.. note::
+    When deleting multiple objects in a single command, if one object cannot
+    be deleted then the whole command will fail and none of the specified
+    objects will be deleted. The :option:`--dry-run` can be useful as a check
```
mtbc: renders as, "The --dry-run can be useful ..." which seems odd.

mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/users/cli/chgrp.html and http://www.openmicroscopy.org/site/support/omero5.3-staging/users/cli/delete.html.

ximenesuk:
These last two commits hopefully address @mtbc's comments.

omero/users/cli/chgrp.txt
```diff
@@ -54,6 +54,19 @@ project 51 and datasets 53 and 54 to group 5.
...
+to specify an ID range. This form can also be mixed with comma-separated IDs.
+::
+
+    $ bin/omero chgrp 5 Project:51 Dataset:53-56
+    $ bin/omero chgrp 5 Dataset:53-56,65,101-105,201,202
+
+.. note::
+    When moving multiple objects in a single command, if one object cannot
+    be moved then the whole command will fail and none of the specified
+    objects will be moved. The :option:`--dry-run` option can be useful as a
```
mtbc, line 67: It's a pity that this link takes us to a discussion of deleting, not moving, but possibly an issue outwith this PR?

ximenesuk: Ah, that's sphinx! I don't know if there is a way of using the markup more discriminately. I assume the definition of `--dry-run` as an option reference on the delete page is the one that is parsed last and so the one that is indexed. @hflynn may know if there is a way to restrict references to the same page while preserving the desired markup.

mtbc: @sbesson: Any thoughts? Given that @hflynn is back after @ximenesuk leaves I'm fine with this being merged as-is and the `--dry-run` indexing issue being noted on a 5.3.0 docs card for fixing in a separate PR.

ximenesuk: It's more than that one option potentially. The option markup is effectively broken if we have commands with the same option names, which we do. Often the description might still apply but this isn't guaranteed and moving off the page seems wrong. Unless there is some clever way of restricting references in Sphinx it may be that options use a more explicit reference. This probably requires a card.

I'd suggest either, merging this PR as is or dropping the option markup in these two note boxes and letting the reader scroll down.

mtbc: **I'd be fine with merging this PR as is**, just add a 5.3.0 card about the option markup issue and we can correct either way before release.

ximenesuk: See: https://trello.com/c/IG1JLPxA/514-option-markup-and-sphinx (on the Docs board for now. I'll let someone move it to the 5.3.0 board if necessary)

mtbc:
Good to merge.

PR #1549, opened 2016-10-03
Improve explanation of advanced scripts workflow

pwalczysko:
Add example of output after starting processor for advanced (non-admin) scripting workflow.
See https://www.openmicroscopy.org/site/support/omero5.2-staging/developers/scripts/advanced.html for buit version.

The motivation is also described in [trello card](https://trello.com/c/81ISVYsi/118-cannot-import-script-via-cli#comment-57b59d9982efadf70e294798).

cc @jburel @will-moore

PR #1550, opened 2016-10-04
Improve explanation of advanced scripts workflow (rebased onto develop)

pwalczysko:
This is the same as gh-1549 but rebased onto develop.

---

Add example of output after starting processor for advanced (non-admin) scripting workflow.
See https://www.openmicroscopy.org/site/support/omero5.3-staging/developers/scripts/advanced.html for buit version.

The motivation is also described in [trello card](https://trello.com/c/81ISVYsi/118-cannot-import-script-via-cli#comment-57b59d9982efadf70e294798).

cc @jburel @will-moore

### DIFF
--- a/omero/developers/scripts/advanced.txt
+++ b/omero/developers/scripts/advanced.txt
@@ -32,7 +32,7 @@ the client (see diagram, below).
 
       OMERO scripting workflow
 
--  You need to download 'Ice' from ZeroC and set the environment
+-  Install 'Ice' from ZeroC and set the environment
    variables, as described in the
    :doc:`server installation page </sysadmins/unix/server-installation>`.
 -  You also need the OMERO server download. Go to the :downloads:`OMERO
@@ -69,8 +69,8 @@ start the user processor and give it the user's ID:
 ::
 
     $ cd Desktop/OMERO.server-5.3.x-icexx-bxx/
-    $ bin/omero -s openmicroscopy.org.uk -u will user list
-    $ bin/omero -s openmicroscopy.org.uk -u will script serve user=5
+    $ bin/omero user list
+    $ bin/omero script serve user=5
 
 From this point on, the user and admin workflows are the same, except
 for a couple of options that are not available to regular users. Also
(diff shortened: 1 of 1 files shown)


pwalczysko:
--rebased-from #1549

hflynn:
Building on staging now

hflynn:
@pwalczysko since you have this open anyway, can you get rid of the OMERO.server-Beta-4.2 references in the first section too please?

omero/developers/scripts/advanced.txt
```diff
@@ -46,18 +46,29 @@ openmicroscopy.org.uk and user: will
...
 collaborative group you need to set up your local user processor to
 accept scripts from that user. First, find the ID of the user, then
 start the user processor and give it the user's ID:
 
 ::
 
-    $ cd Desktop/OMERO.server-Beta-4.2/
+    $ cd Desktop/OMERO.server-5.3.x-icexx-bxx/
     $ bin/omero -s openmicroscopy.org.uk -u will user list
     $ bin/omero -s openmicroscopy.org.uk -u will script serve user=5
```
will-moore: Not part of this PR, but you could remove the "-s openmicroscopy.org.uk -u will" from these commands, since you'll be logged in already.

will-moore:
Looks fine. I haven't tested this workflow recently but it reads fine.
I wonder if we want to update the diagram at the start, since this one still has the (very) old Insight logo.
Also maybe "...need to download 'Ice' " sounds a bit wrong. Probably better to use 'installl Ice'.

hflynn:
@will-moore is the diagram useful? Are there any other changes you'd suggest or shall I just update the logo?

pwalczysko:
This should fix the "text changing" comments of @will-moore

will-moore:
@hflynn I always like a diagram, even if it's very simple.
It's possible there's more detail that could be added to the diagram, but I think even as it is it's fine.

will-moore:
Looks good, thanks. Good to merge.

hflynn:
--rebased-to #1564

PR #1553, opened 2016-10-07
Differentiate CL options with same name

hflynn:
See https://trello.com/c/IG1JLPxA/196-option-markup-and-sphinx - sphinx can't index options properly if they have the same name so I've used the program option to differentiate their context so the links remain on the same page. Hopefully it isn't too confusing to have the command and flag listed after each other given the examples show that text is required between them for actual usage. 

Staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/users/cli/chgrp.html and https://www.openmicroscopy.org/site/support/omero5.3-staging/users/cli/delete.html and also check https://www.openmicroscopy.org/site/support/omero5.3-staging/genindex.html to make sure there are no [1] marked double entries for command line options anymore.


will-moore:
The links work fine and I think it reads OK too. Good to merge.

PR #1574, opened 2016-11-18
Update Windows support life links

hflynn:
See https://ci.openmicroscopy.org/view/Failing/job/OMERO-DEV-merge-docs/538/warnings3Result/

Although only the Windows 10 link is broken at the moment, this new link is much more user-friendly than the old life cycle index pages so I've replaced all the references it covers.

Staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/version-requirements.html

Should make the merge build green again.

### DIFF
--- a/omero/sysadmins/version-requirements.txt
+++ b/omero/sysadmins/version-requirements.txt
@@ -271,7 +271,7 @@ Microsoft Windows
       - Dropped
       - Dropped
       - Dropped
-      - `Ref <http://support.microsoft.com/lifecycle/?p1=11737>`__
+      - `Ref <https://support.microsoft.com/en-gb/help/13853>`__
     * - Win 7
       - from Oct 2009
       - to Jan 2015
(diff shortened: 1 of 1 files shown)


hflynn:
Ok, apparently the build doesn't like this link :(

hflynn:
This is waiting on the outcome of https://trello.com/c/ccx2knE7/528-check-whether-sphinx-can-be-upgraded-for-all-docs-jobs as these links need to go on the ignore list if we don't upgrade the linkchecker to cope with links that don't work with HEAD requests.

hflynn:
Don't mind me, branch cleaning before my cuppa had kicked in obviously!

hflynn:
As established yesterday and described on the Trello cards, updating sphinx will require build fixes too and we haven't yet been able to establish that versions prior to 1.5 will fix the issue and 1.5 is still in beta at present. So merging this and whitelisting the links for now instead.

PR #1577, opened 2016-11-28
Add m6 to history

jburel:
cc @hflynn


hflynn:
Ah, sphinx doesn't like the link text, I'll fix this up on a fresh PR

PR #1578, opened 2016-11-28
M6 history

hflynn:
Replaces #1577 with an additional fix for the BF doc link - sphinx doesn't like you repeating the link text with a different URL. I'll open a corresponding PR on the codebase to make these match again.

Staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/users/history.html

cc @jburel


hflynn:
To be replaced by autogen version

PR #1583, opened 2016-12-01
Copy missing files

jburel:
Copy the setup files
Reported in the following post https://www.openmicroscopy.org/community/viewtopic.php?t=8163&p=17699#p17699

PR #1585, opened 2016-12-06
Copy missing files (rebased onto develop)

jburel:
This is the same as gh-1583 but rebased onto develop.

----

Copy the setup files
Reported in the following post https://www.openmicroscopy.org/community/viewtopic.php?t=8163&p=17699#p17699


jburel:
--rebased-from #1583

jburel:
--exclude

PR #1597, opened 2017-01-04
Path to folder

jburel:
Fix path to doc folder.

This should fix the autogen job failures 
The version of ansible installed on spider-5 is 2.2.0, the relative path used in omeroweb-install implies that the folder is now created under ``ansible``

An omeroweb-install will follow so we have the same outcome with ansible version 2.2 and 2.1


jburel:
Merging now that the corresponding PR in omeroweb-install has been merged

PR #1598, opened 2017-01-05
omero: cpp: Remove instructions for building via build.py

rleigh-codelibre:
See https://trello.com/c/7rOcKabK/255-clean-up-c-build

- Remove the ant build instructions
- Change the VS2010 sample to use 2012 (the latest supported for Ice 3.5)


waxenegger:
I could not find any typos/syntax errors in the lines that were changed, rest is removal, so I think that's all fine.

PR #1601, opened 2017-01-19
Html doc zip

jburel:
Zip the html doc.
The zip is added to  ``omero/_build``, created using for example ``ant html -Domero.release="5.3.0-m7"``
The directory is deleted with ``ant clean``


dominikl:
I'm not very much familiar with the build system, can't say much about the code changes, but it works like expected: The docs zip is created in the `build` directory. 👍

jburel:
Discussed with @sbesson today, i am going to reuse ``omero/_build`` directory instead of creating a new directory.
Another commit to be pushed

sbesson:
:+1: for the re-usage of the `sphinx.builddir` and the removal of `JENKINS_JOB` as our live documentation should not depend on CI.

sbesson:
Tested locally using e.g. ` ant html -Domero.release="5.3.0-m6"`. The logic works as expected, the zip is correctly created under `omero/_build` and cleaned using the `clean` target. Today's [CI job](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/586/) was also green with this PR including the zipping of the HTML documentation as a bundle.

As discussed with @jburel, the main unknown remain the name of the bundle as `OMERO-x.y.z.zip` is quite ubiquitous. However, since the Javadoc bundle is currently named as `OMERO.docs-x.y.z.zip` in https://github.com/openmicroscopy/openmicroscopy, we might want to include both bundles in the naming discussion for OMERO 5.3.0 (as an external reference, Bio-Formats currently uses `-doc` vs `-javadoc`).

hflynn:
Re naming: `-doc` and `-javadoc` to match BF would suit me

PR #1611, opened 2017-02-13
Remove ice35

jburel:
Remove installation pages for ice 3.5

cc @hflynn


hflynn (approved):
Looks good on staging thanks!

PR #1613, opened 2017-02-14
Ubuntu 16.04

jburel:
Add ubuntu 16.04 installation page

cc @hflynn

### DIFF
--- a/omero/sysadmins/unix/server-ubuntu1604-ice36.txt
+++ b/omero/sysadmins/unix/server-ubuntu1604-ice36.txt
@@ -89,15 +89,7 @@ Configure:
 
 Installing a web server
 -----------------------
-To Deploy OMERO.web, you can use either Nginx or Apache.
-Follow the steps to install your chosen web server.
-
-
-
-Nginx
-^^^^^
-
-See also :doc:`install-web/install-nginx`.
+OMERO.web is deployed using Nginx see :doc:`install-web/install-nginx`.
 
 **The following steps are run as root.**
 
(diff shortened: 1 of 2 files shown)


omero/sysadmins/unix/server-ubuntu1604-ice36.txt
```diff
@@ -0,0 +1,161 @@
...
+-----------------------
+
+OMERO.web is deployed using Nginx see :doc:`install-web/install-nginx`.
+
+**The following steps are run as root.**
+
+Install Nginx 1.8, copy the Nginx OMERO configuration file into the
+Nginx configuration directory, and disable the default configuration:
+
+.. literalinclude:: walkthrough/walkthrough_ubuntu1404.sh
```
mtbc: It's okay that we use the 14.04 one here?

jburel: well spotted

hflynn:
I'm completely failing to see any differences between this walkthrough and the one for Ubuntu 14.04 -  do we expect them to diverge or could we have one walkthrough for both versions?

hflynn:
If we are keeping a separate walkthrough, a link also needs to be added to the top of https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/unix/server-installation.html

jburel:
There is a small diff in the postgres installation
``apt-get -y install postgresql`` will install 9.3 on ubuntu 14.04

maybe we can combine the 2

hflynn:
Ah, I see! Yes, i think we could combine them if the postgres install is the only difference.

jburel:
Closing this PR, will combine the ubuntu installation once gh-1614 is merged

PR #1618, opened 2017-02-17
Add deprecated warning to search bridge page

hflynn:
See https://github.com/openmicroscopy/openmicroscopy/pull/5105 and https://trello.com/c/BKRBO167/220-to-deprecate for context

Staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/developers/Modules/Search/Bridges.html

### DIFF
--- a/omero/developers/Modules/Search/Bridges.txt
+++ b/omero/developers/Modules/Search/Bridges.txt
@@ -1,6 +1,10 @@
 Search bridges
 ==============
 
+.. Warning:: All search bridge classes have been deprecated in 5.3.0. This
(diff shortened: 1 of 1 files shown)


joshmoore:
I would likely change this to "never widely used" (certainly used by Glencoe, and we don't know who might be using it) and perhaps add the fact that it's a performance bottleneck. We could even go so far as to say that it's preventing us from upgrading Hibernate.

hflynn:
@joshmoore hopefully better now

mtbc (approved):

joshmoore:
WFM

PR #1631, opened 2017-03-02
Add `--style json` to cli/fs docs

manics:
Updates mentions of `--style` in the existing docs to match https://github.com/openmicroscopy/openmicroscopy/pull/4906

As far as I can tell these are the only docs that mention `--style` even though the flag is present on several other commands.

Trello: https://trello.com/c/TKhujUiT/217-document-cli-style-json

### DIFF
--- a/omero/sysadmins/cli/fs.txt
+++ b/omero/sysadmins/cli/fs.txt
@@ -33,11 +33,12 @@ The options available to this subcommand are:
 
     Display the help for this subcommand.
 
-.. option:: --style <{plain,csv,sql}>
+.. option:: --style <{plain,csv,json,sql}>
 
     This option determines the output style, tabular `sql` being the default as
     in the previous example. The `csv` style is comma-separated values with an
     initial header row, `plain` is the same as `csv` but without the header row.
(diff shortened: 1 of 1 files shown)


mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/cli/fs.html.

mtbc (approved):
While we're here anyway, do we want to make the use of angle brackets for `--style` consistent on this page?

hflynn:
Consistency would be nice, we don't have the additional `<>` on other options, do we?

hflynn:
Although, I can fix that up myself later if you don't have time now @manics

manics:
Updated (Removed `<>`)

manics:
@mtbc Or did you mean add them?

mtbc:
I don't know. :smiley: We also have options listed on that page like `--order <{newest,oldest,prefix}>`. Whatever leaves it all making sense in the end is fine with me.

manics:
I've removed all  `< >`. This is consistent with the output of the Python argparse `--help` message.

PR #1632, opened 2017-03-02
Add symbols & tables to version requirements

hflynn:
See https://trello.com/c/0rTzMr4p/542-use-a-table-on-the-version-requirements-page

@manics suggested symbols in tables would be easier to read so I've replaced all the support level terminology with symbols and converted the summaries at the top into tables too.

Also a couple of wording tweaks as I was going through.


hflynn:
Staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/version-requirements.html

omero/conf.py
```diff
@@ -73,6 +73,21 @@
...
+.. |Dropped| image:: /images/dropped.png
+              :alt: Dropped
+.. |Recommended| image:: /images/recommended.png
+                  :alt: Recommended
+.. |Supported| image:: /images/supported.png
+                :alt: Supported
+.. |Unsupported| image:: /images/unsupported.png
+                  :alt: Unsupported
+.. |Upcoming| image:: /images/upcoming.png
+               :alt: /images/upcoming.png
```
joshmoore: "Upcoming" found via elinks:

screen shot 2017-03-02 at 15 58 56

manics:
Thanks, it's a lot clearer now! 👍  I still think seven different symbols isn't ideal, but I don't have a better solution, and since it's a definite improvement it should be left for a future discussion.

A couple of unrelated points that have become more apparent:

**Bitness** https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/version-requirements.html#bitness
Mention 64-bit in the table and have 5.3-5.4 as columns instead of rows?

**Components** https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/version-requirements.html#components
- **Python 2.6 / OMERO 5.3**: I think this would look better if there was a line-break between web and server instead of having them side-by-side.
- **Python 3.x**: Shouldn't this be broken instead of unsupported since it's known not to work?

hflynn:
Changed python 3 to broken (cc @jburel )
Can't do linebreaks in list tables so I've made separate columns for OMERO.web and OMERO.server for version 5.3 instead @manics

jburel:
for python 3 this is a bit of both. Marking it as broken works and it is probably clearer

jburel:
it does not. This is more semantics
As i said, I am ok with ``broken``

omero/sysadmins/version-requirements.txt
```diff
@@ -53,28 +80,75 @@ Operating systems
...
+   :header-rows: 1
+
+   * - Bitness
+     - OMERO 5.3
+     - OMERO 5.4
+   * - 32-bit
+     - |Deprecated| for Ice and native code [client]
+     - |Dropped| for Ice and native code [client]
+   * - 64-bit
+     - |Supported|
```
manics: Recommended instead of just supported?

omero/sysadmins/version-requirements.txt
```diff
@@ -502,27 +622,27 @@ OMERO support policies
...
       - to TBA
-      - Supported
-      - Recommended
-      - Supported
+      - |Supported|
+      - |Recommended|
+      - |Supported|
       - :zerocforum:`Ice 3.6.0 <announcements/6631-ice-3-6-0-and-ice-touch-3-6-0-released>`
-        :zerocforum:`Ice 3.6.1 <announcements/45941-ice-3-6-0-and-ice-touch-3-6-1-released>` (broken)
+        :zerocforum:`Ice 3.6.1 <announcements/45941-ice-3-6-0-and-ice-touch-3-6-1-released>` |Broken|
```
manics: screen shot 2017-03-03 at 10 07 30
A minor point: this was a bit confusing when I first saw it since the symbol looks out of place, but after seeing the source code it makes sense. Not a blocker, if you're happy just leave it as it is.

manics (changes requested):

omero/sysadmins/version-requirements.txt
hflynn: Another case where the lack of being able to do line-breaks makes it difficult - I could put this version and the symbol in brackets, would that help?

manics: Like `Ice 3.6.0 [Ice 3.6.1 <!>] Ice 3.6.2 Ice 3.6.3`? I think that's a good idea.

hflynn:
I ended up with rounded brackets because the square ones looked even worse with the alignment offset caused by the symbol but I could also try using a smaller symbol if you think it looks too odd.

manics:
I think it's good enough, no point spending too much time on it. This builds locally for me so good to merge.

hflynn:
👍 I rebuilt staging just to double-check but agree that it's probably not worth messing about with image sizes, it's a big improvement overall.

PR #1666, opened 2017-03-22
homebrew formula doc

jburel:
Due to an error in the formula zeroc-ice35, install with 3.6

To be merged when the formula is available. doc to be generated afterwards
cc @hflynn


jburel:
https://ci.openmicroscopy.org/job/OMERO-DEV-merge-homebrew/ICE=3.6,OSX=snapper/437/ is green 
I will copy the file manually from omero-install to avoid changing the autogen job

PR #1669, opened 2017-03-28
server-security wording s/critical connection/critical data

kennethgillen:
The wording of "critical connection which is passed" seems wrong to me. It looks like it's trying to say "critical data"


hflynn:
For future reference, you didn't need to revert the commit on the wrong branch, you could have just made the fresh branch and cherry-picked the original commit.

PR #1679, opened 2017-04-25
Update IDR URL

hflynn:
I grepped & this is the only place we link to the IDR in the Docs.

Staged at https://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/system-requirements.html


hflynn:
N.B. I checked Help too but there are no references there.

mtbc (approved):

PR #1687, opened 2017-05-10
Add wellRow and wellColumn variables with explanations

bramalingam:
Please check if the comments added above the wellRow and wellColumn variables make sense.

cc: @sbesson @jburel


jburel:
https://github.com/openmicroscopy/openmicroscopy/blob/develop/examples/Training/matlab/ReadData.m should also be updated to be sure it is tested daily

bramalingam:
PR related to the comment above : https://github.com/openmicroscopy/openmicroscopy/pull/5293

PR #1688, opened 2017-05-10
Fix typo on PlaneGenerator

bramalingam:
Small typo fix.

@will-moore @jburel


jburel:
To match the example, you should change the name of the method to ``plane_gen`` i.e. wrong fix
see 
https://github.com/openmicroscopy/openmicroscopy/blob/develop/examples/Training/python/Create_Image.py

jburel:
The name of the function line 1088 ``def planeGen()`` should also be fixed

jburel:
Thanks @bramalingam

PR #1695, opened 2017-05-15
Drop the .web

jburel:
cc @kennethgillen 

I was using the wrong command


kennethgillen:
👍  good to merge, cheers.

hflynn:
Thanks for catching this

PR #1698, opened 2017-06-10
python version requirement

jburel:
Clarify python version requirement
See https://github.com/openmicroscopy/openmicroscopy/pull/5321

cc @mtbc @sbesson

### DIFF
--- a/omero/sysadmins/version-requirements.txt
+++ b/omero/sysadmins/version-requirements.txt
@@ -488,7 +488,8 @@ OMERO support policies
       - from Oct 2008
       - to Oct 2013
       - |Deprecated|
-      - |Dropped| (Web only) |Supported| (*)
+      - * |Dropped| [1]_ 
+        * |Supported| [2]_ 
       - |Dropped|
       - `Reference <http://legacy.python.org/dev/peps/pep-0361/>`__
     * - 2.7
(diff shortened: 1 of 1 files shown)


mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/version-requirements.html#python.

mtbc:
Is it possible to cause an in-cell linebreak before the checkmark?

reqs

mtbc:
(If not then **good to merge**! The row will disappear for 5.4 anyway.)

jburel:
The icons are aligned on staging, not much we can do in the "sphinx" table

mtbc:
I was looking on staging; seems to depend on browser width. But if not much we can do then fine as-is.

jburel:
@mtbc I have used bullet list and reference
let me know if it is better

mtbc:
Alas, the indent from the list makes it worse overall: go ahead and revert or force-push that latest commit away and merge up to https://github.com/openmicroscopy/ome-documentation/pull/1698/commits/aae7b3bfe68857f5c0aba6b5aebac9fc07db4580. Thank you for trying!

jburel:
i will remove the bullet point and keep the ref

mtbc:
Feel free to relist for me, we'll get there in the end!

mtbc (approved):

PR #1699, opened 2017-06-15
open_with update warning

will-moore:
See discussion at https://github.com/openmicroscopy/openmicroscopy/pull/5317/
This PR describes the problem with previously-configured open_with overriding the new 'Image Viewer' config and links to the PR above for how to manually update.

cc @jburel.

### DIFF
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -107,16 +107,13 @@ so that the app is available from the context menu on the webclient tree::
 Updating 'Open with' config
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The default ``Open with`` configuration changed in OMERO 5.3.3 so that
-``Open with > Image Viewer`` *always* uses the ``webgateway`` image viewer instead of
-using the viewer configured by :property:`omero.web.viewer.view`.
-However, the new default ``Open with`` configuration will be ignored if you have previously
-appended any new items to it, since the whole ``open_with`` list
-will then be stored in ``etc/grid/config.xml``.
-
 If you have configured :property:`omero.web.open_with` prior to OMERO 5.3.3 and
-you wish to ensure that ``Open with > Image Viewer`` uses the ``webgateway`` viewer,
-you need to manually update ``open_with`` as
+also set the default viewer with :property:`omero.web.viewer.view`, for example
+as described for `OMERO.iviewer <https://pypi.python.org/pypi/omero-iviewer>`_
+then you will find that ``Open with > Image Viewer`` also opens the OMERO.iviewer
+instead of the ``webgateway`` viewer.
+
+To fix this, you need to manually update ``open_with`` as
 `described on this PR <https://github.com/openmicroscopy/openmicroscopy/pull/5317/>`_.
 
 Memoization files invalidation
--- a/omero/sysadmins/server-upgrade.txt
+++ b/omero/sysadmins/server-upgrade.txt
@@ -108,17 +108,16 @@ Updating 'Open with' config
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The default ``Open with`` configuration changed in OMERO 5.3.3 so that
-``Open with > Image Viewer`` *always* uses the old ``webgateway`` image viewer instead of
+``Open with > Image Viewer`` *always* uses the ``webgateway`` image viewer instead of
 using the viewer configured by :property:`omero.web.viewer.view`.
 However, the new default ``Open with`` configuration will be ignored if you have previously
 appended any new items to it, since the whole ``open_with`` list
-will then be stored in /etc/grid.xml.
+will then be stored in ``etc/grid/config.xml``.
 
 If you have configured :property:`omero.web.open_with` prior to OMERO 5.3.3 and
-you wish to have the old ``webgateway`` viewer available via ``Open with > Image Viewer``
-when the :property:`omero.web.viewer.view` is configured to a different viewer,
+you wish to ensure that ``Open with > Image Viewer`` uses the ``webgateway`` viewer,
 you need to manually update ``open_with`` as
-`described on the PR <https://github.com/openmicroscopy/openmicroscopy/pull/5317/>`_.
+`described on this PR <https://github.com/openmicroscopy/openmicroscopy/pull/5317/>`_.
 
 Memoization files invalidation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


omero/sysadmins/server-upgrade.txt
```diff
@@ -104,7 +104,21 @@ so that the app is available from the context menu on the webclient tree::
...
+
+The default ``Open with`` configuration changed in OMERO 5.3.3 so that
+``Open with > Image Viewer`` *always* uses the old ``webgateway`` image viewer instead of
+using the viewer configured by :property:`omero.web.viewer.view`.
+However, the new default ``Open with`` configuration will be ignored if you have previously
+appended any new items to it, since the whole ``open_with`` list
+will then be stored in /etc/grid.xml.
+
+If you have configured :property:`omero.web.open_with` prior to OMERO 5.3.3 and
+you wish to have the old ``webgateway`` viewer available via ``Open with > Image Viewer``
```
jburel: maybe remove old

omero/sysadmins/server-upgrade.txt
```diff
@@ -104,7 +104,21 @@ so that the app is available from the context menu on the webclient tree::
...
-
+Updating 'Open with' config
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The default ``Open with`` configuration changed in OMERO 5.3.3 so that
+``Open with > Image Viewer`` *always* uses the old ``webgateway`` image viewer instead of
+using the viewer configured by :property:`omero.web.viewer.view`.
+However, the new default ``Open with`` configuration will be ignored if you have previously
+appended any new items to it, since the whole ``open_with`` list
+will then be stored in /etc/grid.xml.
```
mtbc: `:file:` markup -- and really in the top-level `/etc/` ?

joshmoore:
Would a table of example help here? e.g. "If you did X in version Y, then now do Z..."

omero/sysadmins/server-upgrade.txt
will-moore: In the ```OMERO-SERVER/etc/grid.xml``` unless I've got that completely wrong? I just assumed the "omero-server" bit wasn't needed, but I'll add it...

mtbc: `etc/grid.xml` would be fine but `/etc/grid.xml` does suggest it starts at the top of the filesystem.

omero/sysadmins/server-upgrade.txt
will-moore: I just wanted to convey that this wasn't something new (I don't think users will recognise the "webgateway" name). Maybe "existing" or "standard" or "regular"? (don't want "default" as I've used that term elsewhere to refer to configs).

omero/sysadmins/server-upgrade.txt
manics: Shouldn't it be `etc/grid/config.xml`?

omero/sysadmins/server-upgrade.txt
jburel: existing will be better

will-moore:
@joshmoore I'm all for improving clarity but I'm not sure a table would help (unless I'm not quite understanding you)? There's only a single case where you do need to do the manual update, so I'd prefer not to add a matrix of all the other cases where you don't need to.

joshmoore:
Fair enough, but here's a heads up that I found the prose here confusing. Perhaps at least an example of before/after. If we've essentially broken a production configuration value, we should be take care to make sure it's clear.

cc: @kennethgillen

hflynn:
I find this text very confusing too. 

I'm also not sure we want to start a precedent of linking to PRs for instructions rather than just further reading if people are particularly interested in the nuts and bolts.

omero/sysadmins/server-upgrade.txt
will-moore: On second thoughts, I'll simply remove "old" to keep the text simpler.

will-moore:
I have tried to simplify the text a little.
I could make the issue clearer by describing a lot more about ```open_with``` config and the ```viewer.view``` config and how settings are saved to ```etc/grid/config.xml``` and the issues we have with updating single items in the lists (as mentioned on the PR) as well as the fix itself?
However, I was trying not to make this too bloated since it is an edge case and won't apply to most users. I'm hoping that most users will simply read the "If.... " statement and understand that they don't need to do anything.
I'm happy to have suggestions on how to make the text clearer, but "it's confusing" is not so helpful as I've already tried to make it as clear as I can.

will-moore:
As discussed with @hflynn: Move the 'if' statement first, remove the background logic mentioning ```config.xml``` and mention iviewer as example because that's the only case where we've told users to config both webclient.view and open_with.

hflynn:
👍 I think this is clearer

omero/sysadmins/server-upgrade.txt
```diff
@@ -104,7 +104,17 @@ so that the app is available from the context menu on the webclient tree::
     $ bin/omero config append omero.web.open_with '["omero_figure", "new_figure",
       {"supported_objects":["images"], "target": "_blank", "label": "OMERO.figure"}]'
 
+Updating 'Open with' config
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+If you have configured :property:`omero.web.open_with` prior to OMERO 5.3.3 and
+also set the default viewer with :property:`omero.web.viewer.view`, for example
+as described for `OMERO.iviewer <https://pypi.python.org/pypi/omero-iviewer>`_
+then you will find that ``Open with > Image Viewer`` also opens the OMERO.iviewer
```
mtbc, line 113: I think our markup for this is,
```rst
:menuselection:`Open with --> Image Viewer`
```

omero/sysadmins/server-upgrade.txt
```diff
@@ -104,7 +104,17 @@ so that the app is available from the context menu on the webclient tree::
...
+instead of the ``webgateway`` viewer.
+
+To fix this, you need to manually update ``open_with`` as
+`described on this PR <https://github.com/openmicroscopy/openmicroscopy/pull/5317/>`_.
```
joshmoore: I'd strongly suggest that the example be put here in this page and not require users to go to the PR. We should be the ones to go the extra step, not the users. For comparison, I don't see any other mention of a PR in this repo:

```
$ git grep -E "github.*pull" -- .
common/_ext/edit_on_github.py:Loosely based on https://github.com/astropy/astropy/pull/347
contributing/using-git.txt:(PRs) <https://github.com/openmicroscopy/openmicroscopy/pulls>`_. These
contributing/using-git.txt:-  `Using Pull Requests <http://help.github.com/articles/using-pull-requests>`_
contributing/using-git.txt:-  `Pull Requests 2.0 <https://github.com/blog/712-pull-requests-2-0>`_
omero/index.txt:.. _Pull Requests: http://help.github.com/articles/using-pull-requests
```

joshmoore (changes requested):

omero/sysadmins/server-upgrade.txt
jburel: I agree that having the example is better. PR should not be used for reference

jburel:
@joshmoore Happy with the last commit?

joshmoore (approved):

joshmoore:
Yes, thank you.

PR #1704, opened 2017-06-29
add developer docs for light administrator privileges

mtbc:
Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/developers/Server/LightAdmins.html though it is really 5.4-only.

### DIFF
--- a/omero/developers/Server/LightAdmins.txt
+++ b/omero/developers/Server/LightAdmins.txt
@@ -3,12 +3,14 @@ The server's view of administrator restrictions
 
 OMERO 5.4 introduced the concept of a restricted administrator. These
 are generally more powerful than normal OMERO users and group owners but
-they do not have all the powers of full administrators such as ``root``
-because they are restricted in specific ways by their
+they do not have all the powers of full administrators such as ``root``.
+A restricted administrator is the same as a full administrator except
+that *in specific ways* it is as if they are not an administrator at all.
+Those specific ways are exactly the restrictions listed in their user's
 ``Experimenter.config`` property. For that property, a name of
 ``AdminPrivilege:Dance`` (just an example!) with a value of ``false``
-indicates that the user's ability to dance is not that of a full
-administrator but is instead that of a normal user who is not a member
+indicates that the restricted administrator's ability to dance is not that of a full
+administrator. Their ability to dance is instead that of a normal user who is not a member
 of the ``system`` group. A member of the ``system`` group with an empty
 ``config`` property is a full administrator without restrictions.
 
@@ -114,8 +116,8 @@ Working with restrictions
 Restricted administrators
 """""""""""""""""""""""""
 
-The :javadoc:`admin service <slice2html/omero/api/IAdmin.html>` offers
-new operations for managing restrictions on administrators.
+Since OMERO 5.4 the :javadoc:`admin service <slice2html/omero/api/IAdmin.html>` offers
+operations for managing restrictions on administrators.
 ``createLightSystemUser`` creates a new restricted administrator.
 ``getAdminPrivileges`` and ``setAdminPrivileges`` manage the
 restrictions on an existing administrator.
@@ -124,15 +126,18 @@ Using ``setAdminPrivileges`` to set an empty list of privileges fills
 that user's ``Experimenter.config`` property with a ``false`` value for
 every ``AdminPrivilege`` name. That user does not thus become like a
 normal user: they retain all administrative powers not explicitly
-restricted, such as being able to read all users' images. A normal
-user's restrictions can be set: they simply have no effect if the user
-is not an administrator as there is then nothing to restrict.
+restricted, such as being able to read all users' images.
+
+Even for a normal user who is not a member of the ``system`` group and
+has no administrative powers, restrictions can still be set in their
+``Experimenter.config`` property. Such restrictions have no effect while
+that user is not an administrator as they have no administrative powers
+to restrict.
 
 ``getCurrentAdminPrivileges`` is useful for OMERO clients to find how
 the currently logged-in administrator is restricted.
 ``getAdminsWithPrivileges`` identifies the administrators who are
-sufficiently unrestricted in a given way so that users know which
-administrators may be able to help them.
+sufficiently unrestricted in a given way.
 
 
 Permissions on model objects
@@ -140,7 +145,8 @@ Permissions on model objects
 
 OMERO model objects have a :ref:`details property <model details
 property>` that bears information on :ref:`object permissions
-<permissions object>`; the ``canChgrp`` and ``canChown`` methods were
+<permissions object>`. In addition to the existing methods like
+``canEdit`` and ``canDelete``, the ``canChgrp`` and ``canChown`` methods were
 introduced in OMERO 5.4. Client software may find these permissions
 methods a useful guide as to what the current administrator may do to
 which objects.
@@ -155,7 +161,9 @@ available from the admin service, has additional data members:
 
 * ``adminPrivileges`` that lists the restrictions *not* applying to the
   current session. For non-administrators this list is empty as if they
-  are wholly restricted.
+  are wholly restricted. For restricted administrators it lists only the
+  privileges that they enjoy. For full administrators all privileges are
+  listed.
 
 * ``sudoerId``, ``sudoerName`` that for sudo sessions notes which
   administrator it was who became the current user.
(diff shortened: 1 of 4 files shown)


omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
+The server's view of administrator restrictions
+===============================================
+
+OMERO 5.4 introduced the concept of a restricted administrator. These
+are generally more powerful than normal OMERO users and group owners but
+they do not have all the powers of full administrators such as ``root``
+because they are restricted in specific ways by their
+``Experimenter.config`` property. For that property, a name of
+``AdminPrivilege:Dance`` (just an example!) with a value of ``false``
+indicates that the user's ability to dance is not that of a full
```
pwalczysko: From here on, I think this idea needs a reformulation, this sentence is too long. Genuinely do not understand it. Nothing wrong with the ``dance``, but the further discerning of the ``ability``, between ``administrator`` (restricted administrator or full, hypothetical or real ?) and  ``that of a normal user`` is too big an ask.

omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
...
+------------
+
+The :source:`meta.ome.xml mapping
+<components/model/resources/mappings/meta.ome.xml>` lists all the values
+of the :javadoc:`AdminPrivilege
+<slice2html/omero/model/AdminPrivilege.html>` enumeration. If an
+administrator has any of the below restrictions then their ability to
+perform the corresponding operation does not take account of that they
+are an administrator. It instead becomes as if they were but a normal
+user in that respect.
```
pwalczysko: This seems to me like a very clear formulation, maybe take some approaches from here to the ``dance`` part above ?

omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
...
+new operations for managing restrictions on administrators.
+``createLightSystemUser`` creates a new restricted administrator.
+``getAdminPrivileges`` and ``setAdminPrivileges`` manage the
+restrictions on an existing administrator.
+
+Using ``setAdminPrivileges`` to set an empty list of privileges fills
+that user's ``Experimenter.config`` property with a ``false`` value for
+every ``AdminPrivilege`` name. That user does not thus become like a
+normal user: they retain all administrative powers not explicitly
+restricted, such as being able to read all users' images. A normal
```
pwalczysko: ``A normal user's ....`` - I think this is too short, again, I would need a preparatory sentence, "easing" into the idea. ``...as there is nothing to restrict.`` - at this point I am lost.

pwalczysko: Reading it 3 times, I think I understand now. The problem is that in ``A normal user's...`` part it is not totally obvious what and who is the ``normal user``, so maybe starting more slowly like ``If you want to set restriction on a user which is neither an administrator (or a member of system group), this is possible, but kinda makes no sense, because... etc" would be a less elegant, but for me much more understandable construction.

omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
...
+user's restrictions can be set: they simply have no effect if the user
+is not an administrator as there is then nothing to restrict.
+
+``getCurrentAdminPrivileges`` is useful for OMERO clients to find how
+the currently logged-in administrator is restricted.
+``getAdminsWithPrivileges`` identifies the administrators who are
+sufficiently unrestricted in a given way so that users know which
```
pwalczysko: who are ``users`` here ? The same users as mentioned above ? General real life people ? ``Experimenters`` in OMERO ? Would actually prefer if the part ``... so that...`` etc. would get deleted.

omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
...
+administrators may be able to help them.
+
+
+Permissions on model objects
+""""""""""""""""""""""""""""
+
+OMERO model objects have a :ref:`details property <model details
+property>` that bears information on :ref:`object permissions
+<permissions object>`; the ``canChgrp`` and ``canChown`` methods were
```
pwalczysko: Would not go with the semicolon. I think this shoiuld be a new sentence, or some more explanation before the colon, like  "these are the famous ``can*`` methods", which are so helpful (of course this is exagerrated and too clumsy). Would use the word ``additionally`` or ``new`` for the introduction of the ``canChgrp`` and ``canChown``

omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
...
+introduced in OMERO 5.4. Client software may find these permissions
+methods a useful guide as to what the current administrator may do to
+which objects.
```
pwalczysko: Actually this is the sentence which I was looking for in my comment above https://github.com/openmicroscopy/ome-documentation/pull/1704/files#r125008781. Maybe use parts of this to break the ``shorthand`` of previous sentence ?

omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
...
+
+
+Event context
+"""""""""""""
+
+Since OMERO 5.4 the :javadoc:`event context
+<slice2html/omero/sys/EventContext.html>` for the current session,
+available from the admin service, has additional data members:
+
+* ``adminPrivileges`` that lists the restrictions *not* applying to the
```
pwalczysko, line 164: ``restrictions`` and ``not applying`` is a double negative which makes the sentence hard to read. As the name of the ``member`` suggests, these are privileges. Maybe use one more sentence using the word ``privileges`` to ease the reading ?

omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
...
+   <ome/security/basic/OmeroInterceptor.html>` and frequently cleared by
+   :javadoc:`LightAdminPrivilegesCleanup
+   <ome/security/basic/LightAdminPrivilegesCleanup.html>`
+
+An example database trigger would be ``user_config_delete_trigger`` on
+the ``experimenter_config`` table. This trigger raises an exception if,
+for example, an ``AdminPrivilege:Dance`` name with a ``false`` value is
+to be removed from the ``config`` of a member of the ``system`` group by
+a user who themself is restricted from dancing. This then prevents the
+restricted administrator from dancing even vicariously by lifting that
```
pwalczysko: Sorry, I just fathom and guess what ``vicariously`` might mean. The sentence seems understandable, the word is quit e unusual for me.

pwalczysko:
Two general questions:

   - the sysadmin doc uses ``Administrators with restricted privileges``, here ``Restricted Administrators`` are used -> do we want to unify ? (Is it wise to unify ?)
   - should these two docs refer to each other ? (danger of confusion by the implemented "double negatives" is great though

mtbc:
I'd vote for "restricted administrators" as being much less of a mouthful though "Administrators with restricted privileges" might be a useful introduction to the concept. Could unify.

omero/developers/Server/LightAdmins.txt
mtbc: Thanks, I'd not been sure if it would be obvious why one would want such a method.

omero/developers/Server/LightAdmins.txt
bramalingam: Felt the same way, somehow the statement doesn't drill down the point in a simple way (assuming that the example was picked with such intention).

May be something within the list of available options, like chggrp or chown, itself might drive the point better and keep the context within omero?

omero/developers/Server/LightAdmins.txt
```diff
@@ -0,0 +1,244 @@
...
+Restrictions
+------------
+
+The :source:`meta.ome.xml mapping
+<components/model/resources/mappings/meta.ome.xml>` lists all the values
+of the :javadoc:`AdminPrivilege
+<slice2html/omero/model/AdminPrivilege.html>` enumeration. If an
+administrator has any of the below restrictions then their ability to
+perform the corresponding operation does not take account of that they
+are an administrator. It instead becomes as if they were but a normal
```
bramalingam: This point "If an administrator has any of the below....." , was a little difficult to understand. Its formulated very well, but makes me stop and think about it a couple of times, but it would be great if this line could be broken down into multiple sentences.

omero/developers/Server/LightAdmins.txt
```diff
@@ -22,13 +22,13 @@ Restrictions
...
 of the :javadoc:`AdminPrivilege
-<slice2html/omero/model/AdminPrivilege.html>` enumeration. If an
-administrator has any of the below restrictions then their ability to
-perform the corresponding operation does not take account of that they
-are an administrator. It instead becomes as if they were but a normal
+<slice2html/omero/model/AdminPrivilege.html>` enumeration. Each of those
+values corresponds to a kind of operation for which full administrators
+have greater privilege. If an administrator has any of the below
+restrictions then they do *not* have greater privilege for the
+corresponding operation. It instead becomes as if they were but a normal
```
bramalingam, line 29: This sounds great!

mtbc:
@bramalingam: Created https://trello.com/c/kDxd1MeO/57-cross-reference-light-admin-docs for further work. This PR now good to merge?

bramalingam:
yup looks good to merge, if the cross-referencing can be handled in a separate PR.

PR #1707, opened 2017-07-04
Debian9

jburel:
Add a new page for Debian 9 installation
Recent posts on the forum indicates that Debian 9 is used in various locations

The web section will need to be adjusted
This will require some changes to omeroweb-install

cc @mtbc 

https://trello.com/c/00CYQi2t/29-debian-installation-page

### DIFF
--- a/omero/sysadmins/unix/server-debian9-ice36.txt
+++ b/omero/sysadmins/unix/server-debian9-ice36.txt
@@ -1,0 +1,147 @@
+OMERO.server installation on Debian 9
+=====================================
+
+This is an example walkthrough for installing OMERO on Debian 9, using
+a dedicated system user, and should be read in conjunction with
+:doc:`install-web`. You can use this as a guide
+for setting up your own test server. For production use you should also read
+the pages listed under :ref:`index-optimizing-server`.
+
+This guide describes how to install the **recommended** versions, not all
+the supported versions.
+This should be read in conjunction with :doc:`../version-requirements`.
+
+These instructions assume your Linux distribution is configured with a UTF-8
+locale (this is normally the default).
+
+For convenience in this walkthrough the main OMERO configuration options have
+been defined as environment variables. When following this walkthrough you can
+either use your own values, or alternatively source the following file:
+
+.. literalinclude:: walkthrough/settings.env
+   :start-after: Substitute
+
+:download:`settings.env <walkthrough/settings.env>`
+
+
+Installing prerequisites
+------------------------
+
+**The following steps are run as root.**
+
+Install Java 1.8, Ice 3.6 and PostgreSQL 9.6:
+
+To install Java 1.8 and other dependencies:
+
+.. literalinclude:: walkthrough/walkthrough_debian9.sh
+    :start-after: #start-step01
+    :end-before: # install Ice
+
+To install Ice 3.6:
+
+.. literalinclude:: walkthrough/walkthrough_debian9.sh
+    :start-after: #start-recommended-ice
+    :end-before: #end-recommended-ice
+
+To install PostgreSQL 9.6:
+
+.. literalinclude:: walkthrough/walkthrough_debian9.sh
+    :start-after: # install Postgres
+    :end-before: #end-step01
+
+Create an omero system user, and a directory for the OMERO repository:
+
+.. literalinclude:: walkthrough/walkthrough_debian9.sh
+    :start-after: #start-step02
+    :end-before: #end-step02
+
+Create a database user and initialize a new database for OMERO:
+
+.. literalinclude:: walkthrough/walkthrough_debian9.sh
+    :start-after: #start-step03
+    :end-before: #end-step03
+
+
+Installing OMERO.server
+-----------------------
+
+**The following steps are run as the omero system user.**
+
+Download, unzip and configure OMERO. The rest of this walkthrough assumes the
+OMERO.server is installed into the home directory of the omero system user.
+
+Note that this script requires the same environment variables that were set
+earlier in `settings.env`, so you may need to copy and/or source this file as
+the omero user.
+
+You will need to install the server corresponding to your Ice version.
+
+Install ``server-ice36.zip``:
+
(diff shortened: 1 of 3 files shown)


omero/sysadmins/unix/server-debian9-ice36.txt
```diff
@@ -0,0 +1,191 @@
...
+Patching OMERO.server
+---------------------
+
+Weaker ciphers like ADH are disabled by default in OpenSSL 1.1.0,
+version installed on Debian 9.
+This means that it is not possible to connect to an OMERO.server
+using any OMERO clients e.g. the Java Desktop client,
+the OMERO.web client or the CLI.
+The parameter ``@SECLEVEL=0``, enabling the weaker ciphers, needs to be
+added in two files in order to allow connection.
```
mtbc, line 100: This may be too scary to suggest but, for what it's worth, I've been using,
```shell
sed -ie 's/\("IceSSL.Ciphers".*"ADH\b\)/\1:@SECLEVEL=0/' lib/python/omero/clients.py etc/templates/grid/templates.xml
```

omero/sysadmins/unix/walkthrough/walkthrough_debian9.sh
```diff
@@ -0,0 +1,111 @@
...
+#start-recommended-ice
+apt-get -y install libssl-dev libbz2-dev libmcpp-dev libdb++-dev libdb-dev libdb-java
+apt-get -y install zeroc-ice-all-runtime
+pip install "zeroc-ice>3.5,<3.7"
+#end-recommended-ice
+
+
+# install Postgres
+apt-get -y install postgresql
+service postgresql start
```
mtbc: interesting it doesn't start automatically, hadn't noticed that; if not, I wonder if it'd survive a reboot

omero/sysadmins/unix/server-debian9-ice36.txt
jburel: this actually might be good option to use instead of having the line number
actually I should include it the omero-install branch
since the patch will be needed in order to have the server up and running

omero/sysadmins/unix/walkthrough/walkthrough_debian9.sh
jburel: you will need to start pg otherwise the steps  below will not work

mtbc: When I install PostgreSQL with `apt-get install postgres...` then it starts PostgreSQL automatically. If for some reason it isn't in this case then I wonder if it even starts after that machine is rebooted.

mtbc: (I similarly wonder if nginx really doesn't already start after the `apt-get install` and doesn't need the `service nginx start`.)

jburel: That could be a docker specific thing
if I do not start pg
we have the following error:
```
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

jburel: https://github.com/ome/omero-install/pull/158/commits/b28d4ac4dbc9d5b18323e098afc0af5f02164e40 will not include pg start in walkthrough

omero/sysadmins/unix/server-debian9-ice36.txt
jburel: Added patching script in https://github.com/ome/omero-install/pull/158

omero/sysadmins/unix/walkthrough/walkthrough_debian9.sh
manics: Does Debian 9 use systemd? If so it won't be enabled in the default Debian Docker image, though should be in a VM.

mtbc: Yes, I'd expect vanilla Debian 9 to default to systemd as init system. Does this suggest that in some circumstances Docker won't be an accurate test of our install walkthroughs?

manics: The CentOS7 omero-install test uses a systemd-enabled image: https://github.com/ome/omero-install/blob/develop/linux/test/centos7_nginx/Dockerfile#L5 (this is one we wrote ourselves, though CentOS have since released their own systemd Docker image). Could check if Debian has something similar?

jburel: I doubt that something will already be available for debian 9
The change mentioned above will imply that it will work in docker 
but won't show up in the doc
This is probably an happy medium for now
We can revisit before 5.4 goes out.
Something might be available by that time

jburel: note that the change I mentioned above will only be visible in the doc when omero-install PR is merged

omero/sysadmins/unix/server-debian9-ice36.txt
```diff
@@ -0,0 +1,165 @@
...
+
+.. literalinclude:: walkthrough/walkthrough_debian9.sh
+    :start-after: #end-release-ice36
+    :end-before: #end-step04
+
+Patching OMERO.server
+---------------------
+
+Weaker ciphers like ADH are disabled by default in OpenSSL 1.1.0,
+version installed on Debian 9.
```
mtbc: "the version ..."

jburel:
@mtbc 
Staged at http://www.openmicroscopy.org/site/support/omero5.3-staging/sysadmins/unix/server-debian9-ice36.html

jburel:
I will fix the web link in another PR since it requires some changes in another repo

jburel:
https://github.com/ome/omeroweb-install/pull/17 omeroweb-install is now opened
Merging this PR
I will fix the link when the omeroweb-install PR  is in

PR #1708, opened 2017-07-04
Document how to disable unencrypted Ice connections

manics:
Based on internal discussions this is officially supported but not currently documented.

See also https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=8252&p=18063#p18066

# Testing

1. Setup a normal OMERO.server but block port 4063 (e.g. with a firewall). Attempt a remote import, it should fail.
2. Set the configuration property described in this PR and restart the server.
3. A remote import should now work.

### DIFF
--- a/omero/sysadmins/server-security.txt
+++ b/omero/sysadmins/server-security.txt
@@ -190,6 +190,10 @@ all ports with 1, use :property:`omero.ports.prefix`::
 
     $ bin/omero config set omero.ports.prefix 1
 
+You can disable unencrypted connections by redirecting clients to the |SSL|
+port using the server property :property:`omero.router.insecure`::
+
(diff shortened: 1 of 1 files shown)


dominikl:
👍 Looks alright, the described server setting works as well. Good to merge.

Just one remark: Without the server setting you can still import even 4063 is closed, you just have to tick the lock icon in Insight/Importer when connecting to the server. It will only fail when the "lock is open", however with the server setting the import will work in any case.

joshmoore:
This is certainly a downside of this config property is that it doesn't somehow automatically disable the lock. An alternative (on the server-side) would be to return an empty string and teach the clients to ignore the empty property.

PR #1710, opened 2017-07-05
Bump version in contributing

jburel:
upgrade version


sbesson:
Discussed earlier today, a following adjustment will be required as the `site/support/omero5.4{-staging}`  URLs will be decommissioned. To be opened as a follow-up PR.

PR #1718, opened 2017-07-19
Improve description of pixeldata properties

sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2017-July/006561.html

This PR aims at explaining the intent of the `omero.pixeldata.max_plane_{width,height}` configuration properties.

Opened initially against ome-documentation to review the wording. Next this should be properly opened against `etc/omero.properties`.

### DIFF
--- a/omero/sysadmins/config.txt
+++ b/omero/sysadmins/config.txt
@@ -1180,12 +1180,22 @@ Default: `pixelDataEventLogQueue`
 omero.pixeldata.max_plane_height
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+With :property:`omero.pixeldata.max_plane_width` specifies
+the plane size cutoff above which a pixel pyramid should
+either be read directly from the file format or generated
+by the pixeldata service.
+
 Default: `3192`
 
 .. property:: omero.pixeldata.max_plane_width
 
 omero.pixeldata.max_plane_width
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With :property:`omero.pixeldata.max_plane_height` specifies
+the plane size cutoff above which a pixel pyramid should
+either be read directly from the file format or generated
+by the pixeldata service.
 
 Default: `3192`
 


omero/sysadmins/config.txt
```diff
@@ -1180,13 +1180,23 @@ Default: `pixelDataEventLogQueue`
 omero.pixeldata.max_plane_height
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+With :property:`omero.pixeldata.max_plane_width` specifies
+the plane size cutoff above which a pixel pyramid should
+either be read directly from the file format or generated
+by the pixeldata service.
```
mtbc, line 1186: So for planes smaller than this size will it never read pyramids from the file format?

rleigh-codelibre:
Looks fine to me.  You might want a comma before `specifies` for both paragraphs.

omero/sysadmins/config.txt
sbesson: That's a good question. I could rephrase this and limit the scope to the server aspect of things e.g. `the plane size cutoff above which a pixel pyramid will be generated by the pixeldata service unless subresolutions can be read from the file format`.

When it comes to reading, I think  services like rendering are re-using the same cutoff to determine whether an image has pyramidal levels or not. Hence for our own clients, your statement should be true. That being said, it might be possible to use the low-level API to access subresolutions if they exist.

mtbc: I do like the direction of your suggested rephrasing but am also happy not to block this PR going in as-is.

joshmoore:
This will also need opening against etc/omero.properties

jburel:
It might be better to modify ``autogen_docs`` so that the file is copied and used ``literalinclude`` to include the desired section

sbesson:
As mentioned in the description, once we collectively agree on the text content, the next step is definitely to open these changes against `etc/omero.properties`. Unless there is a better suggestion, I will push an extra commit implementing https://github.com/openmicroscopy/ome-documentation/pull/1718#discussion_r128442943 for review

jburel:
Could we not only do the change in ``omero.properties``
and modify autogen_docs
since we already do that for ``history.txt``

sbesson:
OK probably a misunderstanding here, yes the change only needs to be done in `etc/omero.properties. The autogeneration logic already completely generates `sysadmin/config.txt` from the content of `etc/omero.properties` and `omeroweb/settings.py` - see https://github.com/openmicroscopy/ome-documentation/blob/develop/omero/autogen_docs#L18. 

Closing this and I will re-open against etc/omero.properties to minimize the confusion.

PR #1720, opened 2017-07-20
add note about needing mock module for tables tests

mtbc:
Says what to do if `tables.test()` fails with the mock error. Cc: @bramalingam.

### DIFF
--- a/omero/sysadmins/server-tables.txt
+++ b/omero/sysadmins/server-tables.txt
@@ -80,6 +80,16 @@ After that, the following should succeed:
     Detected cores:    8
     …
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
Looks fine and good to merge, thanks.

jburel:
After the numpy/pillow saga, the decision was to avoid installation from pypi when possible
e.g. for ubuntu use ``apt-get install python-mock``.

mtbc:
Adjusted accordingly, thank you. (Certainly don't want to get into a per-OS discussion in a brief note though!)

jburel:
@mtbc Thanks for the adjustment

PR #1721, opened 2017-07-31
Improve accuracy of logback.xml reload explanation

manics:
The current docs say `Entry.java` automatically reloads the `logback.xml` config. As far as I can tell it's the `logback.xml` file itself that ensures it's automatically reloaded at regular intervals.


mtbc:
:+1: for the accuracy but we seem to have lost the hint that a change may not take immediate effect.

manics:
@mtbc How does this sound?

mtbc:
Certainly addresses my concern, thank you.

kennethgillen:
👍  - it's got Mark's OK, too, so good to merge.

PR #1723, opened 2017-08-01
Update auto-generated documentation

snoopycrimecop:
Repository: openmicroscopy/ome-documentation
Already up-to-date.

Generated by build [OMERO-DEV-latest-docs-autogen#924](https://ci.openmicroscopy.org/job/OMERO-DEV-latest-docs-autogen/924/).

----
--no-rebase

PR #1724, opened 2017-08-01
Fix host

jburel:
wrong host was used
Only noticed while looking at https://github.com/openmicroscopy/ome-documentation/pull/1723
cc @kennethgillen


jburel:
Will merge when travis is green so we can move forward with the doc

jburel:
better now https://github.com/openmicroscopy/ome-documentation/pull/1723

PR #1727, opened 2017-08-04
Chown doc

pwalczysko:
See https://trello.com/c/JWI2G9rm/35-document-chown-for-cli

Adds missing documentation for the ``chown`` command on CLI.

Used the start on the form of the doc by @mtbc as described in the [trello comment](https://trello.com/c/JWI2G9rm/35-document-chown-for-cli#comment-59760178260a7f3fc885c829).

To test:

Check that the build is passing, check wording and sensibility. 

@mtbc @jburel


omero/sysadmins/cli/index.txt
```diff
@@ -13,6 +13,7 @@ need.
     config
     admin
     usergroup
+    data-ownership
```
joshmoore: I'd suggest we use the same strategy for naming here: either hypenated or camelcase. cc: @hflynn

pwalczysko:
@joshmoore @hflynn : The last commit should address @joshmoore 's comment - went for hyphenated form, as this form seems to be prevalent in other ``intex.txt`` type of files here.

omero/users/cli/chown.txt
```diff
@@ -0,0 +1,176 @@
...
+
+    $ bin/omero chown 10 Experimenter:1,3,7
+
+Here ownership of all the objects belonging to users 1,3 and 7
+would be transferred to user 10.
+
+Including and excluding objects
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. program:: chown
```
sbesson: This should probably be `.. program:: omero chown` else it seems we are documenting the system level `chown` command

omero/sysadmins/cli/data-ownership.txt
```diff
@@ -0,0 +1,53 @@
...
+^^^^^^^^^^^^^^^^
+
+* :doc:`OMERO.cli </users/cli/chgrp>`
+* :help:`OMERO.web <sharing-data#moving>`
+
+
+Giving data to another user
+---------------------------
+
+Who may give data
```
joshmoore: Has "give" here been changed to "transfer" in other locations?

omero/users/cli/chown.txt
```diff
@@ -0,0 +1,176 @@
...
+    $ bin/omero chown 5 Project/Dataset/Image:51
+
+Here the all of images under the project 51 would be transferred. It is not
+necessary to specify intermediate objects in the hierarchy and so::
+
+    $ bin/omero chown 5 Project/Image:51
+
+would have the same effect as the call above.
+
+Transferring lower level objects
```
joshmoore: This heading needs updating

joshmoore:
Barring the one heading change, this looks generally good. I do wonder if "transfer" isn't better than "give" (across the board) though.

hflynn:
👍 to using hyphenated file names rather than camelcase

I agree with Josh that "transfer" is better than "give"

omero/sysadmins/cli/data-ownership.txt
```diff
@@ -0,0 +1,53 @@
...
+administrators must sometimes move data to a different group and/or give
+it to a different user.
+
+.. warning::
+
+  Administrators may move data into a group of which the data owner is
+  not a member. Also, administrators may give data to a user who is not
+  a member of the data's group. These actions may be reasonable steps in
+  a larger workflow but do note the potential to confuse users who find
+  that they own data that is not in any of their groups.
```
will-moore: I think this is a little confusing. Here's my attempt at it...
"Data does not need to be assigned to a group where the data owner is a member, and administrators may wish to change the ownership of data or move it between groups in several steps of a larger workflow. However, it is generally expected that data should end up in a group where the data owner is a member, so that they can view their data in the OMERO clients."

will-moore:
I'm finding it confusing to have 2 similar pages "Command Line Interface as an OMERO client" and "Command Line Interface as an OMERO admin tool" that look very similar and have some of the same links. It's easy to follow a link from one of these "parent" pages, then click on the link trail on the child page to return to the parent, only to find that the page is the other parent (but not obviously different).

Maybe the "CLI as admin tool" page should not contain links to pages that are also under "CLI as client" page since there is already a link to that page. Do we need to make this more obvious? I missed it since it's in "See also" instead of the main list of commands.

We can then remove the ```sysadmins/cli/data-ownership.txt``` page with the info it contains about who can chown/chgrp being moved down to the respective child pages.
The warning about putting data in group where owner is not a member is applicable to both the chown and chgrp child pages.

The link on "CLI as client" page to "Giving objects to other users" and the naming of that page are inconsistent with the link on "CLI as admin tool" ("Changing the ownership of data") as well as the command itself: "chown". This would make a better title for that page.

omero/sysadmins/cli/data-ownership.txt
```diff
@@ -0,0 +1,53 @@
+Changing the ownership of data
```
hflynn: This also talks about moving data between groups which isn't obviously about ownership - maybe it would be clearer to call this 'Data management' to match the 'User/group management' title also used in the admin section.

That also means you could instead use this title or one more like it for the user page

will-moore: My suggestion was to delete this page entirely and simply link directly from the "CLI and a client" page to the pages named "Changing the ownership of data" and "Moving data between groups"?

omero/sysadmins/cli/index.txt
mtbc: I was just following the pattern in `developers/Server/` so there may be a more general issue here, outwith this PR.

mtbc:
Transfer/give: Depends somewhat on context: for more general data-management discussion, "transfer" is vague and could easily mean "chgrp" without extra clarification. Whereas, "chown" changes the ownership of things from somebody to somebody else, which when we do in real life we call it giving. So, I've been generally favoring "give" as being both simpler and more specific. Just my 2¢, ICBW.

omero/sysadmins/cli/data-ownership.txt
hflynn: Yeah, that probably works actually

joshmoore:
Other options: bestow, cede, consign, hand over, ...

will-moore:
"Changing owner(ship)" is preferable to "giving" since giving implies a gift and that the data belongs to the person doing the giving, when it could actually be the sysadmin who is changing owner from one user to another. Also, if I'm "giving" data to one user, I must be "taking" it from the other user, but this language is a bit too possessive!

mtbc:
Hmm, I can't give my son's stuff to my daughter once she outgrows it? Happy to accept I may just be unusual in how I define it. :+1: for @joshmoore thinking of less-vague alternatives!

pwalczysko:
As for ``give`` or ``transfer``, I opted for ``change ownership`` in the headers and for transfer in all more technical-exampley bits. Reasons:
  - majority of reviewers here seem to be against ``give``, and so am I
  - the ``transfer`` is strictly used in the ``omero chown -h`` text which was introduced by @ximenesuk 
  - I think inside the example-bits, it is really clear that the chown part is being read and what the word transfer means there 


I did not implement all the changes suggested by @will-moore in [the big comment](https://github.com/openmicroscopy/ome-documentation/pull/1727#issuecomment-321515190), just the renaming (see above in this comment).
Reasons:
   - the two docs, the ``cli.txt`` and ``chown.txt`` which should take the whole brunt of the cli - chown and cli-chgrp workflow are living under ``users/....`` -> putting the clauses about admins moving data out of groups where the owner is not a member is quite confusing
    - just to contradict myself in the previous bullet point, maybe the ``chown.txt`` and ``chgrp.txt`` should actually not live under ``users/...`` because no user can do any chown at all and even for chgrp on Cli there is a question which "normal user" would use CLI at all for moving just their own data around
    - so to summarize the two bullet points, the actions suggested by @will-moore probably need more thinking, about our docs structure

pwalczysko:
The last 6 commits should implement the [big comment](https://github.com/openmicroscopy/ome-documentation/pull/1727#issuecomment-321515190) of @will-moore

pwalczysko:
@will-moore : Fixed the conflicts , thanks to @sbesson

will-moore:
Looks much better, thanks.
Good to merge.

jburel:
Merging so we can move forward with file extension change
see https://github.com/openmicroscopy/ome-documentation/pull/1729

PR #1728, opened 2017-08-09
Version requirements

jburel:
Following today's discussion during sysadmin meeting
Update the requirement page

General idea: status quo

cc @joshmoore @sbesson


joshmoore:
Ice for 5.5 has no option which is not suboptimal?

jburel:
that's a mistake when I was copying and pasting columns

jburel:
Closing this PR for now 
so file extension can go in
https://github.com/openmicroscopy/ome-documentation/pull/1729

mtbc:
I knew that there was some present issue somewhere within our CI with getting Ice 3.6 working but is there a bigger problem that requires us to officially offer the community support for Ice 3.5 for the duration of OMERO 5.4?

jburel:
Travis is another problem, ice35 only is available and I hope that when issues with travis are fixed we can roll back the changes made in https://github.com/openmicroscopy/openmicroscopy/pull/5409 (similar on marshal)
Note that we have not yet been able to use the same approach for omego (See https://github.com/ome/omego/pull/110 and https://github.com/manics/omego/pull/1 for more info) so dropping ice 3.5 will imply removing travis from our validation process
We could certainly use 3.6 in some situations on travis but it might not always be possible

rleigh-codelibre:
@jburel Custom build of ice 3.6 for travis as we did for 3.5?

mtbc:
It seems like difficulty with testing Ice 3.6 is causing us to decide to publicly support Ice 3.5 *and* Ice 3.6 which just additionally commits us to keep testing Ice 3.5 even after CI is fixed? This is why I was thinking that maybe the problems with Ice 3.6 must have implications rather beyond the current state of our CI. 

Why does dropping official support for Ice 3.5 imply removing travis from our validation process? If we officially support Ice 3.6 then that *still* needs testing somehow and if travis passes using Ice 3.5 then it's at least far more likely that Ice 3.6 is good so CI still brings much value.

jburel:
I will avoid making custom build again if possible
When we indicate "dropped" in that document we usually remove any trace of it in the code => ice 3.5 won't work

Probably more suitable to move the discussion via voice: a topic for monday 5.4 chat

mtbc:
Ah, thank you, that is the missing puzzle piece for me: I hadn't realized the importance of being prompt in removing traces of dropped support.

hflynn:
What's the status of this? Is chat still needed?

mtbc:
No blockers from me.

jburel:
I need to do a final check and will list it for review

jburel:
Ice version was a point of discussion
After second though, I will keep it as "deprecated" for 5.4. Our download and doc encourage people to ice 3.6 but we have not done anything at all in the code so technically still functional and 
tested via travis
Not marking it as "dropped" is to avoid confusion as we have seen in the past when we marked pg 9.1 as dropped even if omero was still working with that version.

The discussion is certainly a good start for min requirements for 5.5

omero/sysadmins/version-requirements.rst
```diff
@@ -62,15 +65,19 @@ Operating systems
      * - Linux version
        - OMERO 5.3
        - OMERO 5.4
+       - OMERO 5.5
      * - 12.04
        - |Dropped|
        - |Dropped|
+       - |Dropped|
```
joshmoore: Since nothing is supported from `12.04`, should we unlist it?

joshmoore (commented):
Also: 

* The description block at the top needs information on `5.5` being tentative.
 * Perhaps also drop 5.2 from the RHEL and Ubuntu sections under "OS support"

omero/sysadmins/version-requirements.rst
jburel: good idea

jburel:
@joshmoore: adjustments made

joshmoore:
The latest adjustments certainly look good. It might be worth someone with a fresh eye reading over this. If not, I'll give my eyes a day to uncross and then will give it another go.

mtbc:
Definitely easy to get lost in, I'd advise somebody else's eyes before this is approved.

Summary seems to be that for 5.4 we'll still tolerate CentOS 6, 32-bit clients, PostgreSQL 9.3, Ice 3.5.

There seems to be some inconsistency about if PostgreSQL 9.4 is deprecated for OMERO 5.4: compare the two tables.

The jump between versions and components as one reads across this table is kind of confusing, but it's not a blocker:
python

jburel:
Thanks @mtbc i will review it.
We improved the readability of the page for 5.3 but still some improvements
When we "split" as discussed the web and server this will probably make things a bit clearer

jburel:
@mtbc I have made few adjustments. I will relist it

omero/sysadmins/version-requirements.rst
```diff
@@ -668,7 +668,8 @@ OMERO support policies
       - :zerocforum:`3.6.0 <announcements/6631-ice-3-6-0-and-ice-touch-3-6-0-released>`
         (:zerocforum:`3.6.1 <announcements/45941-ice-3-6-0-and-ice-touch-3-6-1-released>` |Broken|),
         :zerocforum:`3.6.2 <announcements/46347-ice-ice-e-and-ice-touch-3-6-2-released>`,
-        :zerocforum:`3.6.3 <announcements/46475-ice-ice-e-and-ice-touch-3-6-3-released>`
+        :zerocforum:`3.6.3 <announcements/46475-ice-ice-e-and-ice-touch-3-6-3-released>`,
+        :zerocforum:`3.6.4 <announcements/46550-ice-ice-e-and-ice-touch-3-6-3-released>`
```
mtbc: Should be `discussion/46550/ice-and-ice-touch-3-6-4-released`.

jburel: This is in prep.Planning to work 3.6.4 PR today to re-activate the build

mtbc:
> seems to be some inconsistency

Now fixed, ta.

omero/sysadmins/version-requirements.rst
mtbc: It turns out that having the "wrong" text (3-6-3) after the article number doesn't actually break the link.

mtbc:
We now seem to have inconsistency among tables over how OMERO 5.3, 5.4 regard nginx 1.8. (I'm wondering if these duplicate tables are more trouble than they are worth!)

mtbc:
Worth including https://github.com/openmicroscopy/openmicroscopy/pull/5481#issuecomment-330764699 note?

mtbc:
> now seem to have inconsistency

Looks to be fixed, ta.

jburel:
@mtbc I added a note for pytables version already
This has never been on the version requirement page
see https://github.com/openmicroscopy/ome-documentation/pull/1739/files#diff-4f1690607b26288b56afc5164c52dce0

mtbc (approved):

hflynn:
Anything else here or is this good to merge now @jburel ?

jburel:
waiting for 3.6.4 to be tested since it is in the list

jburel:
ice 3.6.4 PR has been merged
I think we are good now @hflynn

PR #1734, opened 2017-08-15
Url updates

hflynn:
See https://trello.com/c/MawOtyOJ/109-review-docs-intro-pages-external-urls-in-light-of-new-website

Remaining hardcoded links in the history will be fixed on a following code PR, leaving just a version 5.0 specific docs link which hasn't been unpacked to the new docs site yet (cc @sbesson) and the logo link issue mentioned on https://trello.com/c/92AzORoa/103-check-doc-links which is on /sysadmins/config.rst


omero/conf.py
```diff
@@ -46,10 +46,9 @@
     release = 'UNKNOWN'
 
 rst_prolog = """
-.. note:: **This documentation is for the new OMERO 5.4 version.** See the `latest OMERO 5.3.x
-    version <http://www.openmicroscopy.org/site/support/omero5.3/>`_ or the
-    :legacy_plone:`previous versions <>` page to find documentation for the
-    OMERO version you are using if you have not upgraded yet.
+.. note:: **This documentation is for the new OMERO 5.4 version.** See the
```
hflynn: I updated this for now but then I see @sbesson has suggested removing it and adding the docs index link to the side bar instead - certainly open to doing that

omero/downloads/cli/help.out
```diff
@@ -47,5 +47,5 @@ Examples:
   $ importer-cli -s localhost -u username -w password -d 50 --debug ALL foo.tiff
 
 For additional information, see:
-http://www.openmicroscopy.org/site/support/omero5.2/users/cli/import.html
+http://docs.openmicroscopy.org/latest/omero/users/cli/import.html
```
hflynn: @jburel has this been updated in the code?

omero/themes/globalomerotoc.html
```diff
@@ -1,5 +1,5 @@
 <h3><a href="{{ pathto(master_doc) }}">{{ _('OMERO') }}</a></h3>
 {{ toctree()}}
 <a href="http://downloads.openmicroscopy.org/latest/omero/">{{ _('Downloads') }}</a></br>
-<a href="//www.openmicroscopy.org/site/products/omero/feature-list">{{ _('Feature List') }}</a></br>
-<a href="//www.openmicroscopy.org/site/about/licensing-attribution">{{ _('Licensing') }}</a>
+<a href="https://www.openmicroscopy.org/omero/new/">{{ _('Features') }}</a></br>
```
hflynn: I can add the docs index here too if the consensus is to remove the versioning banner as Seb suggested on the card

omero/downloads/cli/help.out
jburel: It should have been omero5 and not omero5.2 so it did get updated
Point at the latest makes more sense than to a specific version

contributing/conf.py
```diff
@@ -50,9 +50,9 @@
     'figure_scc_branch' : (scc_github_root + '/figure/tree/%s', ''),
     
     # Doc links
-    'omero_doc' : (oo_site_root + '/support/omero/%s', ''),
-    'bf_doc' : (oo_site_root + '/support/bio-formats/%s', ''),
-    'model_doc' : (oo_site_root + '/support/ome-model/%s', ''),
+    'omero_doc' : (docs_root + '/omero/%s', ''),
```
sbesson: These would need to be updated to use the `latest/{omero,bio-formats,ome-model}` redirects

sbesson (changes requested):
One issue with extlink that breaks the contributing documentation linkcheck.

Older OMERO documentation sets have been unpacked (since 4.4.9) so the 5.0 server upgrade page should be available at http://docs.openmicroscopy.org/omero/5.0.0/sysadmins/server-upgrade.html.

common/conf.py
```diff
@@ -150,22 +151,15 @@ def get_previous_version(majornumber, minornumber=0):
...
-    'cpp_plone' : (oo_site_root + '/products/ome-files-cpp/%s', ''),
+    # Website links
+    'community' : (oo_root + '/support/%s', ''),
+    'omero' : (oo_root + '/omero/%s', ''),
+    'bf' : (oo_root + '/bio-formats/%s', ''),
     # One branch only doc links. Branched docs links in conf.py files for
     # individual doc sets
-    'model_doc' : (oo_site_root + '/support/ome-model/%s', ''),
-    'devs_doc' : (oo_site_root + '/support/contributing/%s', ''),
+    'model_doc' : (docs_root + '/ome-model/5.5.7/%s', ''),
```
hflynn, line 160: latest is in place for model docs yet so I've added a versioned link to fix the build

contributing/conf.py
```diff
@@ -50,9 +50,9 @@
     'figure_scc_branch' : (scc_github_root + '/figure/tree/%s', ''),
     
     # Doc links
-    'omero_doc' : (oo_site_root + '/support/omero/%s', ''),
-    'bf_doc' : (oo_site_root + '/support/bio-formats/%s', ''),
-    'model_doc' : (oo_site_root + '/support/ome-model/%s', ''),
+    'omero_doc' : (docs_root + '/latest/omero/%s', ''),
+    'bf_doc' : (docs_root + '/latest/bio-formats/%s', ''),
+    'model_doc' : (docs_root + '/ome-model/5.5.7/%s', ''),
```
hflynn, line 55: again, versioned until a latest redirect is in the place

hflynn:
https://ci.openmicroscopy.org/view/Failing/job/CONTRIBUTING-merge-docs/1147/ is green again now although there are still a bunch of old URLs being used to link to minutes and team stuff that I'll need to update in a following PR.

omero/developers/json-api.rst
```diff
@@ -210,7 +210,7 @@ It uses the python ``requests`` library to connect to the JSON api, login, query
 These steps are covered in more detail below.
 
 For an example how to use the API with Java,
-see :omedocs:`blob/develop/omero/examples/java/JSONClient.java`.
+see :omedocs:`JSONClient.java <blob/develop/omero/examples/java/JSONClient.java>`.
```
hflynn: @sbesson I think this will work fine when merged because it works locally, but it's making the merge build yellow because the path isn't right for snoopy's branch - is there any way to fix this other than hardcoding the link instead?

hflynn:
Remaining plone links in the OMERO docs are to the script sharing page - are we just going to ditch that @jburel ?
N.B. `:about_plone:` will go as soon as I update the history page in the code

omero/developers/json-api.rst
sbesson: Someway I had missed we are now including example files in this repository. In this case, instead of hardcoding the GitHub URL and dealing with branch, it might be simpler to use ```:download:`JSONClient.java </examples/java/JSONClient.java>` ``` as we do for the Bio-Formats examples for instance.

sbesson:
One comment regarding the download link (use `downloads` rather than hyperlink). Otherwise http://docs.openmicroscopy.org/omero/5.0.0/sysadmins/server-upgrade.html should now be available for referencing.
How do you want to coordinate wrt https://github.com/openmicroscopy/ome-documentation/pull/1734#issuecomment-323043902. Should we open an auto-generated documentation PR before/after this PR or combine them together (there are a few stagingchanges )?

omero/developers/json-api.rst
hflynn: It was added fairly recently when the linkcheck wasn't working so slipped through.

hflynn:
@sbesson I could open the PR, merge it and then merge develop into this branch and do the remaining fixes?

sbesson:
If things merge cleanly in this order, your suggestion should probably be the safest route.

hflynn:
Slight delay as the autogen PR needs https://github.com/openmicroscopy/openmicroscopy/pull/5456 merging to pass Travis. I also notice the help output fix is there so I'll remove it here so there's no conflict

hflynn:
With this last commit, there should now be no plone links in the OMERO docs except those referring to the secvuln alias (which will be updated via a redirect and not require further work on this repo).

Only remaining plone links are now in the contributing docs with the extlinks referring to them migrated down to the conf.py within that folder. I will fix those on a separate PR when we have a new workflow in place for meeting notes.

sbesson (approved):

sbesson:
The last changes all make sense. Merging to reduce the [number of broken links in the daily docs](https://ci.openmicroscopy.org/job/OMERO-DEV-latest-docs/327/warnings4Result/) jobs. A couple of follow-up actions/thoughts:

- as discussed this morning the broken help links could be solved by adding stub anchors for now (and removing them in favor of the new sections when the help is promoted with the 5.4 additions)
- the broken Ice anchors might have to be reviewed - might be solved by the Sphinx 1.6 upgrade
- in a couple of places, we are linking to the raw OMERO downloads URL while explicitly mentioning a section on a page (source code, Java...). Linking to the www.openmicroscopy.org would better match the test but the links would become obsolete as new versions are released.

PR #1735, opened 2017-08-16
rename of graphical client for IceGrid

mtbc:
Probably don't need to specifically point to the 3.6 docs as despite the name change the 3.7 instructions look to be just fine for getting it going and basic use of it.

### DIFF
--- a/omero/sysadmins/grid.rst
+++ b/omero/sysadmins/grid.rst
@@ -64,8 +64,8 @@ JAR is provided in the OMERO source code under :file:`lib/repository`.
 
     :zerocdoc:`icegridadmin Command Line Tool <display/Ice/icegridadmin+Command+Line+Tool>`
         Chapter of the ZeroC_ manual about the :command:`icegridadmin` CLI
-    :zerocdoc:`IceGrid Admin Graphical Tool <display/Ice/IceGrid+Admin+Graphical+Tool>`
-        Chapter of the ZeroC_ manual about the GridGUI tool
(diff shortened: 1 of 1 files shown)


sbesson:
The new links work as expected . With regard to 3.6/3.7 discussion, I realize that all the `zerocdoc` links are pointing at the unversioned Ice documentation which now redirects to 3.7. Given the [requirements for 5.4](https://github.com/openmicroscopy/ome-documentation/pull/1728), I wonder whether we should review this policy and review our ZeroC extlinks to make them point at our recommended Ice documentation instead i.e.
`zeroc` -> `https://doc.zeroc.com/display/Ice36/%s` in `common/conf.py`.

mtbc:
Just to warn, that may well mean that all the links from old docs would break: ZeroC don't seem to keep docs for old versions up in the way that many other projects do, at least I can't find HTML docs for Ice 3.3.

sbesson:
Fair point, it would certainly be a shame to start linking to resources which are meant to disappear. From https://doc.zeroc.com/spacedirectory/view.action, it seems that all documentation sets since Ice 3.4 are preserved at the moment. Happy to hear secondary thoughts @hflynn @jburel @joshmoore

mtbc:
Part of my thinking is that usually the Ice docs mostly don't change all that much and we will sometimes (often?) be supporting multiple versions anyway.

sbesson:
Raised the issue this morning during the OMERO 5.4 review call. Either way, some maintenance will be required either to update links when a new Ice is released as in the current case or when updating the Ice version is we were to use versioned links. Given that no-one strongly objected to preserving the statu quo, merging as it is to fix the broken link.

PR #1740, opened 2017-08-24
doc extlink URL fixes

hflynn:
Updates the Model docs extlink to use the latest redirect now that works and also I spotted that I'd updated the BF doc link in the contributing docs conf.py but not the OMERO one so I've moved it to common conf.py (since both doc sets use it) and updated it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -436,8 +436,8 @@ ZeroC                       https://zeroc.com
 Ice                         https://zeroc.com
 Jenkins                     http://jenkins-ci.org
 roadmap                     https://trac.openmicroscopy.org/ome/roadmap
-Open Microscopy Environment http://www.openmicroscopy.org
-Glencoe Software, Inc.      http://www.glencoesoftware.com/
+Open Microscopy Environment https://www.openmicroscopy.org
+Glencoe Software, Inc.      https://www.glencoesoftware.com/
 =========================== ==============================================
 
 Abbreviations
--- a/README.rst
+++ b/README.rst
@@ -436,7 +436,7 @@ ZeroC                       https://zeroc.com
 Ice                         https://zeroc.com
 Jenkins                     http://jenkins-ci.org
 roadmap                     https://trac.openmicroscopy.org/ome/roadmap
-Open Microscopy Environment http://www.openmicroscopy.org/site
+Open Microscopy Environment http://www.openmicroscopy.org
 Glencoe Software, Inc.      http://www.glencoesoftware.com/
 =========================== ==============================================
 
--- a/common/conf.py
+++ b/common/conf.py
@@ -155,10 +155,10 @@ extlinks = {
     'community' : (oo_root + '/support/%s', ''),
     'omero' : (oo_root + '/omero/%s', ''),
     'bf' : (oo_root + '/bio-formats/%s', ''),
-    # One branch only doc links. Branched docs links in conf.py files for
-    # individual doc sets
-    'model_doc' : (docs_root + '/ome-model/5.5.7/%s', ''),
+    # Doc links
+    'model_doc' : (docs_root + '/latest/ome-model/%s', ''),
     'devs_doc' : (docs_root + '/contributing/%s', ''),
+    'bf_doc' : (docs_root + '/latest/bio-formats/%s', ''),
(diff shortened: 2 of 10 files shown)


common/conf.py
```diff
@@ -155,10 +155,10 @@ def get_previous_version(majornumber, minornumber=0):
     'community' : (oo_root + '/support/%s', ''),
     'omero' : (oo_root + '/omero/%s', ''),
     'bf' : (oo_root + '/bio-formats/%s', ''),
-    # One branch only doc links. Branched docs links in conf.py files for
-    # individual doc sets
-    'model_doc' : (docs_root + '/ome-model/5.5.7/%s', ''),
+    # Doc links
+    'model_doc' : (docs_root + '/latest/ome-model/%s', ''),
     'devs_doc' : (docs_root + '/contributing/%s', ''),
+    'bf_doc' : (docs_root + '/latest/bio-formats/%s', ''),
```
sbesson: Is this used in the contributing documentation only or in the OMERO documentation? In the former, this makes sense as the contributing documentation largely documents our latest software. For OMERO, arguably, we should be able to point at the documentation for the specific version of Bio-Formats shipped with OMERO.

hflynn: It's in both but I can make the OMERO one version specific if we prefer.

hflynn: That would mean we'd need to update it regularly though.

sbesson: Agreed. The main reason for hardcoding from my side is that https://github.com/openmicroscopy/ome-documentation/blob/v5.4.0-m3/omero/users/cli/import.rst refers to supported formats table which is tightly coupled to the the version included into OMERO.

Some of the autogeneration scripts read properties from the OMERO version (patch versions) for consumption in the docs, should we consider adding the bf_version there as well to make sure it's updated via the auto-generation PR?

hflynn:
BF doc extlinks moved back into their respective subfolders so we can use latest for the contributing docs and an autogen'ed BF version for OMERO.

Likely we'll want to have similar for the Model docs in future.

README.rst
```diff
@@ -436,7 +436,7 @@ ZeroC                       https://zeroc.com
 Ice                         https://zeroc.com
 Jenkins                     http://jenkins-ci.org
 roadmap                     https://trac.openmicroscopy.org/ome/roadmap
-Open Microscopy Environment http://www.openmicroscopy.org/site
+Open Microscopy Environment http://www.openmicroscopy.org
```
jburel: https?

jburel: same for gs

omero/conf_autogen.py
```diff
@@ -1,4 +1,5 @@
 omero_db_version = "OMERO5.4DEV"
 omero_db_patch = "3"
+versions_bioformats = "5.6.0"
```
jburel, line 3: Could that value be extracted from ``omero.properties``?

jburel:
2 remaining ``:bf_doc:`` in ``contributing/data-model-schema.rst and contributing/index.rst`` that will point Do you want to keep them?

omero/conf_autogen.py
hflynn: I thought that was what Seb had done

jburel: done in ``autogen_db_version.py``

hflynn:
@jburel the remaining bf_docs links in the contributing docs are fine - they are set to latest in the conf.py for contributing as those docs are consistently documenting the latest state of play. I've changed the OMERO ones specifically so that the supported formats table is always showing the right info for the current version.

jburel:
Great! I was not sure if they were missed or not

omero/conf_autogen.py
hflynn: @sbesson I'll leave this one to you

sbesson: Isn't that done in c132c4f0cc2244df1019a7b87f104e45a0d447f6? See  the commit message of 741450ce83dd4bef349ecb5fc7981529a948ef5d

jburel: @hflynn sorry that's what I meant by done in ``autogen_db_version.py``
I first missed it
no action required

PR #1741, opened 2017-08-25
Remove anchors from Ice documentation links

sbesson:
These anchors are being marked as broken by the Sphinx linkchecker. Since these
are subsections of the page, this commit proposes to drop the anchor suffix
and link to the page instead which includes a top-level menu.

With this PR https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ should turn back to green

### DIFF
--- a/omero/sysadmins/grid.rst
+++ b/omero/sysadmins/grid.rst
@@ -308,7 +308,7 @@ permissions verifier, so that anyone with a proper OMERO account can
 access the server.
 
 See :zerocdoc:`Controlling Access to IceGrid Sessions
-<display/Ice/Resource+Allocation+using+IceGrid+Sessions#ResourceAllocationusingIceGridSessions-ControllingAccesstoIceGridSessions>`
+<display/Ice/Resource+Allocation+using+IceGrid+Sessions>`
 of the Ice manual for more information.
 
 Unique node names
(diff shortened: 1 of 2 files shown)


mtbc:
Do we have any idea why this is needed? Is it a Sphinx bug, perhaps related to that we pin to an old version?

sbesson:
Interesting, I cannot reproduce locally (the link is redirected as expected). I would be fine with marking this on hold to reduce the false positives while we investigate the Sphinx upgrade.

rleigh-codelibre:
Either way, the change looks fine to me with both links being valid.  But we may well want to update to a current Sphinx version--don't all our docs build with it now?

hflynn:
Sorry, I knew these were actually working which is why I hadn't removed them earlier. Sorting the Sphinx upgrade issues is on my list, I'd assumed that was the problem.

mtbc:
Ignoring OMERO-DEV-latest-docs failure while this PR is pending.

hflynn:
I think this should be unnecessary once we get the CI updated to latest sphinx - I'm not seeing this error locally with Sphinx 1.6

sbesson:
Excluding now that #1745 is in the build.

sbesson:
Since https://ci.openmicroscopy.org/job/OMERO-DEV-latest-docs/351/, these anchor warnings seem to have been fixed. Closing this. The Sphinx 1.6 upgrade is scheduled independently of this PR.

PR #1742, opened 2017-08-25
omero.web.public.get_only

dominikl:
Adds the new 'omero.web.public.get_only' property to the docs.
Are there other places where this should be mentioned?

See https://github.com/openmicroscopy/openmicroscopy/pull/5406 and https://trello.com/c/BdVULJRw/138-publicgetonly-docs .

### DIFF
--- a/omero/sysadmins/public.rst
+++ b/omero/sysadmins/public.rst
@@ -63,8 +63,8 @@ To set this up on your OMERO.web installation:
 - By default the public user is only allowed to perform GET requests. This means
   that the public user will not be able to Create, Edit or Delete data which
   requires POST requests.
-  You can change that by setting the :property:`omero.web.public.get_only`
-  property:
+  If you want to allow these actions from the public user, you can change the
+  :property:`omero.web.public.get_only` property:
   ::
 
       $ bin/omero config set omero.web.public.get_only false
--- a/omero/sysadmins/public.rst
+++ b/omero/sysadmins/public.rst
@@ -111,6 +111,12 @@ To set this up on your OMERO.web installation:
 
      $ bin/omero config set omero.web.public.server_id 1
 
+- By default the public user is only allowed to perform GET requests. You can 
+  change that by setting the :property:`omero.web.public.get_only` property:
+  ::
+
+     $ bin/omero config set omero.web.public.get_only false
+
 If you enable public access to the main webclient but still wish registered
 users to be able to login, the login page can always be accessed using a link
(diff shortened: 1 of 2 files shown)


joshmoore:
Not that it must be done in this PR, but before we let the card move to "Done" we need to explain to the users *why* they might want to change this setting as well as the effects on the url filter.

omero/sysadmins/public.rst
```diff
@@ -111,6 +111,12 @@ To set this up on your OMERO.web installation:
 
      $ bin/omero config set omero.web.public.server_id 1
 
+- By default the public user is only allowed to perform GET requests. You can
```
will-moore: Maybe add a little more info. 2nd sentence:
"This means that the public user will not be able to Create, Edit or Delete data which requires POST requests."

Also, since this config affects the ```url_filter``` setting, maybe this section should be moved up, above the ```url_filter``` section.

will-moore:
The ```url_filter``` paragraph starting "You can use the full webclient UI for public browsing of images." can  be much simplified now, since most/all of those filters in that example are POST and are now prevented. Also, I think we can remove the "was not designed for public use" etc since we use it for idr now.

E.g. 
"You can use the full webclient UI for public browsing of images. Attempts by public user to create, edit or delete data will fail silently with the default ```get_only``` setting above. You may also choose to disable various dialogs for these actions such as launching scripts or OME-TIFF export, for example: 

```
$ bin/omero config set omero.web.public.url_filter '^/(webadmin/myphoto/|webclient/(?!(script_ui|ome_tiff|figure_script))|webgateway/(?!(archived_files|download_as)))'
```

omero/sysadmins/public.rst
```diff
@@ -60,6 +60,15 @@ To set this up on your OMERO.web installation:
 
      $ bin/omero config set omero.web.public.password '<password>'
 
+- By default the public user is only allowed to perform GET requests. This means
```
dominikl, line 63: I moved this section above the url_filter section as @will-moore suggested. But I'm a bit worried, that it now actually looks like something the user is supposed to do, as part of setting up the public access.

will-moore:
I think this looks OK now.
If you want to make it clearer that this is optional, the second sentence could be:

"If you want to allow these actions from the public user, you can change the :property:`omero.web.public.get_only` property: "

dominikl:
Thanks @will-moore , that sounds better. Changed it accordingly.

PR #1714, opened 2017-07-06
tell sysadmins awkward CLI means to manage light admins

mtbc:
This new page omero/sysadmins/cli/LightAdmins.html explains how light administrator privileges can be manipulated from the CLI. It is brief and uninviting because we don't recommend this approach and soon we should be able to provide a better CLI means via `bin/omero user ...` or somesuch but if need be we can point people to this summary how-to in the meantime if its warnings don't scare them off. See https://trello.com/c/iZeXX5FD/6-manage-admin-privileges-from-cli for more background.

PR #1744, opened 2017-08-31
further link light admin pages

mtbc:
Beyond those from #1714 adds a couple more links among light admin pages. See https://trello.com/c/kDxd1MeO/57-cross-reference-light-admin-docs for background.

### DIFF
--- a/omero/sysadmins/admins-with-restricted-privileges.rst
+++ b/omero/sysadmins/admins-with-restricted-privileges.rst
@@ -17,11 +17,11 @@ facility managers, image analysts, or anybody who needs to organize
 users and data of others in OMERO.
 
 Full administrators in OMERO can create new administrators with
-restricted privileges using the OMERO.web
-interface, see the :help:`create new users <sharing-data#admin>` section
-of our Help documentation. With considerably more difficulty one may
-also :doc:`use OMERO.cli to adjust the restrictions <cli/light-admins>`
-on an administrator.
+restricted privileges using the OMERO.web interface, see the
+:help:`create new users <sharing-data#admin>` section of our Help
+documentation. With considerably more difficulty one may also :doc:`use
+OMERO.cli to adjust the restrictions <cli/light-admins>` on an
+administrator.
 
 We suggest here four setups that should cover the four mainstream
 workflows. Nevertheless, you can combine the privileges
--- a/omero/sysadmins/admins-with-restricted-privileges.rst
+++ b/omero/sysadmins/admins-with-restricted-privileges.rst
@@ -21,7 +21,7 @@ restricted privileges using the OMERO.web interface, see the
 :help:`create new users <sharing-data#admin>` section of our Help
 documentation. OMERO.cli does not yet support easy management of
 restrictions nor does it offer the helpful :doc:`permissions mapping
-<mapping-restricted-admins>` so we do not recommend :doc:`using
+<mapping-restricted-admins>` but advanced users may :doc:`use
 OMERO.cli to adjust the restrictions <cli/light-admins>` on an
 administrator.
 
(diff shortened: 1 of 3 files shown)


omero/sysadmins/admins-with-restricted-privileges.rst
```diff
@@ -17,9 +17,11 @@ facility managers, image analysts, or anybody who needs to organize
 users and data of others in OMERO.
 
 Full administrators in OMERO can create new administrators with
-restricted privileges using the OMERO.web
-interface, see the :help:`create new users <sharing-data#admin>` section
-of our Help documentation.
+restricted privileges using the OMERO.web interface, see the
+:help:`create new users <sharing-data#admin>` section of our Help
+documentation. With considerably more difficulty one may also :doc:`use
```
hflynn: Do we want to rephrase this to say it is possible but not recommended? Maybe:
"Support for creating restricted administrators is yet to be added to OMERO.cli. It is possible to [use OMERO.cli to adjust restrictions] on an existing administrator but we do not recommend this as the [permissions mapping] may cause confusion."

hflynn:
I realise it's not strictly within the scope of this PR but I also find the title of the permissions mapping page somewhat difficult to parse and also wonder whether that page would be better NOT listed in the main index and instead added to an invisible TOC at the bottom of the main light admins page.

mtbc:
Invisible TOC sounds scary but if my suggested retitling does not suffice then I'll happily accept a PR. :smiley:

hflynn:
Your retitling is indeed helpful, we can probably leave it there (someone else will probably complain if we hide the page ;))

omero/sysadmins/admins-with-restricted-privileges.rst
```diff
@@ -17,9 +17,13 @@ facility managers, image analysts, or anybody who needs to organize
...
 
 Full administrators in OMERO can create new administrators with
-restricted privileges using the OMERO.web
-interface, see the :help:`create new users <sharing-data#admin>` section
-of our Help documentation.
+restricted privileges using the OMERO.web interface, see the
+:help:`create new users <sharing-data#admin>` section of our Help
+documentation. OMERO.cli does not yet support easy management of
+restrictions nor does it offer the helpful :doc:`permissions mapping
+<mapping-restricted-admins>` so we do not recommend :doc:`using
```
joshmoore: Because there could be a mis-configuration or because it's more complicated? If the latter, perhaps this could be phrased as "advanced use".

omero/sysadmins/cli/light-admins.rst
```diff
@@ -16,11 +16,13 @@ administrator restrictions in the awkward manner described hereunder.
...
-  checking *one* box in the web interface may lift *multiple* underlying
-  restrictions from an administrator. The recommended OMERO.web
-  management interface may thus prove confusing if OMERO.cli has been
-  used to set a combination of restrictions that does not correspond to
-  those bundles of related restrictions available in OMERO.web.
+  the :doc:`permissions mapping <../mapping-restricted-admins>` is such
+  that checking *one* box in the web interface may lift *multiple*
+  underlying restrictions from an administrator. The recommended
+  OMERO.web management interface may thus prove confusing if OMERO.cli
+  has been used to set a combination of restrictions that does not
```
joshmoore, line 23: This certainly worries me. Is this something that's slated for 5.4.1? and/or should we consider a CLI plugin that we can rollout more quickly?

joshmoore (commented):
The specific changes read fine though I'm worried about the tone we're setting. It sounds somewhat undone, which is perhaps what you're trying to get across but it'd be good to be clear on the next steps.

omero/sysadmins/cli/light-admins.rst
mtbc: Noted on https://trello.com/c/iZeXX5FD/17-manage-admin-privileges-from-cli though does require someone to more than just tweak existing CLI code.

hflynn:
Having looked at what else we are doing on the index page, I'm thinking we should move the mapping page down a level in the TOC but I can do that on a new PR. Are you happy with this now @joshmoore ?

joshmoore:
Thanks, @mtbc

PR #1745, opened 2017-08-31
Make OMERO docs compatible with new sphinx

hflynn:
See https://trello.com/c/ccx2knE7/141-check-whether-sphinx-can-be-upgraded-for-all-docs-jobs

This updates Travis to use latest Sphinx and fixes all instances of `:option:` and `:omerocmd:` in the docs so the build passes. All changes should be backwards compatible so the merge docs currently still building with Sphinx 1.2.3 should render the same as before.

PR #1746, opened 2017-09-04
'omero/plugins' PYTHON path

dominikl:
With https://github.com/openmicroscopy/openmicroscopy/pull/4619 any 'omero/plugins' directory on the PYTHON path will
be searched for plugins. This PR just modifies the doc accordingly.

See:
https://trello.com/c/qMrZfcoU/78-external-cli-plugins
https://trello.com/c/DeEfMN0E/140-doc-cli-plugin-deployment

### DIFF
--- a/omero/developers/cli/extending.rst
+++ b/omero/developers/cli/extending.rst
@@ -12,8 +12,8 @@ will automatically include them.
 For testing purposes the ``--path`` argument can be used to point to other plugin
 files or directories, too.
 
-|CLI| plugins are also pip-installable. Search for ``omero-cli-*`` on `PyPI
-<https://pypi.python.org/pypi>`_.
+|CLI| plugins are also pip-installable. Search for ``omero-cli-*`` on
+:pypi:`PyPI <>`.
 
 Thread-safety
 ^^^^^^^^^^^^^
--- a/omero/developers/cli/extending.rst
+++ b/omero/developers/cli/extending.rst
@@ -6,7 +6,7 @@ directory. On execution, all plugins in that directory are registered
 with the |CLI|. 
 
 Alternatively, plugins can be added to any directory ending with
-``omero/plugins``. If this directory is part of the ``PYTHON`` path the |CLI|
+``omero/plugins``. If this directory is part of the ``PYTHONPATH`` the |CLI|
 will automatically include them.
 
 For testing purposes the ``--path`` argument can be used to point to other plugin
--- a/omero/developers/cli/extending.rst
+++ b/omero/developers/cli/extending.rst
@@ -3,8 +3,17 @@ Extensions
 
 Plugins can be written and put in the ``lib/python/omero/plugins``
 directory. On execution, all plugins in that directory are registered
-with the |CLI|. Alternatively, the :option:`--path` argument can be used to
-point to other plugin files or directories.
+with the |CLI|. 
+
+Alternatively, plugins can be added to any directory ending with
+``omero/plugins``. If this directory is part of the ``PYTHON`` path the |CLI|
+will automatically include them.
+
+For testing purposes the ``--path`` argument can be used to point to other plugin
+files or directories, too.
+
+|CLI| plugins are also pip installable. Search for ``omero-cli-*`` on `PyPI
+<https://pypi.python.org/pypi>`_.
 
 Thread-safety
 ^^^^^^^^^^^^^
--- a/omero/developers/cli/extending.rst
+++ b/omero/developers/cli/extending.rst
@@ -12,7 +12,7 @@ will automatically include them.
 For testing purposes the ``--path`` argument can be used to point to other plugin
 files or directories, too.
 
-|CLI| plugins are also pip installable. Search for ``omero-cli-*`` on `PyPI
+|CLI| plugins are also pip-installable. Search for ``omero-cli-*`` on `PyPI
 <https://pypi.python.org/pypi>`_.
 
 Thread-safety


omero/developers/cli/extending.rst
```diff
@@ -1,9 +1,11 @@
...
 ----------
 
-Plugins can be written and put in the ``lib/python/omero/plugins``
-directory. On execution, all plugins in that directory are registered
-with the |CLI|. Alternatively, the :option:`--path` argument can be used to
+Plugins can be written and put in any directory called ``.../omero/plugins``.
+Add this directory to your ``PYTHON`` path and on execution all plugins in that
+directory are registered with the |CLI| automatically. 
+
+Alternatively, the :option:`--path` argument can be used to
```
hflynn: Newer versions of Sphinx only allow the option markup to be used where the option is included in a definitions list. To save me having to fix this in future, can you just use double back ticks markup here please.

omero/developers/cli/extending.rst
```diff
@@ -1,10 +1,12 @@
 Extensions
 ----------
 
-Plugins can be written and put in the ``lib/python/omero/plugins``
-directory. On execution, all plugins in that directory are registered
-with the |CLI|. Alternatively, the :option:`--path` argument can be used to
-point to other plugin files or directories.
+Plugins can be written and put in any directory called ``.../omero/plugins``.
+Add this directory to your ``PYTHON`` path and on execution all plugins in that
```
mtbc: Bit ambiguous - is "this" directory the "omero" directory that contains "plugins" or is it the "plugins" directory we need in the path and our code checks that its parent is named "omero"?

mtbc:
This feature makes me wonder if we can get debug output from CLI about plugin registrations and if we can tell the user how to check that. Probably outwith this PR though.

joshmoore:
Few thoughts:
 * Let's keep the use of `lib/python` for the "official" or "shipped" plugins, but "It's on the PYTHONPATH by default"
 * Use of `--path` might be described as being for a "one-off" or testing purposes
 * I'd suggest we go ahead and mention pip installation
 * If so, perhaps a search like https://www.google.de/search?q=site%3Apypi.python.org+inurl%3Aomero-cli or https://github.com/search?utf8=%E2%9C%93&q=in%3Aname+omero-cli-+user%3Aome&type= could show potential installables

omero/developers/cli/extending.rst
```diff
@@ -3,8 +3,17 @@ Extensions
...
+with the |CLI|. 
+
+Alternatively, plugins can be added to any directory ending with
+``omero/plugins``. If this directory is part of the ``PYTHON`` path the |CLI|
+will automatically include them.
+
+For testing purposes the ``--path`` argument can be used to point to other plugin
+files or directories, too.
+
+|CLI| plugins are also pip installable. Search for ``omero-cli-*`` on `PyPI
```
mtbc: Hyphenate pip-installable.

mtbc:
Thank you, I find this formulation much clearer.

dominikl:
I have to admit, I find it rather confusing. Now there are apparently four different ways to get a CLI plugin.

joshmoore:
To explain (from the least correct to the most correct way):

 * `--path` was never really very useful. It's a pain to pass it on every invocation. If we want, we could make it disappear.
 * Adding things to `lib/python/omero/plugins` is prone to breakage since every new release you'd need to copy the file over.
 * `pip`-installation and adding to `PYTHONPATH` are essentially the same thing, or rather, you can use pip to get things onto your PYTHONPATH rather than copying them manually.

jburel:
We are more or less in the same situation we were with web apps i.e. copy apps "manually" vs pip
The recommended way will be "pip" but still some work before we have the plugins converted

dominikl:
Ok, thanks. Any more changes required then?

omero/developers/cli/extending.rst
```diff
@@ -3,8 +3,17 @@ Extensions
...
+with the |CLI|. 
+
+Alternatively, plugins can be added to any directory ending with
+``omero/plugins``. If this directory is part of the ``PYTHON`` path the |CLI|
+will automatically include them.
+
+For testing purposes the ``--path`` argument can be used to point to other plugin
+files or directories, too.
+
+|CLI| plugins are also pip-installable. Search for ``omero-cli-*`` on `PyPI
```
jburel: you should use 
```
:pypi:`PyPI <>`

```

jburel: Since we have mentioned of the plugins on PyPI
I have started looking at the plugin doc so it is similar to what we have for the web apps when pushed on PyPI

omero/developers/cli/extending.rst
```diff
@@ -3,8 +3,17 @@ Extensions
 
 Plugins can be written and put in the ``lib/python/omero/plugins``
 directory. On execution, all plugins in that directory are registered
-with the |CLI|. Alternatively, the :option:`--path` argument can be used to
-point to other plugin files or directories.
+with the |CLI|. 
+
+Alternatively, plugins can be added to any directory ending with
+``omero/plugins``. If this directory is part of the ``PYTHON`` path the |CLI|
```
will-moore: This should be ``PYTHONPATH`` (all one word) I think

hflynn:
Can we get this finished off asap so i can merge please? It's going to conflict with https://github.com/openmicroscopy/ome-documentation/pull/1745

dominikl:
Thanks @will-moore , fixed it.

hflynn:
Thanks! Is this good to merge now @jburel ?

jburel:
looks good merging

PR #1749, opened 2017-09-05
Unicode BlitzGateway

will-moore:
Add Unicode info to the BlitzGateway docs page.
See discussion at https://github.com/openmicroscopy/openmicroscopy/pull/5400

To test, check output from various BlitzGateway wrappers, e.g.

```
$ bin/omero shell --login

In [1]: from omero.gateway import BlitzGateway

In [2]: conn = BlitzGateway(client_obj=client)

In [3]: u = conn.getUser()

In [4]: u.getFullName()
Out[4]: u'Will Mo\xf6re'

In [5]: u._obj.getFirstName().val
Out[5]: 'Will'

In [7]: u.getLastName()
Out[7]: u'Mo\xf6re'

In [8]: u.getName()
Out[8]: u'will'

In [9]: i = conn.getObject("Image", 9551)

In [10]: i.getName()
Out[10]: '438_01_R3D_D3D\xc3\xb6.dv'

In [11]: i.getOwnerFullName()
Out[11]: u'Will Mo\xf6re'

In [27]: for c in i.getChannels():
    print type(c.getLut())
   ....:     
<type 'str'>
<type 'NoneType'>
```

### DIFF
--- a/omero/developers/PythonBlitzGateway.rst
+++ b/omero/developers/PythonBlitzGateway.rst
@@ -80,6 +80,40 @@ loading).
        survivin
     >>> conn.close()
 
+
+Unicode
+"""""""
+
+When unwrapping omero.model rstring attributes as shown above,
+the BlitzObjectWrapper subclasses will decode the utf8-encoded
+string and return the resulting unicode string.
+
+However, some methods added to BlitzObjectWrappers
+will return strings as follows::
+
+    BlitzObjectWrapper.getName()            except ExperimenterWrapper.getName() and LogicalChannelWrapper.getName()
+    BlitzObjectWrapper.getDescription()
+    AnnotationWrapper.getNs()
+    FileAnnotationWrapper.getFileName()
+    TagAnnotationWrapper.getValue()
+    CommentAnnotationWrapper.getValue()
+    MapAnnotationWrapper.getValue()  - list of strings
+    PlateWrapper.getColumnLabels() / getRowLabels()   - list of strings or integers
+    WellWrapper.getWellPos()
+    ColorHolder.getHtml() / getCss()
+    ChannelWrapper.getLut()
+    ImageWrapper.getChannelLabels()  - list of strings
+
+When using unicode strings, care must be taken when printing to
+the console if they include non-ASCII characters as
+`described here <https://pythonhosted.org/kitchen/unicode-frustrations.html>`_.
+
+When using the BlitzGateway in the OMERO.web framework,
+Django will convert strings into unicode strings so it's safe to
+use either. See Django
+`Unicode data <https://docs.djangoproject.com/en/1.8/ref/unicode/#general-string-handling>`_.
+
+
 Wrapper coverage
 """"""""""""""""
 
--- a/omero/developers/PythonBlitzGateway.rst
+++ b/omero/developers/PythonBlitzGateway.rst
@@ -84,12 +84,13 @@ loading).
 Unicode
 """""""
 
-When unwrapping omero.model rstring attributes as shown above,
-the BlitzObjectWrapper subclasses will decode the utf8-encoded
+When unwrapping ``omero.rtypes.rstring`` attributes from
+``omero.model.object`` instances as shown above,
+the ``BlitzObjectWrapper`` subclasses will decode the utf8-encoded
 string and return the resulting unicode string.
 
-However, some methods added to BlitzObjectWrappers
-will return strings as follows::
+However, some methods added to ``BlitzObjectWrappers``
+will return regular strings as follows::
 
     BlitzObjectWrapper.getName()            except ExperimenterWrapper.getName() and LogicalChannelWrapper.getName()
     BlitzObjectWrapper.getDescription()
@@ -109,7 +110,7 @@ the console if they include non-ASCII characters as
 `described here <https://pythonhosted.org/kitchen/unicode-frustrations.html>`_.
 
 When using the BlitzGateway in the OMERO.web framework,
-Django will convert strings into unicode strings so it's safe to
+Django will convert strings into unicode strings so it is safe to
 use either. See Django
 `Unicode data <https://docs.djangoproject.com/en/1.8/ref/unicode/#general-string-handling>`_.
 


omero/developers/PythonBlitzGateway.rst
```diff
@@ -80,6 +80,40 @@ loading).
        survivin
     >>> conn.close()
 
+
+Unicode
+"""""""
+
+When unwrapping omero.model rstring attributes as shown above,
```
jburel: ``omero.model`` objects
also better to have ``omero.rtypes.rstring``

omero/developers/PythonBlitzGateway.rst
```diff
@@ -80,6 +80,40 @@ loading).
...
+the BlitzObjectWrapper subclasses will decode the utf8-encoded
+string and return the resulting unicode string.
+
+However, some methods added to BlitzObjectWrappers
+will return strings as follows::
```
jburel: string(s)

omero/developers/PythonBlitzGateway.rst
```diff
@@ -80,6 +80,40 @@ loading).
...
+
```
jburel: Do you have a test or way to check if there is no other methods?
I think having a test with all the methods will be useful if we adjust something we will immediately know the impact

omero/developers/PythonBlitzGateway.rst
```diff
@@ -80,6 +80,40 @@ loading).
...
+    ColorHolder.getHtml() / getCss()
+    ChannelWrapper.getLut()
+    ImageWrapper.getChannelLabels()  - list of strings
+
+When using unicode strings, care must be taken when printing to
+the console if they include non-ASCII characters as
+`described here <https://pythonhosted.org/kitchen/unicode-frustrations.html>`_.
+
+When using the BlitzGateway in the OMERO.web framework,
+Django will convert strings into unicode strings so it's safe to
```
jburel: it is

omero/developers/PythonBlitzGateway.rst
will-moore: methods -> strings. Both are plural. This doesn't imply that a single method returns multiple strings.

omero/developers/PythonBlitzGateway.rst
will-moore: A test with ALL the methods of every object in the BlitzGateway? This will include all the methods of the underlying omero.model objects if we want to test whether any are returning strings. This is not really feasible.
Probably the best solution is to add unicode characters into all the various integration and robot tests so we know that these are being handled correctly.

jburel: The point I was making here is how can I confirm that the list here is valid?
adding check in the tests will be a great start

will-moore:
Fixed a couple of issues.

omero/developers/PythonBlitzGateway.rst
will-moore: You can confirm that the listed methods return strings by using the python shell as shown above. I can add more examples to cover the whole list if needed.
However, confirming that no other methods return strings is tricky. The only way is to check the source code, which is what I did to get that list.
We could go through all our BlitzGateway tests and ```assert type(s) == unicode``` but that would be a lot of work and I don't think our existing tests have complete coverage of all omero.model.object methods.

jburel: I understood that this list was not an example but a complete list of files.
At least the ones we have tests for

will-moore: The ```$ bin/omero shell --login``` in the PR description is the example. The list in the PR is a complete list that I compiled from reading the BlitzGateway source code. I haven't checked the tests coverage for these methods.

omero/developers/PythonBlitzGateway.rst
```diff
@@ -80,6 +80,41 @@ loading).
...
+
+When unwrapping ``omero.rtypes.rstring`` attributes from
+``omero.model.object`` instances as shown above,
+the ``BlitzObjectWrapper`` subclasses will decode the utf8-encoded
+string and return the resulting unicode string.
+
+However, some methods added to ``BlitzObjectWrappers``
+will return regular strings as follows::
+
+    BlitzObjectWrapper.getName()            except ExperimenterWrapper.getName() and LogicalChannelWrapper.getName()
```
jburel, line 95: why such diff?

will-moore: LogicalChannel.getName() returns unicode: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/gateway/__init__.py#L7294
Same for Experimenter: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/src/omero/gateway/__init__.py#L5777

Here I am just documenting the current behaviour, not trying to explain why it's this way.

jburel: I understand but why such a diff in the first place (if you remember)
In case we need to deprecate methods

will-moore: I don't think there's any intention to have some methods return string and some return unicode, but this is how it's been since the BlitzGateway was created:
```
commit f0311e5fd641f7bb5c7ff47995fdb2dc8e06163d (HEAD)
Author: carlos <carlos@05709c45-44f0-0310-885b-81a1db45b4a6>
Date:   Mon Apr 6 14:08:44 2009 +0000
    Adding omero.gateway, a python wrapper for omero.client.
```
This returns String
```
def getName (self):
        return self._obj.getName().val
```
This returns unicode
```
    def __getattr__ (self, attr):
        if hasattr(self._obj, attr):
            rv = getattr(self._obj, attr)
            if hasattr(rv, 'val'):
                return isinstance(rv.val, StringType) and rv.val.decode('utf8') or rv.val
```
And in all the time I've used the BlitzGateway, I've never really been aware that some return unicode.

hflynn:
Is this sorted now?

omero/developers/PythonBlitzGateway.rst
jburel: we should record and start an audit  so we can deprecate method and unify that

jburel:
@hflynn on the doc front I think we are okay for now
unless @will-moore is planning to add more examples

omero/developers/PythonBlitzGateway.rst
will-moore: I'm not sure we can deprecate ```getName()```, since we can't remove it, we can only change the return type to Unicode. We could maybe just add a warning that the return type will change to Unicode at some point (5.5.0)? We should probably also make every BlitzGateway method return Unicode, or all return String. String is easier since that's what we're starting with everywhere.

will-moore:
Not planning to add anything more to this.

omero/developers/PythonBlitzGateway.rst
joshmoore: If we're going to do that review and unification _anyway_, then I'd go back to the suggestion of making it user-configurable.

hflynn: I take it this discussion does not affect me merging this PR @joshmoore ?

joshmoore: Correct.

PR #1768, opened 2017-09-19
database upgrade for OMERO <5.3.4

mtbc:
Look for warning boxes at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ws/src/omero/_build/html/sysadmins/server-upgrade.html.

I figured that this approach was least invasive to the default workflow which favors those who are already running a secure public release. Note that some version numbers may look odd because `conf_autogen.py` is running behind and the "dbver" helpers it defines appear not to work within `:file:` markup.

### DIFF
--- a/omero/sysadmins/server-upgrade.rst
+++ b/omero/sysadmins/server-upgrade.rst
@@ -121,9 +121,10 @@ precautionary checks also done by the actual upgrade script.
    The :file:`sql/psql/OMERO5.4__0/OMERO5.3__1-precheck.sql` script
    referenced by the above :program:`psql` command assumes a planned
    upgrade from OMERO 5.3.4. If you are instead currently running OMERO
-   5.3.3 or an earlier 5.3.x version then you perform the precheck by using the above
-   command with :file:`sql/psql/OMERO5.4__0/OMERO5.3__0-precheck.sql`.
-   That script verifies that the database contains no trace of
+   5.3.3 or an earlier 5.3.x version then you perform the precheck by
+   using the above command with
+   :file:`sql/psql/OMERO5.4__0/OMERO5.3__0-precheck.sql`. That script
+   verifies that the database contains no trace of
    :secvuln:`2017-SV5-filename-2` having been exploited; this
    vulnerability was fixed in OMERO 5.3.4.
 
@@ -250,9 +251,10 @@ intermediate major release.
 
    The :file:`sql/psql/OMERO5.4__0/OMERO5.3__1.sql` script referenced by
    the above :program:`psql` command assumes upgrade from OMERO 5.3.4.
-   If you are instead currently running OMERO 5.3.3 or an earlier 5.3.x version then you
-   upgrade the database directly to OMERO 5.4.0 by using the above
-   command with :file:`sql/psql/OMERO5.4__0/OMERO5.3__0.sql`.
+   If you are instead currently running OMERO 5.3.3 or an earlier 5.3.x
+   version then you upgrade the database directly to OMERO 5.4.0 by
+   using the above command with
+   :file:`sql/psql/OMERO5.4__0/OMERO5.3__0.sql`.
 
 Delete certain annotations (optional)
 """""""""""""""""""""""""""""""""""""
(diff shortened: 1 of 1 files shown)


sbesson:
Slightly off-topic, our usage of `current_dbver`, `previous_dbver` and `previousversion`seems to be largely limited to the `sysadmins/server-upgrade.rst` page. Assuming we are carefully reviewing this page prior to every release that requires a DB upgrade, this might raise the question of whether these substitutions are effectively bringing more noise than value.

omero/sysadmins/server-upgrade.rst
```diff
@@ -116,6 +116,14 @@ precautionary checks also done by the actual upgrade script.
...
 
 
+.. warning::
+
+   The :file:`sql/psql/OMERO5.4__0/OMERO5.3__1-precheck.sql` script
+   referenced by the above :program:`psql` command assumes a planned
+   upgrade from OMERO 5.3.4. If you are instead currently running OMERO
+   5.3.0 to 5.3.3 then you perform the precheck by using the above
+   command with :file:`sql/psql/OMERO5.4__0/OMERO5.3__0-precheck.sql`.
+
```
jburel: Maybe add a note that the script checks if the DB is "corrupted"
Actually we do not offer a way to fix that issue

mtbc:
@sbesson: Could we somehow define those at the head of the page or something for reuse within the page?

sbesson:
Definining substitutions within a given page is always possible. If I understand correctly https://github.com/openmicroscopy/ome-documentation/pull/1768#issue-258855749, this might not solve the problem of encapsulating these within different types of markup (`file`, `literalinclude`). Which ones do you have in mind?

mtbc:
I was thinking of the ones we have routinely (not these new warning boxes) in the `parsed-literal` blocks.

omero/sysadmins/server-upgrade.rst
```diff
@@ -116,6 +116,17 @@ precautionary checks also done by the actual upgrade script.
...
+.. warning::
+
+   The :file:`sql/psql/OMERO5.4__0/OMERO5.3__1-precheck.sql` script
+   referenced by the above :program:`psql` command assumes a planned
+   upgrade from OMERO 5.3.4. If you are instead currently running OMERO
+   5.3.0 to 5.3.3 then you perform the precheck by using the above
+   command with :file:`sql/psql/OMERO5.4__0/OMERO5.3__0-precheck.sql`.
+   That script verifies that the database contains no trace of
+   :secvuln:`2017-SV5-filename-2` having been exploited; this
+   vulnerability was fixed in OMERO 5.3.4.
```
jburel: Do we suggest to contact us?
or we keep that for announcement

mtbc: I think they'll contact us anyway if they run into this, otherwise they can't upgrade. Nobody has reported an upgrade issue for 5.3.4 and nightshade's challenging database was also fine so probably not a wide concern.

jburel: true
so in that case, let's keep it as it is

omero/sysadmins/server-upgrade.rst
```diff
@@ -235,6 +246,14 @@ intermediate major release.
    connected to the database using **db_user** before running the script. See
    :forum:`this forum thread <viewtopic.php?f=5&t=7778>` for more information.
 
+.. warning::
+
+   The :file:`sql/psql/OMERO5.4__0/OMERO5.3__1.sql` script referenced by
+   the above :program:`psql` command assumes upgrade from OMERO 5.3.4.
+   If you are instead currently running OMERO 5.3.0 to 5.3.3 then you
```
hflynn: This "5.3.0 to 5.3.3" confused me because on first glance I parsed it as running the upgrade from 5.3.0 to 5.3.3 - could you rephrase to "running OMERO 5.3.3 or an earlier 5.3.x version" please?

hflynn:
Thanks

hflynn:
Is this good to merge now?

pwalczysko:
Reads well and builds well, the red checkboxes make sense. The comments here seem to be addressed.

Ready to merge FMPOV.

PR #1769, opened 2017-09-19
Secvul

jburel:
In this PR
* Update the link so we go directly to the new website
* Update term to match web site

 Ref in history updated in https://github.com/openmicroscopy/ome-documentation/pull/1769

### DIFF
--- a/common/conf.py
+++ b/common/conf.py
@@ -155,6 +155,7 @@ extlinks = {
     'community' : (oo_root + '/support/%s', ''),
     'omero' : (oo_root + '/omero/%s', ''),
     'bf' : (oo_root + '/bio-formats/%s', ''),
+    'secvuln' : (oo_root + '/security/advisories/%s', ''),
     # Doc links
     'model_doc' : (docs_root + '/latest/ome-model/%s', ''),
     'devs_doc' : (docs_root + '/contributing/%s', ''),
@@ -169,7 +170,6 @@ extlinks = {
     'djangodoc' : ('https://docs.djangoproject.com/en/1.8/%s', ''),
     'doi' : ('http://dx.doi.org/%s', ''),
     'pypi': ('https://pypi.python.org/pypi/%s', ''),
-    'secvuln' : (oo_root + '/security/advisories/%s', '')
     }
 
 rst_epilog = """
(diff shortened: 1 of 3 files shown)


common/conf.py
```diff
@@ -169,8 +169,7 @@ def get_previous_version(majornumber, minornumber=0):
     'djangodoc' : ('https://docs.djangoproject.com/en/1.8/%s', ''),
     'doi' : ('http://dx.doi.org/%s', ''),
     'pypi': ('https://pypi.python.org/pypi/%s', ''),
-    # Remaining Plone links
-    'secvuln' : (oo_root + '/info/vulnerabilities/%s', '')
+    'secvuln' : (oo_root + '/security/advisories/%s', '')
```
sbesson: Worth moving up alonsgide with the rest of the website extlink?

omero/sysadmins/server-security.rst
```diff
@@ -21,7 +21,7 @@ client including:
 .. note:: The OMERO team treats the security of all components with care and
     attention. If you have a security issue to report, please do not hesitate
     to contact us using our private, secure mailing list as described on the
-    :secvuln:`security vulnerabilities <>` page.
+    :secvuln:`Security Advisories <>` page.
```
sbesson: This should actually link to the top-level  http://www.openmicroscopy.org/security/ page.

sbesson (commented):
The link in the `server-security` page should be reviewed. Additionally, as we are moving away from the security vulnerabilities semantics in favor of the security advisories, I wonder if we should consider deprecating/aliasing the `secvuln` extlink in favor of a more general `security` one.

omero/sysadmins/server-security.rst
jburel: In that case we will need a new link, we cannot use ``:secvuln:``

hflynn:
Happy with this now @sbesson ?
We can always move towards using the new `:sec:` with `advisories/xxxx` in future and deprecate the secvuln ext link after this release cycle.

sbesson:
I think the latest state of this PR is fine and indeed the new security alias allows us to deprecate `secvuln` in the future if we want to. Only minor interrogation was whether we should name it `security` rather than `sec` but I don't think this should block this PR unless someone else has a strong opinion.

jburel:
I will rename it now

PR #1783, opened 2017-09-27
Managing OMERO docs re-organization

hflynn:
Partially based on the suggestions on https://docs.google.com/spreadsheets/d/1Swi9irN-GLqsAoYkRmpoGvoSoLGVDxfLh8UpwMcz1Mg/edit#gid=0 this highlights a managing OMERO subsection of the sysadmin docs aimed at restricted admins and better cross-references the CLI docs for this user group. 

This is a conservative approach with only updates one URL which is currently only 'live' in milestone releases anyway.

PR #1784, opened 2017-09-29
Refer to 5.3 docs for windows migration

hflynn:
As noted in https://github.com/openmicroscopy/ome-documentation/pull/1783#issuecomment-333086345 I started updating this page to point to the 5.3 docs and realized people would likely get hopelessly lost swapping between versions so it would be easier just to refer to all the content on 5.3 instead. I haven't deleted the page though so no redirect needed.


hflynn:
https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-merge-docs/860/ updated to include this PR

omero/sysadmins/windows-migration.rst
```diff
@@ -1,65 +1,4 @@
 Migration from OMERO 5.2 on Windows to OMERO 5.3 on Linux
 =========================================================
```
mtbc: Can we just get rid of this page? Or is this kind of like an API deprecation?

hflynn: I defer to @joshmoore here. Although since the page has only existed since 5.3 maybe that adds weight to the argument we could delete it altogether.

mtbc: 5.4.0 allows breaking change. :smiley:

hflynn: I guess we could also point people at the page when we announce that they must upgrade from 5.2

joshmoore: Is there an statement somewhere to say, "if you are updating from a version not listed on this page, then please see `|LASTVERSION|` documentation" ? If so, I'd assume that suffices.

hflynn: There is both that AND I've added a specific link to this page underneath that statement. In that case, I'll go ahead and delete it.

PR #1786, opened 2017-10-04
have deployed web URLs always allow login

mtbc:
Staged at https://ci.openmicroscopy.org/job/CONTRIBUTING-merge-docs/ws/src/contributing/_build/html/ci-omero.html. Adopts @manics' idea of using `webclient/login/` in URL to avoid forced login as public user.


kennethgillen:
👍  looks good to merge.

PR #1790, opened 2017-10-10
Removing text Linux version referring to Dist.

kennethgillen:
Not sure what others think, but "linux version" doesn't seem right to me. 

I don't want to risk confusing our users though - perhaps "Distribution version" is clear enough to everyone?

### DIFF
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -44,7 +44,7 @@ Operating systems
   .. list-table::
      :header-rows: 1
 
-     * - Dist. version
+     * - Distribution version
        - OMERO 5.3
        - OMERO 5.4
      * - 6.x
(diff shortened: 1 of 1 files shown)


omero/sysadmins/version-requirements.rst
```diff
@@ -39,12 +39,12 @@ Operating systems
     deployment, although the server can be expected to build on versions with
     current security support for development purposes.
 
-* Linux (CentOS/RHEL)
+* Linux distributions CentOS/RHEL
```
mtbc: Maybe singular here following pattern with Ubuntu below.

mtbc:
You're right, Linux version typically will *not* match distribution version, good catch.

omero/sysadmins/version-requirements.rst
kennethgillen: https://github.com/openmicroscopy/ome-documentation/pull/1790#discussion_r143927475 addresses plurality.

mtbc (approved):

PR #1799, opened 2017-10-26
Simplify and update EditingOmeroWeb.rst

will-moore:
PR to update web development docs as part of Training.

More edits to come depending on the outcomes of https://trello.com/c/VdQ4HAl6/65-using-nginx-in-development

TODO:
 - [ ] Add my build script to docs https://gist.github.com/will-moore/bbcb487b99502e9275c4fc6d80fb3b9c after cleaning up (use params instead of hard-coded paths) as suggested in https://github.com/ome/presentations/pull/70 discussion.

### DIFF
--- a/omero/developers/Web/EditingOmeroWeb.rst
+++ b/omero/developers/Web/EditingOmeroWeb.rst
@@ -2,53 +2,25 @@ Editing OMERO.web
 =================
 
 If you need to make changes to OMERO.web itself, then you
-will want to check out the OMERO source code. You can directly
-edit and run the OMERO.web code.
-This means that you benefit from the convenience of editing,
-saving and refreshing your browser without any build step.
-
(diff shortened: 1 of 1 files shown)


hflynn:
Are you expecting to finish this before the xmas break @will-moore ?

will-moore:
This could be considered done for now.
The only question is whether to move my full build script https://gist.github.com/will-moore/bbcb487b99502e9275c4fc6d80fb3b9c into this page, but it needs cleaning-up to remove hard-coded paths etc.

rgozim (approved):

hflynn:
Are you planning to add that script on this PR @will-moore or shall I merge this and leave you to card that extra action?

jburel:
we cannot use that script in our doc.

hflynn:
Merging in that case, thanks for the clarification J-M

PR #1800, opened 2017-10-27
separate words, unify acronym style

kennethgillen:
Some formatting fixes.


hflynn:
Funnily enough, this page is covered in the docs training @kennethgillen - the content is generated from server properties and edits made directly in the docs page will just be overridden next time the autogen job is run. 

You'll need to open these changes against the code repo instead - see for example https://github.com/openmicroscopy/openmicroscopy/pull/4656

PR #1812, opened 2017-11-20
use correct user

jburel:
Clarify the user who needs to run the command

https://trello.com/c/8Tg5qkYn/76-remove-pip-install-as-omero-system-user

The card was not marked as critical but maybe something to consider for this round.

cc @hflynn


hflynn:
Staging looks good and this matches my understanding of the card, do we need someone else to check this or shall I just merge?

jburel:
If you are happy with the changes it should be good to go

PR #1818, opened 2017-12-01
Remove linkcheck parser from docs pages

hflynn:
See https://trello.com/c/a1E2H77k/608-update-contributing-docs-to-reflect-latest-docs-build-changes - when Sphinx was upgraded, the jobs were reconfigured and no longer use the linkcheck parser so they now fail (red) when there are broken links and you need to look in the console output to find the issue.


contributing/ci-docs.rst
```diff
@@ -363,38 +363,9 @@ All are currently built from the master branches despite the build names.
...
-Priority FAILED UNSTABLE
-======== ====== ========
-High	 0
-Normal			0
-Low				10
-======== ====== ========
+Linkchecking
+^^^^^^^^^^^^
+
+Use of the linkcheck output parser was discontinued when the version of Sphinx
```
sbesson: Reading through it for the first time, the immediate question would be "which linkckeck output parser". 
I am unclear whether this section is required at all or if we should drop if and have this policy change recorded via the Git/GitHub history.

sbesson (commented):
One comment about the linkchecking section. Otherwise good to merge and deploy.

contributing/ci-docs.rst
hflynn: I initially just dropped it altogether then wondered if I should leave something. I'll delete it again.

hflynn:
Fixed @sbesson comment, merging and deploying

PR #1821, opened 2017-12-08
Fix version replacement

hflynn:
See https://trello.com/c/OUY9xIlY/613-missing-version-number-from-541-sysadmin-docs
|version| wasn't working on http://docs.openmicroscopy.org/omero/5.4.1/sysadmins/server-upgrade.html (even though |previousversion| is) because this file was overriding conf.py

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -137,6 +137,9 @@ To modify this value set the environment variable ``OMERO_RELEASE`` e.g.::
     cd omero && OMERO_RELEASE=5.2.6 make clean html
     cd omero && OMERO_RELEASE=5.2.6 ant clean html
 
+This variable needs to be set to build all the version-dependent content
+used in the system administrator documentation correctly.
+
 The Contributing Developer documentation has the release version removed as
 the intention is to update these files as and when necessary, so that they
 always reflect our current practices.
@@ -152,14 +155,14 @@ By default, running ``ant`` will build as a zipped bundle.
 
 From the top level directory::
 
-    ant zip -Domero.release="5.3.0"
+    OMERO_RELEASE=5.3.0 ant zip -Domero.release="5.3.0"
 
 will generate the HTML documentation for OMERO and Contributing and bundle
 just the OMERO documentation into an OMERO.doc-5.3.0.zip under omero/_build.
 
 From omero directory::
 
-    ant zip -Domero.release="5.3.0"
+    OMERO_RELEASE=5.3.0 ant zip -Domero.release="5.3.0"
 
 will generate the HTML documentation for OMERO and create an
 OMERO.doc-5.3.0.zip under omero/_build.
--- a/README.rst
+++ b/README.rst
@@ -134,8 +134,8 @@ Release number
 The release number of the OMERO documentation is `UNKNOWN` by default.
 To modify this value set the environment variable ``OMERO_RELEASE`` e.g.::
 
-    cd omero && OMERO_RELEASE=5.2.6 make clean html
-    cd omero && OMERO_RELEASE=5.2.6 ant clean html
+    cd omero && OMERO_RELEASE=5.4.0 make clean html
+    cd omero && OMERO_RELEASE=5.4.0 ant clean html
 
 This variable needs to be set to build all the version-dependent content
 used in the system administrator documentation correctly.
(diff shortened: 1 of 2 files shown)


hflynn:
https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-merge-docs/936/ features this PR and the upgrade page displays as it should now.

hflynn:
upgrade_screenshot

hflynn:
Probably needs someone to check there are no wider implications

jburel:
build locally with the same command as the one used in the job
and also by modifying the conf.py
I do not have any banner under ``Upgrade your database``
Am I missing something?

hflynn:
It's only visible for point releases @jburel so you have to set the release version to x.x.y where y >0

jburel:
I think I had set at 5.4.1 so it should be visible
Let me double-check

hflynn:
I'm a bit mystified because it works fine for me locally using `OMERO_RELEASE=5.4.1 make clean html`

hflynn:
Ah, putting the release version in conf.py instead doesn't seem to work, is that what you tested?

hflynn:
I assume this file is still overriding the release version set in the conf.py

jburel:
I have narrowed down the problem.
Running `` OMERO_RELEASE=5.4.1 make clean html`` will work as you mentioned but running
``ant clean html -Domero.release="5.4.1"``  (similar command that the one in the daily build) does not

hflynn:
Can I hand this over to you or @sbesson then please @jburel ? I just copied what Seb told me, I've no idea how to debug this any further.

sbesson:
Reviewed the state of the build by testing various build commands with and without `OMERO_RELEASE` set. Tables below give the values of the `(release, version, previousversion)` variables as read in `conf.py` and the ones effectively used in the generated HTML code.

For the current state of `develop`  with Sphinx 1.6.3:

|   | `conf.py` | HTML
|--|----------|--------
| OMERO_RELEASE=5.4.2 make html | ('5.4.2', '5.4', '5.3') | ('5.4.2', 'UNKNOWN', UNKNOWN')
| make html | ('UNKNOWN', 'UNKNOWN', UNKNOWN') | ('UNKNOWN', 'UNKNOWN', UNKNOWN')
|  OMERO_RELEASE=5.4.2 ant clean html -Domero.release="5.4.2"  | ('5.4.2', '5.4', '5.3')  | ('5.4.2', '5.4', '5.3')
|  ant clean html -Domero.release="5.4.2"  |  ('UNKNOWN', 'UNKNOWN', UNKNOWN')  | ('5.4.2', 'UNKNOWN', UNKNOWN')

Testing the same with `v5.2.8` and Sphinx 1.2.3:

|   | `conf.py` | HTML
|--|----------|--------
| OMERO_RELEASE=5.2.8 make html | ('5.2.8', '5.2', '5.1') | ('5.2.8', '5.2', '5.1')
| make html | ('UNKNOWN', 'UNKNOWN', 'UNKNOWN') | ('UNKNOWN', 'UNKNOWN', 'UNKNOWN')
|  OMERO_RELEASE=5.2.8 ant clean html -Domero.release="5.2.8"  |  ('5.2.8', '5.2', '5.1')  | ('5.2.8', '5.2', '5.1')
|  ant clean html -Domero.release="5.2.8"  |  ('5.2.8', '5.2', '5.1')  | ('5.2.8', '5.2', '5.1')

In summary, I think the changes in https://github.com/openmicroscopy/ome-documentation/pull/1601 introduced the following behavior since 5.3.0:
- setting the `OMERO_RELEASE` envvar is mandatory to populate the internal variables (`version`, `previous_version`)
- specifying `-Domero.release` in the Ant command argument will only override the `release` variable specified in `conf.py`

hflynn:
Having chatted to @sbesson - a move to using the conf.py to set the version is probably something for the future, requiring proper scoping. 

For 5.4.2 we just need to decide if this fix is sufficient (all the CI jobs set the OMERO_RELEASE envar so this isn't going to cause problems for CI builds) or if we want to try and get `-Domero.release` working again. 

I just tested the zip building using `OMERO_RELEASE=5.2.6 ant zip -Domero.release="5.2.6"` and it gives the version number in the zip bundle filename and the version replacements within the docs work fine (without the -Domero.release set, the zip filename is missing the version number) so we could just change the README. What do you think @jburel ?

jburel:
I will test it locally

hflynn:
Shall I list this on standup for you to review tomorrow @jburel ?

jburel:
yes please

jburel:
Tested locally
I think the fix and adjusting the README will be sufficient for 5.4.2
Running the command 
``OMERO_RELEASE=5.4.2 ant zip -Domero.release="5.4.2"``
works as expected. The warning is displayed and the zip correctly created with the version number

We can always come back to it later on

hflynn:
Great, I'll push a commit to update the README shortly then

hflynn:
@jburel README update pushed, good to merge now?

jburel:
minor comment: maybe use more recent OMERO version in README e.g. 5.3 and 5.4.0 instead of 5.2, 5.3

hflynn:
@jburel pushed

jburel:
Thanks 
ready to merge

PR #1822, opened 2017-12-13
Put regen NGINX config in red warning box

hflynn:
See https://trello.com/c/d1SxLNjo/62-omeroweb-upgrade-highlight-regenerate-nginx-config

I also moved the Django section because it seemed like it should belong in the prerequisites section instead (plus multiple note boxes do not make for easy reading)

### DIFF
--- a/omero/sysadmins/omeroweb-upgrade.rst
+++ b/omero/sysadmins/omeroweb-upgrade.rst
@@ -26,13 +26,7 @@ Before starting the upgrade, please ensure that you have reviewed and
 satisfied all the :doc:`system requirements <system-requirements>` with
 :doc:`correct versions <version-requirements>` for installation.
 
-Since OMERO 5.2, the OMERO.web framework no longer bundles a copy of the
-Django package, instead manual installation of the Django dependency is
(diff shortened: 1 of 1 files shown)


manics:
FYI the installation of django should be handled by the dependencies section of this page. Do you think it be worth changing `should` to `must`, in which case you could even drop that paragraph about django? https://github.com/hflynn/ome-documentation/blob/3ba3e772d96301225b62753250e354a30e7e0a2b/omero/sysadmins/omeroweb-upgrade.rst#dependencies

manics:
screen shot 2017-12-14 at 10 34 06

Looks fine. In the Django note do you think it's worth saying the `pip install ...` command will take care of installing Django? Or do you reckon it's clear enough?

hflynn:
@manics added a quick note just incase it isn't clear enough.

manics:
Assuming the next merge build is fine good to merge.

hflynn:
https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-merge-docs/947/ is green and doc page looks as expected, so merging, thanks

PR #1826, opened 2018-01-11
add links from group/user management to relevant data pages

mtbc:
I could imagine an admin clicking onto this page, which is more about managing users and groups, when wanting to adjust the user and/or group of model objects. Staged at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ws/src/omero/_build/html/sysadmins/cli/usergroup.html.


hflynn:
Renders as expected on the staging pages, thanks Mark

PR #1827, opened 2018-01-17
Add note about systemd umask for in-place import

manics:
See https://github.com/ome/omero-install/pull/173#issuecomment-358233940:
> I don't think we specify anywhere assumptions about the omero(-server) user's group, therefore assuming that 0002 is safe on all servers is probably an issue. Without digging deeper, I'd err on the side of making that an addition to the in-place documentation, i.e. a sysadmin is knowingly opening up permissions on a system because of a level of trust granted to the in-place group.

Background: http://lists.openmicroscopy.org.uk/pipermail/ome-users/2018-January/006865.html

### DIFF
--- a/omero/sysadmins/in-place-import.rst
+++ b/omero/sysadmins/in-place-import.rst
@@ -181,6 +181,9 @@ was done for one of our test servers in Dundee:
     [omero_system_user@ome-server ManagedRepository]$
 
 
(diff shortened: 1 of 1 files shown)


rgozim:
Formatting is and linking is fine. Can't really say more than that.

joshmoore:
This seems a bit lost at the bottom of a very large shell dump. Until we break up the latter, there might not be a good place for the former.

joshmoore:
Merging and deferring my visibility comment above to https://github.com/openmicroscopy/ansible-role-omero-server/pull/27#issuecomment-358988180

PR #1833, opened 2018-01-19
Use OMERO.CLI as the search term for OMERO CLI plugins

sbesson:
Since all our plugins deployed on PyPI are standardized on using `OMERO.CLI` as a keyword, this URL should give a more targetted list of results.


hflynn:
wfm thanks

PR #1830, opened 2018-01-19
Adjustments for 5.4.2

joshmoore:
Various last-minute adjustments to the docs for the 5.4.2 release.

cc: @hflynn @pwalczysko @kennethgillen @jburel

PR #1835, opened 2018-01-22
Add link to example NGINX redirects

hflynn:
Adds the example @kennethgillen suggested on #1830

### DIFF
--- a/omero/sysadmins/public.rst
+++ b/omero/sysadmins/public.rst
@@ -266,4 +266,5 @@ This makes use of redirects allowing
 link to the correct group and Project in OMERO, just as the
 DOI above does. Redirects need to be set in the
 `NGINX <http://nginx.org/>`_ component of the OMERO.web installation
-dedicated to publication workflows.
+dedicated to publication workflows. You can find our configuration for this
+example `here on GitHub <https://github.com/openmicroscopy/prod-playbooks/blob/master/nightshade-webclients.yml#L172>`_.
--- a/omero/sysadmins/public.rst
+++ b/omero/sysadmins/public.rst
@@ -267,4 +267,4 @@ link to the correct group and Project in OMERO, just as the
 DOI above does. Redirects need to be set in the
 `NGINX <http://nginx.org/>`_ component of the OMERO.web installation
 dedicated to publication workflows. You can find our configuration for this
-example `here on GitHub <https://github.com/openmicroscopy/prod-playbooks/blob/master/nightshade-webclients.yml#L172>`_.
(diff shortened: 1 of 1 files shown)


kennethgillen:
@hflynn - nice one, thank you.

hflynn:
Rebuilding merge docs now

hflynn:
Staged at https://ci.openmicroscopy.org/view/Docs/job/OMERO-DEV-merge-docs/991/ and build is green

mtbc:
Staged at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ws/src/omero/_build/html/sysadmins/public.html#configuring-urls and looks good to me.

There's a question of if/how to link to a specific version of the playbook: there I have no strong opinion against any of the options.

mtbc (approved):

hflynn:
Re-run the builds just to recheck but all fine so merging with Mark's thumbs up

joshmoore:
I'm worried about this strategy. For the moment, this makes complete sense and clearly identifies what we're talking about. But as we evolve this file (which we will be doing), what warns/reminds us that this line needs to be updated?

kennethgillen:
If we've linked to at least one method of achieving this, does it _need_ to stay current? 

e.g. if we changed the wording to say " You can find an example of configuration we've used for this
 +example" ?

joshmoore:
"an example" would be better. If there's a bug though, or we change strategy, we may not want people copying-n-pasting from this file.

hflynn:
@joshmoore that's why I updated the link to point at the line in a commit so that the content will stay static. The example illustrates what we're talking about, so it doesn't need to be updated.

hflynn:
If we change strategy, we'll likely to be updating the doc page, in which case we'll have the link here to know to update it

joshmoore:
But we won't know when changing prod-playbooks to look here for whether or not a change is needed. I did a search for `blob` throughout our code base and for the bigger repos, couldn't find any use of `blob/$commit` (as opposed to `blob/$branch` or `blob/$tag`). Discussing with @kennethgillen, I propose we add a tag and specify a bit better what it is we're linking to. I'll open momentarily.

kennethgillen:
NB Josh suggested we try linking to >1 lines: tried this out and the URL API supports this, e.g. 
https://github.com/openmicroscopy/prod-playbooks/blob/2c73f0482bb87dc67e8d011d087ae591ad0bd836/nightshade-webclients.yml#L181-L183

hflynn:
Happy to amend to a tag and wider line range, just let me know the new link

PR #1841, opened 2018-02-07
Update OMERO sidebar for proposed navigation

hflynn:
This implements the change proposed on https://github.com/openmicroscopy/www.openmicroscopy.org/pull/195 

Doing this now so I don't forget about it but also applying an exclude label for now as it'll break the linkchecker and also the website proposal hasn't been approved yet.

### DIFF
--- a/common/_templates/globalomerotoc.html
+++ b/common/_templates/globalomerotoc.html
@@ -2,5 +2,5 @@
 {{ toctree()}}
 <a href="https://downloads.openmicroscopy.org/omero/">{{ _('Downloads by version') }}</a></br>
 <a href="https://docs.openmicroscopy.org/omero/">{{ _('Documentation by version') }}</a></br>
-<a href="https://www.openmicroscopy.org/omero/new/">{{ _('Features') }}</a></br>
+<a href="https://www.openmicroscopy.org/omero/features/">{{ _('Features') }}</a></br>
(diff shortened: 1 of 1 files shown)


hflynn:
Proposal accepted so I'll remove the exclude label once it's deployed live - we can let the redirects handle this until the OMERO release

hflynn:
Deployed live, removing the exclude label for inclusion in tomorrow's build

will-moore:
This might have been mentioned on https://github.com/openmicroscopy/www.openmicroscopy.org/pull/195 but when I click on the Features link from docs, I go to https://www.openmicroscopy.org/omero/features/ which looks like "New Features" page, so I was going to suggest that the link be titled "New Features". But then I read the PR and noticed that that's the home page of ALL features.
I think it's confusing because the main title "New with OMERO" comes above the line of links, so it maybe feels like they are sub-sections of "New with OMERO". I can see the same approach is used on other pages too, and I don't think I've found it confusing before, so maybe it's just when landing at the website or that section for the first time? 
Maybe the home page for a section, e.g. features, should have a main title that applies to the whole section. E.g. "New with OMERO" -> "OMERO features" ?

hflynn:
At the moment, the text in the headers is set for all the features page as 'xxx with OMERO' but i could open a PR to change it.

In the meantime though, I assume you're happy for this to be merged?

will-moore:
Yes, good thanks

PR #1843, opened 2018-02-08
Rephrase JMX target doc

hflynn:
See https://trello.com/c/vNz5ejDW/20-jmx-target-docs-arent-quite-correct

### DIFF
--- a/omero/sysadmins/server-advanced-configuration.rst
+++ b/omero/sysadmins/server-advanced-configuration.rst
@@ -116,9 +116,9 @@ JMX configuration
             <option>-Dcom.sun.management.jmxremote.ssl=false</option>
         </target>
 
-The JMX target activates the monitoring of the Blitz server via JMX. If
-you need to modify the "jmxport" or "jmxhost" variables, you will need
-to do so directly in the application descriptor XML.
+The JMX target enables remote connections for external monitoring of the Blitz
+server. If you need to modify the "jmxport" or "jmxhost" variables, you will
+need to do so directly in the application descriptor XML.
 
 Changing ports / multiple servers on a single host
 --------------------------------------------------


joshmoore:
No objections to the wording, but it may be as easy to just allow setting the properties as it is to explain how to modify the XML: https://github.com/openmicroscopy/openmicroscopy/blob/develop/etc/templates/grid/templates.xml#L176

hflynn:
Does that mean you want me to rewrite the next sentence too? Happy to do so if you have a suggestion

joshmoore:
Only when/if there's a merged code change. Leaving to you if you want to do that here or elsewhere.

hflynn:
Ahhh, I get you now. Seems unlikely there'll be a code change there for 5.4.4 unless that's something you want to get in @jburel ?

jburel:
From this morning discussion I don't think @manics or @kennethgillen will work on it for 5.4.4.
I will confirm tomorrow

manics:
I've no idea what the code change is

joshmoore:
The code change would be to make `jmxhost` and `jmxport` configurable (likely via `omero.jmx.host` and `omero.jmx.port`). One catch is that if we want to enable for PixelData and Indexer (or in the future for microservices) then we need more than 1 port.

manics:
In case it's relevant the IDR uses a javaagent:
- https://github.com/IDR/deployment/blob/IDR-0.4.4/ansible/management-prometheus.yml#L32
- https://github.com/IDR/deployment/blob/IDR-0.4.4/ansible/management-prometheus.yml#L95

joshmoore:
Happy with this change. Getting this in, and we can examine wider JMX changes later.

PR #1844, opened 2018-02-08
Add sudo example for the java gateway

dominikl:
Adds an example for using 'sudo' with the Java Gateway to the Java examples, as @bramalingam recently tried to do something similar and noticed that there's actually no example/documentation about it.

~~I just tested the code manually against eel. There's no working example in the 'training' examples, as it's quite a hassle to set that up (first have to create the user to sudo for, then  create a dataset for that user (while actually already using 'sudo'), etc.).~~ - See https://github.com/openmicroscopy/openmicroscopy/pull/5652 

Staged at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ws/src/omero/_build/html/developers/Java.html

### DIFF
--- a/omero/developers/Java.rst
+++ b/omero/developers/Java.rst
@@ -993,12 +993,13 @@ in OMERO. Similar approach can be applied when uploading an image.
 Sudo (working within another user's context)
 --------------------------------------------
 
-The next example shows how one can work within another user's context. For example
-a data analyst doing some analysis on behalf of a user and attaching the results
-to the user's data. The important point is that the user will be the owner of
-these results and can work with them as usual. The user and analyst don't have to be
-member of a read-annotate group. But the 'analyst' user has to be a 'light admin'
-user with 'sudo' permission.
+The next code snippet shows how you can work within another user's context. This
+could for example be a data analyst doing some analysis on behalf of a user and
+attaching the results to the user's data. The important point is that the user will
+be the owner of these results and can work with them as usual. The user and 'analyst'
+do not have to be member of a read-annotate group (see :doc:`Server/Permissions`),
+but the 'analyst' has to be a 'light administrator' with 'sudo' permission,
+see :doc:`Server/LightAdmins`.
 
 ::
 
--- a/omero/developers/Java.rst
+++ b/omero/developers/Java.rst
@@ -1006,7 +1006,7 @@ see :doc:`Server/LightAdmins`.
     AdminFacility admin = gateway.getFacility(AdminFacility.class);
 
     // Look up the experimenter to sudo for
-    ExperimenterData sudoUser = admin.lookupExperimenter(ctx, sudoUserName);
+    ExperimenterData sudoUser = admin.lookupExperimenter(ctx, sudoUsername);
 
     // Create a SecurityContext for this user within the user's default group
     // and set the 'sudo' flag (i.e. all operations using this context will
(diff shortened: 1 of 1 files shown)


jburel:
Will you open a corresponding code PR?

omero/developers/Java.rst
```diff
@@ -990,6 +990,63 @@ in OMERO. Similar approach can be applied when uploading an image.
         store.close();
     }
 
+Sudo (working within another user's context)
+--------------------------------------------
+
+The next example shows how one can work within another user's context. For example
```
mtbc: I feel like these first two sentences could be linked differently for the better; @hflynn may have a suggestion.

omero/developers/Java.rst
```diff
@@ -990,6 +990,63 @@ in OMERO. Similar approach can be applied when uploading an image.
...
+a data analyst doing some analysis on behalf of a user and attaching the results
+to the user's data. The important point is that the user will be the owner of
+these results and can work with them as usual. The user and analyst don't have to be
```
mtbc: "do not", try to avoid contractions

omero/developers/Java.rst
```diff
@@ -990,6 +990,63 @@ in OMERO. Similar approach can be applied when uploading an image.
...
+member of a read-annotate group. But the 'analyst' user has to be a 'light admin'
```
mtbc: same sentence linking question here

omero/developers/Java.rst
```diff
@@ -990,6 +990,63 @@ in OMERO. Similar approach can be applied when uploading an image.
...
+user with 'sudo' permission.
```
mtbc: maybe a link out to relevant docs?

omero/developers/Java.rst
```diff
@@ -990,6 +990,63 @@ in OMERO. Similar approach can be applied when uploading an image.
...
+
+    // Look up the experimenter to sudo for
+    ExperimenterData sudoUser = admin.lookupExperimenter(ctx, sudoUserName);
+
+    // Create a SecurityContext for this user within the user's default group
+    // and set the 'sudo' flag (i.e. all operations using this context will
+    // be performed as this user)
+    SecurityContext sudoCtx = new SecurityContext(sudoUser.getGroupId());
+    sudoCtx.setExperimenter(sudoUser);
+    sudoCtx.sudo();
```
bramalingam, line 1016: Is the sudoCtx necessary, I was able to achieve this without setting the analyst as sudo, Please take a look at this:

https://github.com/bramalingam/omero-example-scripts/blob/c06a5517a759ac5069da6508adff334ce95507d0/processing_scripts/Workflow7_ImageJ_Jython_Analyst.py#L138

dominikl: You'd only need the `sudo()` when writing data on behalf of the user, not for just reading it. That might not be very clear in the example, will try to rephrase that a bit maybe.

dominikl: I'll leave that part like it is at the moment. A bit further down in the code snippet there's actually also an example of creating a group context without sudo, that matches more your use case.

omero/developers/Java.rst
```diff
@@ -990,6 +990,64 @@ in OMERO. Similar approach can be applied when uploading an image.
...
+
+    // Get the sudouser's datasets
+    BrowseFacility browse = gateway.getFacility(BrowseFacility.class);
+    Collection<DatasetData> datasets = browse.getDatasets(sudoCtx, sudoUser.getId());
+    Iterator<DatasetData> i = datasets.iterator();
+    DatasetData first = null;
```
waxenegger: maybe `DatasetData first = datasets.iterator().next()` would suffice to illustrate the use of the first dataset. I realize there is a potential for an exception but that happens with the code as is if the user has no dataset and I think that's ok because the code is there for illustration.

waxenegger:
my code comment is in the spirit of "succinct and to the point" I guess :wink: otherwise great example, easy enough to get the information needed across, and best of all, no need for lengthy preparation/object creation.

omero/developers/Java.rst
mtbc: Could simply add a brief comment that makes the assumption explicit.

waxenegger: that's true

jburel:
The text matches the sample code tested by ci-job
Merging now that the code PR has been merged

PR #1856, opened 2018-02-28
improve linkchecker for all builds

hflynn:
Implements the solution from https://github.com/openmicroscopy/ome-internal/pull/381 for the OMERO and contributing docs


joshmoore:
Due to https://github.com/hflynn/ome-documentation/blob/c07a77857f07c113d45ca01b32e39ef8fd02e2c6/contributing/conf.py#L19 I'd expect the variables to be included from `common`.

hflynn:
Ah, I figured I may as well include it everywhere, but can remove again if you prefer

hflynn:
Builds all green, merging this for 5.4.4 too

PR #1862, opened 2018-03-05
app example

jburel:
Fix example coming from webtest
The code is correct in omero-webtest but it was not correctly copied across (not included)
https://trello.com/c/toYDGCS4/355-bug-createapp-in-web-doc-bug

cc @pwalczysko


hflynn:
Renders as expected in the merge docs, and fix matches Josh's suggestion on the linked card, good to merge for 5.4.4

PR #1872, opened 2018-04-23
fix trivial markup typo

mtbc:
noticed a minor typo at the bottom of https://docs.openmicroscopy.org/omero/5.4.5/sysadmins/server-performance.html


hflynn:
Thanks for spotting that

PR #1882, opened 2018-06-20
Fix script examples

knabar:
The `client.createSession()` call is no longer needed an in fact raises an error.

Also renamed the import of `omero.scripts` in a few places to make it more consistent across the examples.


jburel:
``import omero, omero.scripts as script`` this should also be updated in ``matlab-scripts.rst``
to match the second commit

jburel:
outside the scope of this PR
but do we want to continue promoting the usage of a project without activity for several years now?
cc @sbesson @joshmoore

jburel:
Thanks @knabar

PR #1885, opened 2018-07-10
Update BF release jobs

sbesson:
This PR propagates recent changes applied to the CI jobs to the contributing page and should turn https://ci.openmicroscopy.org/view/Failing/job/CONTRIBUTING-merge-docs/ green.

- following the bio-formats-documentation decoupling, the release jobs are simplified to one job which checks out the tag and the artifacts and builds the component
- many of the CI jobs have been migrated towards containerized Jenkins instances and disabled/shelved from https://ci.openmicroscopy.org/ and are removed from the list.

As an addition, we could potentially expose the current https://web-proxy.openmicroscopy.org/west-ci/ and https://web-proxy.openmicroscopy.org/east-ci/ URLs in these pages as most of the Bio-Formats development happens in these devspaces.


dgault:
Changes look good and updated pages display correctly. PR is good to merge

PR #1888, opened 2018-07-17
LDAP ignore case thorough documentation

chris-allan:
The impact of `omero.security.ignore_case` on non-LDAP users is currently downplayed.

This makes the documentation of a change of this property more thorough and gives an example how to make all usernames to lower case.


manics (approved):

jburel:
Merging so it is included in the coming 5.4.7 release

PR #1889, opened 2018-07-26
remove obsolete notes from OMERO docs

mtbc:
This PR performs some minor tidyup of OMERO documentation, including removing notes about versions long out of security maintenance.


dominikl:
Looks good 👍

joshmoore:
I'd suggest we also give @chris-allan a heads up, but otherwise, I don't see a major blocker to having it in 5.4.8.

chris-allan:
Not convinced loosing the version history is beneficial but I don't feel particularly strongly one way or the other.

PR #1896, opened 2018-08-27
Change the group permissions

jburel:
Indicate how to change the group permissions

cc @pwalczysko

### DIFF
--- a/omero/sysadmins/cli/usergroup.rst
+++ b/omero/sysadmins/cli/usergroup.rst
@@ -160,8 +160,8 @@ optional argument::
 Group change permissions
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-To change the permissions of a group. For example, make the group ``read-annotate-1``
-a read-write group::
+To change the permissions of a group, for example to make the group ``read-annotate-1``
+a read-write group, run::
 
     $ bin/omero group perms --perms='rwrw--' --name='read-annotate-1'
 
--- a/omero/sysadmins/cli/usergroup.rst
+++ b/omero/sysadmins/cli/usergroup.rst
@@ -156,3 +156,15 @@ To copy the group owners, use the same command with the ``--as-owner``
 optional argument::
 
     $ bin/omero group copyusers read-only-1 read-only-2 --as-owner
+
+Group change permissions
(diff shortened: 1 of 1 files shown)


omero/sysadmins/cli/usergroup.rst
```diff
@@ -156,3 +156,15 @@ To copy the group owners, use the same command with the ``--as-owner``
 optional argument::
 
     $ bin/omero group copyusers read-only-1 read-only-2 --as-owner
+
+Group change permissions
```
pwalczysko: ``Group permission level change`` would sound better to me

omero/sysadmins/cli/usergroup.rst
```diff
@@ -156,3 +156,15 @@ To copy the group owners, use the same command with the ``--as-owner``
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+To change the permissions of a group, for example to make the group ``read-annotate-1``
+a read-write group, run::
+
+    $ bin/omero group perms --perms='rwrw--' --name='read-annotate-1'
+
+If you want to change its name to ``read-write-1`` afterwards, run::
```
sbesson, line 168: This feels a bit unrelated to the title of the section which is focused on changing the group permissions. Looking at previous sections, a suggestion I can offer would be to include this information under the `Group management` section i.e. cover the management of the group settings (name, permissions) as well as the group composition (members and owners)?

jburel: I added it there since it was making more sense to change the name after changing its permissions
I think it is better to split into sub-sections so it is easier to spot
Several commands are not immediately "visible" if we put everything under the same banner
you could argue that group copy can also go under group management

sbesson: If a separate section feels more appropriate, no objection. Then the section title needs to be adjusted to reflect the content.

joshmoore: * Group management --> Group membership
 * Group permissions --> Group management or modification

joshmoore:
Works for me. @pwalczysko / @sbesson : any last comments?

sbesson (approved):

PR #1898, opened 2018-08-31
document parallel CLI import

mtbc:
Documents https://github.com/openmicroscopy/openmicroscopy/pull/5828 in the light of https://trello.com/c/5xiUFcMj/37-parallel-import-docs. Staged at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ws/src/omero/_build/html/users/cli/import.html and https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ws/src/omero/_build/html/sysadmins/import-scenarios.html.

### DIFF
--- a/omero/sysadmins/import-scenarios.rst
+++ b/omero/sysadmins/import-scenarios.rst
@@ -258,6 +258,52 @@ Disadvantages
   the same as the file system which the ManagedRepository lives on. This
   prevents the use of network file systems and similar remote shares.
 
+
+.. _parallel_import:
+
+Parallel import
+---------------
+
+Parallel import is an experimental variant of manual CLI import for
+making large-scale imports considerably faster. This feature entails
+*significant risk* and may be changed or removed at any time. If high
+thread counts are specified then the import client or OMERO server may
+stop working. It is therefore imperative that new uses of parallel
+import are first tested with a similar non-production server. Experience
+gained within OME and reported by users will help to make parallel
+import more friendly and safe.
+
+:option:`omero import --parallel-fileset` sets how many filesets are
+imported at the same time. :option:`omero import --parallel-upload` sets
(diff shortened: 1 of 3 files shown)


omero/sysadmins/import-scenarios.rst
```diff
@@ -258,6 +258,52 @@ Disadvantages
...
 
+
+.. _parallel_import:
+
+Parallel import
+---------------
+
+Parallel import is an experimental variant of manual CLI import for
+making large-scale imports considerably faster. This feature entails
+*significant risk* and may be changed or removed at any time. If high
```
joshmoore: > may be changed or removed at any time

May see extensive changes in upcoming version?

omero/sysadmins/import-scenarios.rst
```diff
@@ -258,6 +258,52 @@ Disadvantages
...
+thread counts are specified then the import client or OMERO server may
+stop working. It is therefore imperative that new uses of parallel
```
joshmoore: "stop working"

This doesn't have more risk that other parallelism and we've even taken steps to improve it, so this is likely too strong.

joshmoore (commented):
Thanks, Mark. As high-level feedback, I'd probably be a bit less aggressive with the warnings. I don't think the bar on this new feature has to be higher than general parallelism against the server.

joshmoore (approved):
Thanks, @mtbc. That seems like a good balance between caution and panic. I'll leave you set the bar / deadline / reminder for removing the "experimental" text.

joshmoore:
NB: currently looking into adading an import-bulk.rst similar to import-targets.rst. Perhaps for 5.4.8 we also add an import-performance.rst?

mtbc:
That would make sense.

PR #1900, opened 2018-09-04
5.4.8 updates

joshmoore:
* minor release fixes
 * bulk docs

cc: @dominikl @sbesson @pwalczysko @jburel


omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+can be used to perform a batch of imports with the same or similar
+options. The file is written in a simple YAML syntax and can be placed
+anywhere.
+
+A simple YAML file might look like: ::
+
+    ---
+    continue: "true"
+    transfer: "ln_s"
+    path: "my-files.txt"
```
sbesson: I am going back and forth about whether this section should provide a minimal example i.e. with only `path` or a more informative example. In the first case, we could like to IDR bulk files for real-world examples of usage.

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+-----------------
+
+Path
+^^^^
+
+The ``path`` key specifies a file from which *each individual line*
+will be processed as a separate import. In the simplest case, a single
+file is placed per line as above. For more complex usages, ``path``
+can point to a tab-separated value (TSV) file where each field will
+be interpreted based on ``columns``.
```
sbesson: Is `columns` required if a TSV file is passed or is only the first column considered?

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+and a file with the given name will be written out. Each of these
+scripts can then be used independently.
+
+Common options
+--------------
+
+Otherwise, all the common options from the CLI are available for
+configuration via ``--bulk``:
+
+- checksum_algorithm for faster processing of large files
```
sbesson: eventually this could be cross-linked to the various options in  `import.rst`

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+- continue for processing all files even if one errors
+- exclude for skipping files that have already been imported
+- target for placing imported images into specific containers
+- transfer for alternative methods of shipping files to the server
```
sbesson: Do we want to promote the new `parallel_` options here as well?

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+The ``path`` key specifies a file from which *each individual line*
+will be processed as a separate import. In the simplest case, a single
+file is placed per line as above. For more complex usages, ``path``
+can point to a tab-separated value (TSV) file where each field will
+be interpreted based on ``columns``.
+
+Columns
+^^^^^^^
+
+A fairly regular requirement in importing many files is that for
```
sbesson, line 43: As all the examples below are relying on the common import options, should we move this section above e.g.

1- Common options (vanilla import)
2- Bulk specific options
 a- Path
 b- Columns (with examples including some of the common options)
 c- dry-run

sbesson (commented):
In general, the new page describes most of the logic requires to perform bulk import.

My primary questions are about how to make sure the message is clear. I have put a few ideas as inline comments  but it might be useful to have the opinion of someone who has been less exposed to the details of the code.

omero/users/cli/import-bulk.rst
joshmoore: Yeah, I wasn't sure if we wanted to perhaps conclude with an IDR example.

omero/users/cli/import-bulk.rst
joshmoore: My guess is it will assume the filename has a tab in it.

omero/users/cli/import-bulk.rst
joshmoore: :+1:

omero/users/cli/import-bulk.rst
joshmoore: I was unsure. I didn't want to get into duplicating every option. I tried to pick ones that I had personally used. (I'm less sure one will want to mix performance with correctness options)

omero/users/cli/import-bulk.rst
joshmoore: With "common options" I was trying to get across, "and you can use all the rest". Perhaps a new name and links to other documentation would make that clearer.

joshmoore:
@sbesson : thoughts on moving forward here? Any other thoughts from the team?

sbesson:
I think this could certainly benefit a reading from another member who's less familiar with the code itself but has performed bulk imports (e.g. people involved in IDR and/or training). From my side, happy to see this merged for rc1 and have the option refactoring/cross-linking done as a follow-up for GA.

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
+Using bulk imports
```
pwalczysko: **Using**  ? bulk imports. 

Maybe better simply ``Bulk imports`` 
there might be a false-friend problem with ``Using import targets`` - but this is a different case. Import targets are truly a valid description of a cli option(s) whereas ``bulk imports``  has much wider meaning.

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+==================
+
+The CLI import option ``--bulk`` specifies a configuration file that
+can be used to perform a batch of imports with the same or similar
+options. The file is written in a simple YAML syntax and can be placed
+anywhere.
```
pwalczysko: Probably ``anywhere`` where the command can access it... 
Or maybe rather say

``does not have to be placed in the folder from which the omero commands are run``

omero/users/cli/import-bulk.rst
pwalczysko: Outreach in Basel has shown that the idr examples are far too complex (and even possibly not the most useful for OMERO users without simplifications). 

Also, we were explicitly asked for an example or possiblity to pass a whole folder into the ``tsv`` file instead of single files. 
We tested that on the spot and it worked.
I do not see such example in the whole doc here cc @jburel

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+
+A simple YAML file might look like: ::
+
+    ---
+    continue: "true"
+    transfer: "ln_s"
```
pwalczysko: Another question during the Basel workshop was:

  - how can I use it without the in-place option ?
 
Although it might be argued that this is intuitive (remove the ..ln_s line), it would be still nicer if we would explicitly suggest it (or even not start from a strong silent assumption that in-place is a default here - might make more sense, but obviously users want the full range of options)

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+
+The CLI import option ``--bulk`` specifies a configuration file that
+can be used to perform a batch of imports with the same or similar
+options. The file is written in a simple YAML syntax and can be placed
+anywhere.
+
+A simple YAML file might look like: ::
+
+    ---
+    continue: "true"
```
pwalczysko: never really understood why this line is necessary. But it is again silently assumed it will be in every bulk.yml the users would create... Is this the same as the ``-k`` ? I would say no...

Should be explained and hinted that this is necessary for every bulk.yml if I am right.

omero/users/cli/import-bulk.rst
jburel: I made an example in basel to import folder only
but I used csv.
cvs might be more natural in some cases

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+This can be accomplished with the ``columns`` key. It specifies how
+each of the tab-separated fields of the ``path`` file (e.g. "my-files.txt")
+should be interpreted. For example, a bulk.yml file specifying::
+
+    path: "files.tsv"
+    columns:
+    - name
+    - path
+
+along with a files.tsv of the form::
```
pwalczysko: Question from audience during the Basel workshop:

   what is this mysterious .tsv format ? Do I have to create it in some special software ? Maybe worth mentioning ``tab separated value``

omero/users/cli/import-bulk.rst
jburel: agree that we should not make the users assume that transfer is the "default" (it is technically not a default option) when using the bulk command
also since no files are moved having checksum is confusing for some users

omero/users/cli/import-bulk.rst
jburel: I still think that csv will be more natural than tsv in the example

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+    import-1	fileA
+    import-2	fileB
+
+would match the two calls::
+
+    $ bin/omero import --name import-1 fileA
+    $ bin/omero import --name import-2 fileB
+
+but in a single call. Another key that can be valuable in the ``columns``
+section is ``target``::
```
pwalczysko: ``...columns section``  -> ``columns section of the bulk.yml file`` (be more precise)

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    Dataset:name:training-set	fileA
+    Dataset:name:training-set	fileB
+    Dataset:name:test-set-001	fileC
```
pwalczysko: No ``equivalent to`` example here. Especially the  fact that new datasets will be created is not totally obvious. Maybe link to the ``Using import targets`` section of the docs.

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Include
+^^^^^^^
+
+The ``include`` key specifies another bulk YAML file that should be
+included in the current processing. This allows having a global
+configuration file that you use in various situations::
```
pwalczysko: Pretty unclear how does the ``inclusion`` work. Is it an addition ? Replacement of what is in the ``.yml`` file into which I am inserting this line ? Am I inserting this line into the ``bulk.yml`` file at all ? Not clear either. I am afraid without a full example of the two files and the infrastructure in which these are used, not very helpful.

omero/users/cli/import-bulk.rst
pwalczysko: sorry, missed https://github.com/openmicroscopy/ome-documentation/pull/1900/files#diff-d09750c6e463153813e3849d4ec6f4b8R35 - this is fine then

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+Bulk-only options
+-----------------
+
+Path
+^^^^
+
+The ``path`` key specifies a file from which *each individual line*
+will be processed as a separate import. In the simplest case, a single
+file is placed per line as above. For more complex usages, ``path``
+can point to a tab-separated value (TSV) file where each field will
```
pwalczysko: CSV example during the Basel workshop was

   - working
   - more understandable than the TSV file (see my comment below about TSV, which I made before reading this bit here)

pwalczysko: cc @jburel

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,99 @@
...
+be interpreted based on ``columns``.
+
+Columns
+^^^^^^^
+
+A fairly regular requirement in importing many files is that for
+each file a similar but slightly different configuration is needed.
+This can be accomplished with the ``columns`` key. It specifies how
+each of the tab-separated fields of the ``path`` file (e.g. "my-files.txt")
+should be interpreted. For example, a bulk.yml file specifying::
```
pwalczysko: ``bulk.yml`` here could be formatted better

omero/users/cli/import-bulk.rst
jburel: writing a single file per line might put people off
The question in Basel was immediately: can I specify folder?
The "single file" did not really sell the feature

omero/users/cli/import.rst
```diff
@@ -74,6 +74,20 @@ Scanning folders prior to Import
...
+
+.. program:: omero import
+
+.. option:: --bulk YAML_FILE
+    
+    To import a number of images with a similar configuration::
+
+        $ bin/omero import --bulk bulk.yml
+
+    See :doc:`import-bulk` for more information on bulk imports.
```
pwalczysko, line 88: If the section is renamed to simply ``Bulk imports`` as I suggested above, then here it would go simply

``See Bulk imports for more information`` (with the link on the words ``Bulk imports``.

(Now it reads "See Using bulk imports for more information on bulk imports.")

sbesson:
Apart from grammatical and formulation changes, I think the sections which need more discussion might be:

- page title: `Bulk imports`?
- first example: given the outreach feedback, I would propose to create a minimal bulk YML file with only one line `path` and  a txt file maybe with 2 lines, a file and one folder, showing that both forms are supported
- `columns`: need to explain `csv/tsv` are both supported, define these formats, switch to `csv` (or use one example of both) are and expand the examples,
- advanced import options: add more cross-references to the import page and links to IDR advanced bulk examples

I would vote for fixing minimally the PR title but not blocking its inclusion on the 3 following points. The minimal example as well as the `columns/csv/tsv` section should probably be reviewed for 5.4.8. Advanced import options is a nice-to-have but I would not block 5.4.8 on it.

jburel:
java-development page: https could be used for https://imagej.net/Development_Lifecycle and https://javadoc.io/

pwalczysko:
>page title: Bulk imports?

+1

omero/users/cli/import-bulk.rst
joshmoore: I will add a directory example, but similar to @sbesson's statement, I would prefer to *not* try to get all of your knowledge/examples from training into this PR. If someone would like to open a follow-up, feel free.

omero/users/cli/import-bulk.rst
joshmoore: I don't understand what you mean, @jburel.

omero/users/cli/import-bulk.rst
joshmoore: I'm going to push a commit changing "Using import targets" to just "Import targets". My primary goal was consistently in the index.

omero/users/cli/import-bulk.rst
joshmoore: I'll try an absolute minimal example (i.e. just path) at the beginning and build up from there.

omero/users/cli/import-bulk.rst
joshmoore: > Is this the same as the -k

Yes.

> Should be explained and hinted that this is necessary for every bulk.yml if I am right.

No. It is not needed.

omero/users/cli/import-bulk.rst
joshmoore: Is it clearer now?

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,128 @@
...
+options. The file is written in a simple YAML syntax and can be named
+whatever you would like. It does not need to be placed in the folder
+from which the OMERO commands are run.
+
+A minimal YAML file might look like: ::
+
+    ---
+    path: "my-files.txt"
+
+Assuming that my-files.txt is a list of files such as ::
```
mtbc: Maybe consider `:file:` markup for these. (*Update:* Done.)

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,128 @@
...
+
+    fileA
+    fileB
+    directoryC
+
+this is equivalent to: ::
+
+    $ bin/omero import -k --transfer=ln_s fileA fileB directoryC
+
+where the files fileA and file B and all the files of directoryC will be imported.
```
mtbc: `fileB`

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,128 @@
...
+
+Other options
+-------------
+
+Otherwise, all the regular options from the CLI are available for
+configuration via ``--bulk``:
+
+- checksum_algorithm for faster processing of large files
+- continue for processing all files even if one errors
+- exclude for skipping files that have already been imported
```
mtbc: I feel as if these options need *some* visual distinctiveness from the rest of the line, whether literal markup or some Sphinx notion of definition lists or even just a suffixed colon or something.

joshmoore:
@mtbc: pushed

omero/users/cli/import-bulk.rst
jburel: if you copy one of our examples, the tab usually disappears when the content is copied in a text editor, leading to some errors when trying to run the command. This can be tricky to understand/debug.

Using csv file avoids that issue

sbesson (approved):
With the last changes, the new bulk import page reads fine. I believe most of the outstanding comments by @jburel and @pwalczysko are addressed so I would vote for merging and having follow-up PRs improving specific sub-sections of the new documentation page.

joshmoore:
Force pushed away an unintended diff:

```
diff --git a/contributing/cpp-development.rst b/contributing/cpp-development.rst
index 1aec2d1ae..5df90a37d 100644
--- a/contributing/cpp-development.rst
+++ b/contributing/cpp-development.rst
@@ -102,6 +102,8 @@ the process described in the next section.
 Release process
 ---------------

+TODO: ome-model
+
 Maintainer prerequisites
 ^^^^^^^^^^^^^^^^^^^^^^^^

@@ -117,6 +119,11 @@ Source release
 The first step of the C++ component release is to prepare a source release
 from the Git repository.

+OME model
+
+then acceptance
+and redirects
+
 Prior to a source release, a PR should be opened and merged to:

 - review the ``release-version`` variable in :file:`CMakeLists.txt` and drop
```

Otherwise matches the last comment from @sbesson

omero/users/cli/import-bulk.rst
```diff
@@ -0,0 +1,130 @@
...
+The CLI import option ``--bulk`` specifies a configuration file that
+can be used to perform a batch of imports with the same or similar
+options. The file is written in a simple YAML syntax and can be named
+whatever you would like. It does not need to be placed in the folder
+from which the OMERO commands are run.
+
+A minimal YAML file might look like: ::
+
+    ---
+    path: "my-files.txt"
```
jburel, line 13: This could be a bit confusing we start with ``.txt`` then quickly indicate ``.tsv`` and ``.csv`` below

joshmoore: What would you suggest?

jburel:
One minor comment 
much clearer now thanks

omero/users/cli/import-bulk.rst
jburel: either csv or tsv

jburel: so like that users won't have to change extension when they start including some options described below

joshmoore: For me, a file with one column doesn't immediately map to csv/tsv. For example, if I do:

```
find /my/data -name "*.dv" | tee my-files...
```

I would never think to name that csv or tsv.

jburel: I can see your point.
Let's keep it as it is
The page can be updated later on if required

PR #1904, opened 2018-09-14
neubias prep

jburel:
Example fixes


pwalczysko:
Considering the doc edited in this PR, I have couple of other suggestions:


In http://docs.openmicroscopy.org/omero/5.4.8/developers/Matlab.html#creating-a-connection

I am confused by the code snippets here. They seem overly complex, why do I have to create 2 clients and two sessions ???

Also, the "explanatory" sentence ``Alternatively, you can pass the same parameters to loadOmero; that you would pass to omero.client:`` is rather confusing things - we have to assume that the reader have no idea at this stage about ``what would they pass to omero.client``, and there is no reference to ``omero.client`` docs either.


```
>> omero_client_1 = loadOmero('localhost');
>> omero_client_2 = omero.client('localhost');
....

**Comment by pwalczysko** [comment]
the keep alive seciton

http://docs.openmicroscopy.org/omero/5.4.8/developers/Matlab.html#keeping-your-session-alive

Replace the ``[c,s]`` with ``[client, session]``
Even then, it will not work as written, even as the client and session are already created. The line ``[client,session] = loadOmero();`` just crashes. I do not see why it is there actually, as I made sure about the client and session creation in previous section, I shoiuld be able to just keep them alive, no ?

**Comment by pwalczysko** [comment]
In keep alive section

```
>> [c,s] = loadOmero;
>> t = omeroKeepAlive(c); % Create a 60-second timer and starts it
>> …
>> delete(t);             % Disable the keep-alive
```
does this really create a 60-second timer ? I do not see a ``60`` passed anywhere ? 

Also note that ``Create`` above shoiuld be ``Creates``

In general, I think this section could do with a sentence explaining that the  keep alive routines here are poking the server in 60 second intervals (otherwise, it might seem like we are suggesting to keep the session alive only for 60 seconds, which is confusing). The sentence is semi-implicitly in the java code snippet just below the above one, but I think it should be in the header of the keep alive section.

**Comment by pwalczysko** [comment]
``The Image-Pixels model implies you need to use the Pixels objects to access valuable data about the Image:``

What Image-Pixels model ? I kind of know what is meant, but not from the doc itself, rather from "general knowledge of OMERO" - it is never mentioned in the matlab doc previously, and no link to the explanatory doc about Image-Pixels model is offered.

**Comment by pwalczysko** [comment]
```
myLoadedProjects = getProjects(session, [], true)
loadedProjects = getProjects(session, ids, true)
imageList = loadedProjects(1).linkedDatasetList.get(0).linkedImageList;
```
this bit of code has no explanation with regards:

   - it just loads the images of the first dataset (``get(0)``). I also do not understand why the Projects are iterated from ``(1)`` and the Datasets from ``(0)`` ?
   - the accompanying text actually suggests you will get all the images in the whole project like this, not getting into any details

**Comment by pwalczysko** [comment]
So this code works for me now as ``Get all the images in a Project``. It has some redundant bits (like retrieval of the dataset names, image names, image ids etc.) but I think this might be how you have done it (I do not study your code on purpose, just trying to go through the exercise myself)

```
loadedProjects = getProjects(session, 1001, true);
datasetList = loadedProjects(1).linkedDatasetList;
for i = 0:datasetList.size()-1,
    dataset = loadedProjects(1).linkedDatasetList.get(i);
    dataset.getName().getValue()
    imageList = dataset.linkedImageList;
    for k = 0:imageList.size()-1, 
        image = imageList.get(k);
        image.getName().getValue();
        image.getId().getValue();
    end
end;
```

**Comment by pwalczysko** [comment]
``The plane of an input image at coordinates (z, c, t) can be retrieved using the getPlane function:``

I am guessing that the indices here start from ZERO. This is not stated in the doc.

**Comment by pwalczysko** [comment]
There is no mention of the ``channel`` in the doc, except when retrieving the number of channels. But the exercise wants the name of the channel to be retrieved. This seems hard.

**Comment by pwalczysko** [comment]
Follow-up on the ``channel names`` story https://github.com/openmicroscopy/ome-documentation/pull/1904#issuecomment-426676558 - I am able to get the channels now with

```
...
        pixels = image.getPrimaryPixels();
        pixels.getChannel(0)
```
but I get the answer that


```
Java exception occurred:
omero.UnloadedCollectionException: Error updating collection:channelsSeq
Collection is currently null. This can be seen
by testing "channelsSeqLoaded". This implies
that this collection was unloaded. Please refresh this object
in order to update this collection.
```

This means -> dead end for me, cannot even crack it with our java docs etc., 
This seems not hard, but rather impossible without an example.

**Comment by pwalczysko** [comment]
Okay, so the right answer to https://github.com/openmicroscopy/ome-documentation/pull/1904#issuecomment-426687376 seems to be 

``channels = loadChannels(session, image);``

Nowhere to be found but your script, at least I could not find it.
I guess a section here is missing in this doc, which should be added to the bit mentioned in my comment https://github.com/openmicroscopy/ome-documentation/pull/1904#issuecomment-426657887 - basically, how to properly load objects. Obviously, the answer is different for P/D/I hierarchy and again different for Image/Channels hierarchy.  We could also remove the repetition of the warning how heavy it is to load the whole graph (we prevent this by not telling how to do it in the first place :)

**Comment by pwalczysko** [comment]
Aha, okay. So getting the channel is still not the end - "simple" getting of the channel name (getName().getValue()) does not work either....
Really not sure what the participants levels must be to get this - maybe I am missing some crucial Matlab tricks (but will they know them ?)

File: omero/developers/Matlab.rst [hunk 1]
```diff
@@ -483,52 +509,69 @@ method::
...
             ws.getId().getValue()
             pa = ws.getPlateAcquisition();
         end
     end
 
+-  **Channel**
+
+A channel associated to an image has an object called a logicalChannel associated to it.
+That entity contains valuable information e.g. emission wavelength, name, etc.
+Given an Image, retrieve channels associated to an image on the OMERO server::
```
pwalczysko: ``...retrieve channels associated to an image on the OMERO server:`` -> ``..retrieve channels associated to an image on the OMERO server and the name of the channel:``

pwalczysko:
This all looks good. Ready to merge FMPOV

omero/developers/Matlab.rst
```diff
@@ -51,7 +51,7 @@ As described under |OmeroClients|, there are several ways to configure your
 connection to an OMERO server. OMERO.matlab comes with a few conveniences for
 making this work.
 
-If you run ``client = loadOmero();`` (i.e. loadOmero with an output argument),
+If you run ``client = loadOmero();`` (i.e. loadOmero without an input argument),
```
sbesson, line 54: I see why the formulation was confusing and the change was made but note the original description still applies. Passing an output argument is still mandatory for the connection to be made or said otherwise `loadOmero()` will *not* create a connection - see
https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroM/src/loadOmero.m#L100

sbesson (commented):
One comment about the connection changes. Merging as this can be sorted as a follow-up

PR #1913, opened 2018-10-29
where to get client JAR for CLASSPATH

mtbc:
Until we fix https://trello.com/c/RbxNUNHD/17-omeroclientsjar we should warn people how to find `omero_client.jar` as it was far from clear to me. Staged at https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ws/src/omero/_build/html/developers/Java.html#extended-classpath.

### DIFF
--- a/omero/developers/Java.rst
+++ b/omero/developers/Java.rst
@@ -33,11 +33,13 @@ The required :file:`.jar` files can be obtained in a number of ways:
 Extended classpath
 ------------------
 
-To access all the functionality available in omero\_client.jar or to use the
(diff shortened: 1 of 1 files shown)


mtbc:
To what extent the linked page is usefully a "download page" is another question but that's an issue outwith this PR that is separately tracked in Trello; this PR just tries to stay consistent with the other developer documentation's approach.

joshmoore:
Certainly seems fine as a quick fix. Obviously, the _why_ is confusing but that's also outwith this PR. :smile:

PR #1915, opened 2018-11-12
Update BlitzGateway docs

manics:
Update to include examples of use as a context manager. See https://github.com/openmicroscopy/openmicroscopy/pull/5887

### DIFF
--- a/omero/developers/Python.rst
+++ b/omero/developers/Python.rst
@@ -61,24 +61,70 @@ client to choose suitable data IDs before editing and running the code
 samples.
 
 
+.. _python-code-samples:
+
 Code samples
 ------------
 
 Connect to OMERO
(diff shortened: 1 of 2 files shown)


joshmoore:
Generally looks good. There might be a few small cleanups that I would suggest, but not having a passing build at the moment (cF. [first run](https://web-proxy.openmicroscopy.org/east-ci/job/OMERO-docs/1)) I'm a bit hesitant. I'll either get a build working tomorrow or move to local build for making suggestions.

joshmoore:
`Downloads/OMERO.doc-5.4.6-DEV/developers/Python.html#python-code-samples` looks good. Merging and if further changes are wanted (e.g. bringing "You must close" up higher) then that can be done later. Thanks, @manics.

PR #1923, opened 2018-12-17
Web doc

jburel:
Update the web installation doc

See https://github.com/ome/omeroweb-install/pull/44


joshmoore:
NB: the change in the whitespace seems unfortunate if it wasn't intentional.

jburel:
I noticed that too. Nothing intentional and nothing changes on that front with the generation using the ansible

rgozim:
Docs are looking okay. Links are working.

PR #1944, opened 2019-04-15
Matlab: link to nice downloads page instead of listing

manics:
https://docs.openmicroscopy.org/omero/5.4.10/developers/Matlab.html suggests downloading the Matlab client package from https://downloads.openmicroscopy.org/omero/5.4.10/ which is an unhelpful file listing. This changes the link to https://www.openmicroscopy.org/omero/downloads/.


jburel:
Thanks merging
This might change again. We need to discuss this coming week the location of the artifacts

PR #1950, opened 2019-05-02
set bf value

jburel:
hard code value for now
Further work is required to extract value from properties file


jburel:
Merging to check if we have other problem in the release pipeline

PR #1957, opened 2019-05-09
System prerequisites for OMERO version 5.5.0.

mtbc:
Work on https://trello.com/c/4yy8WRC6/11-doc-version-review to try to bring https://ci.openmicroscopy.org/job/OMERO-DEV-merge-docs/ws/src/omero/_build/html/sysadmins/version-requirements.html in line with OMERO 5.5.0 expectations. Corrections welcome.

### DIFF
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -2,17 +2,25 @@
 Version requirements
 ********************
 
-Introduction
-============
+Summary of changes for OMERO 5.4 and provisional changes for 5.5
+================================================================
 
 Criteria for what is considered to be supportable includes whether
 support by both upstream developers and operating system distributions
-will be available for the lifetime of the |version| release (including
+will be available for the lifetime of the 5.4 release (including
 security support), and also upon our resources allocated to CI and
 testing. If we are not actively testing it, we **cannot** claim it is
 supported or functional. Software components must be provided and
 supported either by an operating system distribution or their original
 developers.
+
+This section contains a summary of the changes made to the minimum
+version requirements for the 5.4 release and also **possible** changes for
+the following 5.5 release, albeit **tentatively at this point**. The
+intent is to provide a roadmap in order that sysadmins may plan ahead
+and ensure that prerequisites are in place ahead of time to ease
+future upgrades. The following sections provide more detailed
+information and the rationale for the changes.
 
 Please check the meaning of the various :ref:`symbols <support-levels>` used
 on this page. Any system/component with a green tick symbol is supported.
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -267,6 +267,10 @@ Components
          - |Upcoming|
          - |Supported|
          - |Recommended|
+       * - 1.16
+         - |Upcoming|
+         - |Upcoming|
+         - |Upcoming|
 
 .. _support-levels:
 
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -2,25 +2,17 @@
 Version requirements
 ********************
 
-Summary of changes for OMERO 5.4 and provisional changes for 5.5
-================================================================
+Introduction
+============
 
 Criteria for what is considered to be supportable includes whether
 support by both upstream developers and operating system distributions
-will be available for the lifetime of the 5.4 release (including
+will be available for the lifetime of the |version| release (including
 security support), and also upon our resources allocated to CI and
 testing. If we are not actively testing it, we **cannot** claim it is
 supported or functional. Software components must be provided and
 supported either by an operating system distribution or their original
 developers.
-
-This section contains a summary of the changes made to the minimum
-version requirements for the 5.4 release and also **possible** changes for
-the following 5.5 release, albeit **tentatively at this point**. The
-intent is to provide a roadmap in order that sysadmins may plan ahead
-and ensure that prerequisites are in place ahead of time to ease
-future upgrades. The following sections provide more detailed
-information and the rationale for the changes.
 
 Please check the meaning of the various :ref:`symbols <support-levels>` used
 on this page. Any system/component with a green tick symbol is supported.
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -752,13 +752,13 @@ If no version is provided, a suitable repository is indicated.
       - N/A
       - N/A
     * - 3.6
-      - 6, 7 (`zeroc <https://zeroc.com/distributions/ice/>`__)
-      - 14.04, 16.04 (`zeroc <https://zeroc.com/distributions/ice/>`__)
+      - 6, 7 (`zeroc <https://zeroc.com/distributions/ice/3.6/>`__)
+      - 14.04, 16.04 (`zeroc <https://zeroc.com/distributions/ice/3.6/>`__)
       - Yes
       - Yes
     * - 3.7
-      - 7 (`zeroc <https://zeroc.com/distributions/ice/>`__)
-      - 16.04, 18.04 (`zeroc <https://zeroc.com/distributions/ice/>`__)
+      - 7 (`zeroc <https://zeroc.com/distributions/ice/3.7/>`__)
+      - 16.04, 18.04 (`zeroc <https://zeroc.com/distributions/ice/3.7/>`__)
       - Yes
       - Yes
     * - Details
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -427,8 +427,6 @@ Package lists
       - `Reference <http://mirror.centos.org/centos/6/os/x86_64/Packages/>`__
     * - CentOS 7 / RHEL 7
       - `Reference <http://mirror.centos.org/centos/7/os/x86_64/Packages/>`__
-    * - CentOS 8 / RHEL 8
-      - `Reference <http://mirror.centos.org/centos/8/os/x86_64/Packages/>`__
     * - Ubuntu
       - `Reference <https://packages.ubuntu.com/search?keywords=foo&searchon=names&suite=all&section=all>`__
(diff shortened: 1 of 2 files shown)


omero/sysadmins/version-requirements.rst
```diff
@@ -141,107 +148,111 @@ Components
...
        :header-rows: 1
 
        * - Java
-         - OMERO 5.3
          - OMERO 5.4
          - OMERO 5.5
+         - OMERO 5.6
        * - 7
          - |Deprecated|
          - |Deprecated|
```
sbesson: Should this be switched to `|Dropped|` ?

omero/sysadmins/version-requirements.rst
```diff
@@ -141,107 +148,111 @@ Components
...
          - |Recommended|
          - |Recommended|
          - |Recommended|
          - |Recommended|
-         - |Recommended|
        * - 3.x
          - |Broken|
          - |Broken|
          - |Broken|
          - |Broken|
```
sbesson, line 179: If we are still using this page to indicate our provisional plans for the next major release of OMERO, Python 3.x support is undoubtedly the next large body of work in terms of requirements.

mtbc: It is. I didn't dare hope that we expect to support it for OMERO 5.6 but perhaps I ought to be more optimistic?

omero/sysadmins/version-requirements.rst
```diff
@@ -2,26 +2,18 @@
...
-will be available for the lifetime of the 5.4 release (including
+will be available for the lifetime of the |version| release (including
 security support), and also upon our resources allocated to CI and
 testing. If we are not actively testing it, we **cannot** claim it is
 supported or functional. Software components must be provided and
 supported either by an operating system distribution or their original
 developers.
 
-This section contains a summary of the changes made to the minimum
-version requirements for the 5.4 release and also **possible** changes for
```
joshmoore, line 18: As @sbesson mentioned, I'd think we either remove the OMERO 5.6 column, or add a versino of text back in to make the tentativeness clear.

omero/sysadmins/version-requirements.rst
```diff
@@ -45,34 +37,42 @@ Operating systems
...
        - |Dropped|
+       - |Dropped|
      * - 7.x
        - |Recommended|
        - |Recommended|
        - |Recommended|
+     * - 8.x
+       - |Upcoming|
+       - |Upcoming|
+       - |Upcoming|
```
joshmoore: Not `Recommended` for the next version? If it's tentative, it seems like we should at least tentatively suggest what the recommendation will be.

omero/sysadmins/version-requirements.rst
```diff
@@ -111,28 +110,36 @@ Components
...
        - |Supported|
        - |Supported|
-       - |Supported|
+       - |Deprecated|
      * - 9.6
-       - |Recommended|
-       - |Recommended|
-       - |Recommended|
+       - |Supported|
+       - |Supported|
```
joshmoore, line 138: Would OMERO 5.5 recommend PG 9.6?

omero/sysadmins/version-requirements.rst
```diff
@@ -111,28 +110,36 @@ Components
...
+       - |Deprecated|
+     * - 10
+       - |Upcoming|
+       - |Supported|
+       - |Supported|
```
joshmoore, line 143: I'd imagine OMERO 5.6 would recommend at least PG 10.

omero/sysadmins/version-requirements.rst
```diff
@@ -141,107 +148,111 @@ Components
...
-         - |Supported|
          - |Broken|
          - |Supported|
          - |Dropped|
+         - |Dropped|
        * - 2.7
          - |Recommended|
          - |Recommended|
          - |Recommended|
          - |Recommended|
```
joshmoore: I'd suggest deprecated if not dropped for Python 2.7 is OMERO 5.6

omero/sysadmins/version-requirements.rst
joshmoore: I think we have to be optimistic about getting to Python 3 at this point.

joshmoore (changes requested):
First round of reading (from the top). I haven't cross-checked the top and the bottom values.

omero/sysadmins/version-requirements.rst
jburel: This will require some adjustments to omero-install
PG10 is already supported. We need to review the autogen part. I will review that

jburel: See https://github.com/ome/omero-install/pull/195

omero/sysadmins/version-requirements.rst
```diff
@@ -561,7 +574,7 @@ OMERO support policies
       - to Jan 2020
       - |Recommended|
       - |Recommended|
-      - |Recommended|
+      - |Deprecated|
```
joshmoore, line 579: Potentially even broken, but I'm happy to take responsibility for elevating that when the time comes.

omero/sysadmins/version-requirements.rst
```diff
@@ -842,24 +913,14 @@ If no version is provided, a suitable repository is indicated.
...
-    * - 1.10
-      - 6 (`EPEL <https://dl.fedoraproject.org/pub/epel/6/x86_64/>`__), 7 (`EPEL <https://dl.fedoraproject.org/pub/epel/7/x86_64/>`__)
-      - 14.04 (`nginx <https://launchpad.net/~nginx/+archive/ubuntu/stable>`__), 16.04
-      - Yes
       - Yes
-    * - 1.12
-      - 6 (`EPEL <https://dl.fedoraproject.org/pub/epel/6/x86_64/>`__), 7 (`EPEL <https://dl.fedoraproject.org/pub/epel/7/x86_64/>`__)
-      - 14.04 (`nginx <https://launchpad.net/~nginx/+archive/ubuntu/stable>`__), 16.04
+    * - 1.14
+      - N/A
```
joshmoore, line 924: What does the N/A mean here? If it means, no repo is needed, then either "Yes" from this table or perhaps "7" from the Java table would be appropriate. I'm a bit confused.

joshmoore (changes requested):
:+1: for the general "each requirement has something `supported`" updates. But one minor outstanding comment though.

omero/sysadmins/version-requirements.rst
mtbc: I don't know what it means but it pervades the existing Homebrew columns. As far as I could see the pattern seemed to be that it tends to follow old versions, hence my guess here, though I wonder if one could go back in Homebrew's history to resurrect them. Would certainly be glad for somebody who knows about Homebrew stuff to either guide or fix; could be a separate PR.

omero/sysadmins/version-requirements.rst
```diff
@@ -141,107 +156,116 @@ Components
...
+         - |Deprecated|
        * - 1.12
+         - |Supported|
+         - |Recommended|
+         - |Supported|
+       * - 1.14
          - |Upcoming|
-         - |Upcoming|
+         - |Supported|
          - |Recommended|
```
manics, line 269: Nightshade OMERO.web and the next IDR release have the latest Enterprise Linux 7 stable version of Nginx: `nginx-1.16.0-1.el7.ngx.x86_64`, so I think we could say `1.16` is recommended.

manics: Though  a quick look at the current distro versions (as opposed to the upstream Nginx repos) shows:
- ubuntu:18.04: 1.14
- centos:7 via EPEL: 1.12
- ubuntu:16.04: 1.10

So it's a toss-up between recommending 1.14 and 1.16 for 5.5

omero/sysadmins/version-requirements.rst
manics: Since https://github.com/ome/omero-install/pull/195 makes PG 10 the default should that also be the recommended version for 5.5?

omero/sysadmins/version-requirements.rst
```diff
@@ -393,6 +427,8 @@ Package lists
       - `Reference <http://mirror.centos.org/centos/6/os/x86_64/Packages/>`__
     * - CentOS 7 / RHEL 7
       - `Reference <http://mirror.centos.org/centos/7/os/x86_64/Packages/>`__
+    * - CentOS 8 / RHEL 8
+      - `Reference <http://mirror.centos.org/centos/8/os/x86_64/Packages/>`__
```
manics: FYI it may be a while before CentOS8 is released https://wiki.centos.org/About/Building_8

omero/sysadmins/version-requirements.rst
manics: This is confusing, there's two Nginx sections in this doc and this one is missing the 1.16 line. Since Nginx isn't as critical as before (it's proxying OMERO.web rather than using FastCGI as in older versions) perhaps this section could be removed to avoid duplication?

omero/sysadmins/version-requirements.rst
```diff
@@ -702,7 +758,7 @@ If no version is provided, a suitable repository is indicated.
       - Yes
     * - 3.7
       - 7 (`zeroc <https://zeroc.com/distributions/ice/>`__)
-      - 16.04 (`zeroc <https://zeroc.com/distributions/ice/>`__)
+      - 16.04, 18.04 (`zeroc <https://zeroc.com/distributions/ice/>`__)
```
manics: `https://zeroc.com/distributions/ice/` goes to the 3.7 downloads. This is fine for this line, but the 3.6 links in the previous line will need changing.

omero/sysadmins/version-requirements.rst
mtbc: I don't think I'm the one to judge. I'll add a 1.16 line for now in the hope that larger changes can be in subsequent PRs. :smiley:

manics (approved):
Looks good to me, any other refinements can happen in later PRs

joshmoore (approved):

PR #1970, opened 2019-05-23
Contributing links

jburel:
Update links to properties files

### DIFF
--- a/contributing/schema-changes.rst
+++ b/contributing/schema-changes.rst
@@ -59,7 +59,7 @@ needed to populate those fields in the proxy object.
 Database patch numbers
 ----------------------
 
-:omero_source:`omero.properties <etc/omero.properties>` contains a
+:omero_model_source:`omero-model.properties <src/main/resources/omero-model.properties>` contains a
 configuration setting for :literal:`omero.db.patch`. An existing OMERO
 database records the patch number of its schema, as demonstrated from
 the :literal:`psql` shell:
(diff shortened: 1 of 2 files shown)


contributing/schema-changes.rst
```diff
@@ -73,10 +73,11 @@ the :literal:`psql` shell:
         (1 row)
 
 indicating that a database is on patch version 4. Correspondingly,
+in :omero_subs_github_repo_root:`omero-model`,
 
 ::
 
-        $ grep ^omero.db.patch= etc/omero.properties
+        $ grep ^omero.db.patch= src/main/java/resources/omero-model.properties
```
mtbc: Shouldn't have the `java/` in there. (Also below.)

contributing/schema-changes.rst
```diff
@@ -140,5 +144,5 @@ diff` reports should help you to create a new
 :file:`sql/psql/OMERO5.1DEV__5/OMERO5.1DEV__4.sql`.
 
 When you commit your code and issue a pull request, include the
-changes to :omero_source:`omero.properties <etc/omero.properties>` and
+changes to :omero_model_source:`omero-model.properties <src/main/resources/omero-model.properties>` and
```
mtbc, line 147: Outwith this PR, in the longer term I wonder if `build-schema` should just be part of the normal build and release process so it goes in the zip but doesn't have the autogen output in version control.

joshmoore: > but doesn't have the autogen output in version control.

No strong objection. There are times where I've `git grep`'d or `git log`'d on that directory, but that can be worked around.

jburel: build-schema needs to be moved "down" in the stack. As discussed we do not need it for 5.5, but one thing  to look after.

mtbc (approved):

PR #1981, opened 2019-06-05
troubleshooting: Fix OpenSSL version section

manics:
https://github.com/ome/omero-install/pull/197 updated the command, but the surrounding text wasn't updated.


jburel:
Thanks @manics

PR #1983, opened 2019-06-07
Make the scope of in place import more explicit

joshmoore:
The fact that one should not create links on data inside the ManagedRepository because of the the logic used for the database management is noted explicitly.

### DIFF
--- a/omero/sysadmins/in-place-import.rst
+++ b/omero/sysadmins/in-place-import.rst
@@ -4,7 +4,7 @@ In-place import
 ===============
 
 In-place import allows files which are already present on the server machine
-to be imported into OMERO without the need to copy them. This requires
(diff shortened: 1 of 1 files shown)


omero/sysadmins/in-place-import.rst
```diff
@@ -79,6 +79,8 @@ Also, there is still some data duplication when
 :model_doc:`pyramids <omero-pyramid/>` are generated. We are
 hoping to find a work-around for this in the future.
 
+Do not use this procedure to create links on data inside the ManagedRepository.
```
mtbc, line 82: Should be fine for hard links?

joshmoore:
Closing since the act of opening the PR myself, of course doesn't give me the rights to push to @fmeyenhofer's branch. Please see https://github.com/ome/ome-documentation/pull/1986

PR #1991, opened 2019-06-14
changed to walkthrough_ubuntu1804 from walkthrough_debian9

MartinSt80:
Postgresql10 installation fails if "http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main" (debian9) is used.
"http://apt.postgresql.org/pub/repos/apt/ bionic-pgdg main" works
Walkthrough link should be changed?


jburel:
Thanks for spotting that error and for fixing it

PR #1992, opened 2019-06-14
Upgrade check

jburel:
Retrieve url from config service

### DIFF
--- a/omero/sysadmins/UpgradeCheck.rst
+++ b/omero/sysadmins/UpgradeCheck.rst
@@ -59,7 +59,7 @@ Developers
 ----------
 
 To use the UpgradeCheck class from your own
-code, it is necessary to have ``omero-gateway-x.y.z.jar`` on your classpath. Then,
+code, it is necessary to have ``omero-common-x.y.z.jar`` on your classpath. Then,
 
 ::
 
--- a/omero/sysadmins/UpgradeCheck.rst
+++ b/omero/sysadmins/UpgradeCheck.rst
@@ -63,10 +63,11 @@ code, it is necessary to have ``omero-gateway-x.y.z.jar`` on your classpath. The
 
 ::
 
-        String version = Yourclass.class.getPackage().getImplementationVersion();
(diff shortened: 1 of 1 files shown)


joshmoore:
This is because the URL is no longer present in the bundle?

jburel:
It is still available via ``ResourceBundle bundle = ResourceBundle.getBundle("omero-common");``
but it means that user will need to know which properties file the desire properties is.
We could change it for that properties since it does not require to be connected to check.

mtbc:
Could be worth an RFE for a property-getting gateway helper that checks all the files. If this pattern becomes the usual way to fetch configuration then clients will break whenever we move a property from one component to another. Does *not* block this PR though.

jburel:
My preferred option is to use the ``IConfig`` service  and do ``service.getConfigValue("omero.pixeldata.max_plane_width");`` for example
but That requires to have a session.
We probably need a section on "how to access configuration option"

mtbc:
Staged at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/lastSuccessfulBuild/artifact/omero/_build/html/sysadmins/UpgradeCheck.html.

joshmoore:
Few things:

 * Getting one's own version with `getImplementationVersion` is probably still valid and prevents a hard-coding. Perhaps we could show both.
 * I keep going back and forth in my deliberations: reducing remote calls is always good, but using only the bundle means that a sysadmin can't override. (cc: @chris-allan)
 * :+1: for not needing to know the specific properties file (which may ultimately _also_ be an argument for making the remote call)

manics:
Not a blocker: This seems way too low-level for the sysadmin docs section

jburel:
The ``bundle`` approach does not work in Matlab even if the jar is in the classpath i.e. the properties file is not found, the other approach i.e. using ``session`` does work.

joshmoore:
Merging since this is a fix for a bug, but I would still appreciate having `String version = Yourclass.class.getPackage().getImplementationVersion();` restored as more correct than hard-coding a version if we review this code again per @manics' remarks.

PR #1993, opened 2019-06-17
missing space

MartinSt80:
missing space leads to concatenation: apt-get install -y -qbuild-essential...


joshmoore:
Thanks, @MartinSt80! If you're interested, the source for this page is https://github.com/ome/omero-install/blob/b3c8f4c61beef20b280d2ae9fc4a5b92967ef5d8/linux/step01_ubuntu1804_ice_deps.sh

We'll migrate the fix over unless you beat us to it.

jburel:
see https://github.com/ome/omero-install/pull/204

jburel:
The doc being generated from ome/omero-install#204
This change will be overridden 
Thanks @MartinSt80 for pointing that error

PR #1994, opened 2019-06-17
set-up

jburel:
Clarify point raised in https://forum.image.sc/t/problem-when-installing-omero-server-5-5-ubuntu-18-04-server-hyper-v/26695
This also removed a compile line no longer required.
the troubleshooting page was fixed but not the walkthrough
cc @MartinSt80


joshmoore:
Thanks @jburel. Is there a related omero-install PR?

jburel:
not for that change

joshmoore:
I guess due to the need for the download instructions. If we're going to stay with the omero-install scripts we might need a way to review what's in and what's out, e.g. a flag which shows the diff. Visual inspection is definitely tricky at the moment.

PR #1996, opened 2019-06-20
Correct name of PostgreSQL setup script.

mtbc:
```
$ cat /etc/centos-release
CentOS Linux release 7.6.1810 (Core) 
$ ls -l /usr/pgsql-10/bin/postgresql*setup
-rwxr-xr-x. 1 root root 9419 May  8 10:25 /usr/pgsql-10/bin/postgresql-10-setup
$ 
```
Staged at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/lastSuccessfulBuild/artifact/omero/_build/html/sysadmins/unix/server-centos7-ice36.html#installing-prerequisites.


jburel:
This needs to be modified in omero-install see https://github.com/ome/omero-install/blob/develop/linux/step01_centos7_pg_deps.sh#L81

mtbc:
Ah, turns out I hadn't fetched omero-install for ages. Is that instead of or in addition to this PR?

mtbc:
(I thought that kind of issue in the automated scripts would be caught by CI?)

jburel:
instead of this PR, closing this one

PR #2002, opened 2019-07-03
New versions (OMERO 5.5.1)

joshmoore:
https://docs.openmicroscopy.org/omero-blitz/5.5.3/ et al.


joshmoore:
Merging new versions in order to run autogen.

PR #2014, opened 2019-08-13
Adjust DropBox doc for new config and bug warning

pwalczysko:
As a follow-up on the g.doc draft https://docs.google.com/document/d/11LjmXRHkUFAb_-l6QAyB5o7vSmqtGMWUMOipQ9FCVYw/edit#heading=h.rm0gur8h7to7 and as tested on outreach,
adjusting the DropBox doc for three issues

   - warn of the bug https://trello.com/c/7P6s5W6H/435-bug-dropbox-importusers
   - make explicitly clear in the examples that you can use importTargets in the config
   - reformulate inaccurate sentence about the group which is targeted during DropBox import

cc @joshmoore @jburel

### DIFF
--- a/omero/sysadmins/dropbox.rst
+++ b/omero/sysadmins/dropbox.rst
@@ -60,8 +60,8 @@ convenience. Copying or moving a file of an importable file type into a
 named directory or nested subdirectory will initiate an automatic import
 of that file for that user. Multi-file formats will be imported after
 the last required file of a set is copied into the directory. Images and
-plates will be imported into the group the user was last logged into, with
-images placed into ``Orphaned images``.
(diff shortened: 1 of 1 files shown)


joshmoore:
Two quick points:
 - not overly sure we want to embed trello links into the docs
 - it might be best to mark importUsers deprecated/disabled until this is remedied

pwalczysko:
@joshmoore adjusted the text accordingly. Not suer how to really disable the functionality in the code though, if this is what we want to do

joshmoore:
Thanks, @pwalczysko. Merging so it's in the next doc builds, but I updated the trello card so that we get this sorted.

PR #2017, opened 2019-08-30
upgrade: hard-code version requirements for the moment

joshmoore:
Moving forward, the DB version number will need to be made more
prominent in the documentation.

see: https://forum.image.sc/t/question-about-postgres-database-during-upgrade-from-5-4-9-to-5-5-1/29119

### DIFF
--- a/omero/sysadmins/server-upgrade.rst
+++ b/omero/sysadmins/server-upgrade.rst
@@ -211,8 +211,8 @@ Upgrade your database
 .. only:: point_release
 
     .. warning::
-        This section only concerns users upgrading from a |previousversion| or
-        earlier server. If upgrading from a |version| server, you do not need
+        This section only concerns users upgrading from a 5.3 or
+        earlier server. If upgrading from a 5.4 or 5.5 server, you do not need
         to upgrade the database.
 
 Ensure Unicode character encoding
@@ -272,7 +272,7 @@ to run.
     (1 row)
 
 
-If you are upgrading from a server earlier than |previousversion| then
+If you are upgrading from a server earlier than 5.3, then
 you must run the earlier upgrade scripts in sequence before the one
 above. There is no need to download and run the server from an
 intermediate major release but you must still study the upgrade


omero/sysadmins/server-upgrade.rst
```diff
@@ -211,8 +211,8 @@ Upgrade your database
 .. only:: point_release
 
     .. warning::
-        This section only concerns users upgrading from a |previousversion| or
-        earlier server. If upgrading from a |version| server, you do not need
+        This section only concerns users upgrading from a 5.3 or
+        earlier server. If upgrading from a 5.4 or 5.5 server, you do not need
```
mtbc, line 215: Doesn't seem to be included in staged docs, guess not considered a point release.

omero/sysadmins/server-upgrade.rst
```diff
@@ -272,7 +272,7 @@ to run.
     (1 row)
 
 
-If you are upgrading from a server earlier than |previousversion| then
+If you are upgrading from a server earlier than 5.3, then
```
mtbc, line 275: Will be staged at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/lastSuccessfulBuild/artifact/omero/_build/html/sysadmins/server-upgrade.html#run-the-upgrade-script once job is green.

mtbc (approved):
Happy to approve on basis of diff given that the current page is wrong, thank you for this fix.

In the longer term we'll have to figure how decoupling DB from server version affects how this page is generated or we'll just make more mistakes.

omero/sysadmins/server-upgrade.rst
joshmoore: > Doesn't seem to be included in staged docs

Because of the job, right?

> guess not considered a point release.

What do you mean?

joshmoore:
> In the longer term we'll have to figure how decoupling DB from server version...

Exactly. It might be that the DB repository (perhaps one with just the SQL?) gets the instructions on the SQL upgrade, and the server points to that. Not sure.

omero/sysadmins/server-upgrade.rst
mtbc: I think the `.. only:: point_release` omits the warning altogether from staging.

joshmoore: Gotcha. So the question might be how to reproduce the version that's on the website.

joshmoore:
Merging and trying to get this released to head off similar issues for other upgraders.

mtbc:
Update applied by hand to live site after discussion with @sbesson:
```diff
250,251c250,251
< <p class="last">This section only concerns users upgrading from a 5.4 or
< earlier server. If upgrading from a 5.5 server, you do not need
---
> <p class="last">This section only concerns users upgrading from a 5.3 or
> earlier server. If upgrading from a 5.4 or 5.5 server, you do not need
304c304
< <p>If you are upgrading from a server earlier than 5.4 then
---
> <p>If you are upgrading from a server earlier than 5.3 then
````

PR #2019, opened 2019-09-04
exclude zeroc forum

jburel:
* No error when checked locally on mac
 * no error when I run ``curl -Is https://forums.zeroc.com/discussion/6631/ice-3-6-0-and-ice-touch-3-6-0-released | head -1
HTTP/1.1 200 OK`` in the devspace container

--exclude

### DIFF
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -186,7 +186,8 @@ linkcheck_ignore += [r'http://localhost:\d+/?', 'http://localhost/',
     'https://testng.org/',
     # Those below may start working with Sphinx 2.1, see sphinx-doc #6381.
     'https://www.cloudflare.com/',
-    'https://www.zenoss.com/'
+    'https://www.zenoss.com/',
+    'https://forums.zeroc.com/*'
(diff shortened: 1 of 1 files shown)


jburel:
Excluding this PR unless we are happy to merge it
It is conflicting with newer work

mtbc:
This probably relates to the Sphinx PR @sbesson submitted re. newer HTTP response codes.

joshmoore:
I'd just close if you are comfortable that zeroc.com isn't going to start failing. If it is (perhaps test via a re-run this afternoon?), then might as well get it merged.

mtbc:
Do we even test with a Sphinx that includes @sbesson's fix? If so, still probably worth a CI run before to check.

sbesson:
As per https://github.com/sphinx-doc/sphinx/blob/6644830/CHANGES#L82, it looks like my fix was cherry-picked on the `master` branch and is scheduled for the Sphinx `2.3.0` release which is still under development.

PR #2023, opened 2019-10-16
Use webgateway URL names in webgateway page

will-moore:
This updates webgateway URL docs to use the URL name instead of the "webgateway.views.method" path to the callable method. 

```reverse("webgateway.views.full_viewer", args=(iid,)))``` no longer works (see https://github.com/ome/omero-webtest/pull/29) and usage of the URL name will be required in future versions of Django.

However, the examples changed in this PR will not actually be broken by
htts://github.com/ome/omero-web/commit/2fff52d8dabfb0087493a1f2dc7068594912d1c4
(and other commits in that PR) since they use ```{% url 'webgateway.views.render_image' id %}``` in the html template and this still works.

### DIFF
--- a/omero/developers/Web/WebGateway.rst
+++ b/omero/developers/Web/WebGateway.rst
@@ -51,7 +51,7 @@ Image viewer
     
         URL: https://your_host/webgateway/img_detail/<imageid>/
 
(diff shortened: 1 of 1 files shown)


joshmoore:
Certainly matches the recent code changes. :+1:

PR #2030, opened 2019-11-19
scripts section

jburel:
make code py3 compatible

see https://github.com/ome/ome-documentation/issues/2028
cc @will-moore


will-moore:
Looks good. 👍

PR #2038, opened 2019-11-28
Install

jburel:
Review walkthroughs
related to https://github.com/ome/omero-install/pull/216

### DIFF
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -70,7 +70,7 @@ rst_epilog += """
 .. |previous_dbver|  replace:: %s
 .. |iceversion| replace:: 3.6.5
 .. |postgresversion| replace:: 10
-.. |javaversion| replace:: 1.8
+.. |javaversion| replace:: 11
 
 .. |Broken| image:: /images/broken.png
              :alt: Broken
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -69,7 +69,7 @@ rst_epilog += """
 .. |current_dbver|  replace:: %s
 .. |previous_dbver|  replace:: %s
 .. |iceversion| replace:: 3.6.5
-.. |postgresversion| replace:: 10
+.. |postgresversion| replace:: 11
 .. |javaversion| replace:: 11
 
 .. |Broken| image:: /images/broken.png
--- a/omero/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos7-ice3.6.rst
+++ b/omero/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos7-ice3.6.rst
@@ -26,9 +26,8 @@ Installing prerequisites
 **The following steps are run as root.**
 
 
-Install dependencies. The number of dependencies to install depends on the way you plan to install OMERO.web. If you wish to install it in a virtual environment with ``--system-site-packages`` *off*, a few more dependencies will be required (**option 2**)::
+Install dependencies::
 
-    # dependencies common to both options
     yum -y install epel-release
 
     yum -y install unzip
@@ -37,30 +36,20 @@ Install dependencies. The number of dependencies to install depends on the way y
 
     yum -y install nginx
 
-    # To install OMERO.web using option 2
-    yum -y install python3-devel libjpeg-devel libtiff libtiff-devel zlib-devel
-
-
-
 
 Creating a virtual environment
 ------------------------------
 
 **The following steps are run as root.**
 
-Create the virtual environment. This is the preferred way to install OMERO.web::
+Create the virtual. This is the recommended way to install OMERO.web::
 
-    # option 1: in a virtual environment with --system-site-packages on
-    python3 -mvenv /opt/omero/web/venv --system-site-packages
-
-    # option 2: in a virtual environment with --system-site-packages off
     python3 -mvenv /opt/omero/web/venv
 
 
 
 Install ZeroC IcePy 3.6::
 
-    # option 1 and option 2
     /opt/omero/web/venv/bin/pip install --upgrade https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.2.1/zeroc_ice-3.6.5-cp36-cp36m-linux_x86_64.whl
 
 
--- a/omero/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos7-ice3.6.rst
+++ b/omero/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos7-ice3.6.rst
@@ -1,13 +1,13 @@
 .. walkthroughs are generated using ansible, see 
 .. https://github.com/ome/omeroweb-install
 
-OMERO.web installation separately from OMERO.server on CentOS 7 and IcePy 3.6
-=============================================================================
+OMERO.web installation on CentOS 7 and IcePy 3.6
+================================================
 
 Please first read :doc:`../../server-centos7-ice36`.
 
 
-This is an example walkthrough for installing OMERO.web decoupled from the OMERO.server in a **virtual environment** using a dedicated system user. Installing OMERO.web in a virtual environment is the preferred way. For convenience in this walkthrough, we will use the **omero system user** and define the main OMERO.web configuration options as environment variables.
+This is an example walkthrough for installing OMERO.web in a **virtual environment** using a dedicated system user. Installing OMERO.web in a virtual environment is the preferred way. For convenience in this walkthrough, we will use the **omero system user** and define the main OMERO.web configuration options as environment variables.
 
 
 **The following steps are run as root.**
--- a/omero/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos7-ice3.6.rst
+++ b/omero/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos7-ice3.6.rst
@@ -8,6 +8,7 @@ Please first read :doc:`../../server-centos7-ice36`.
 
 
 This is an example walkthrough for installing OMERO.web decoupled from the OMERO.server in a **virtual environment** using a dedicated system user. Installing OMERO.web in a virtual environment is the preferred way. For convenience in this walkthrough, we will use the **omero system user** and define the main OMERO.web configuration options as environment variables.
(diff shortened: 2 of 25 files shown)


pwalczysko:
The paragraph https://github.com/ome/ome-documentation/pull/2038/files#diff-4f27337b81297518abca1701b964d086R13 - I think it belongs one level up , to the general server installations overview.  Once I have started to read CentOS7 specifically, I am already commited to a version. Maybe it wants to say something about combination of versions, and really is specific for CentOS7, but then, this is unclear.

pwalczysko:
The sentence https://github.com/ome/ome-documentation/pull/2038/files#diff-4f27337b81297518abca1701b964d086R18 - a little bit confused. So this implicitly suggests that OMERO.web can be also installed non-separately. Or does it ? Reading that, I just do not know, if there are two options

  - install OMERO.web separately
  - install OMERO.web together with OMERO.server

or just one

  - install OMERO.web separately

If the former (2 options), it is unclear from the sentence where to go to install OMERO.web together with OMERO.server.

joshmoore:
@jburel: can I confirm that these changes are coming from https://github.com/ome/omero-install/pull/216 and that the two are currently synced at:

 - https://github.com/ome/ome-documentation/pull/2038/commits/7b387830d3bf9466595c7fb67db5042b1302df61
 - https://github.com/ome/omero-install/pull/216/commits/5e4233d285d91d25af3f47d6c661d41f0640d126

?

jburel:
They should be in synch

pwalczysko:
This comment https://github.com/ome/ome-documentation/pull/2038/files#diff-f0918414fab9e2e9aec7d5e57110e9ebR9 reads weird - I understand that it comes "automatically" from the sh script, but the end result is a mild confusion for me - what has been installed for convenience ? Simple order like "now do that" would be more reassuring.

pwalczysko:
some postgres problem reported in https://github.com/ome/omero-install/pull/216#issuecomment-559807248

pwalczysko:
I think I am blocked by the problem https://github.com/ome/ome-documentation/pull/2038#issuecomment-559807358 - the next command cannot find the directory 

```
[root@b9e774350374 /]# sed -i.bak -re 's/^(host.*)ident/\1md5/' /var/lib/pgsql/10/data/pg_hba.conf
sed: can't read /var/lib/pgsql/10/data/pg_hba.conf: No such file or directory
```

pwalczysko:
Yes, I am blocked. Even as further commands after the ``sed`` work, I think the postgres install is not right and the problem comes

```
[root@b9e774350374 /]# echo "CREATE USER $OMERO_DB_USER PASSWORD '$OMERO_DB_PASS'" | su - postgres -c psql
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
[root@b9e774350374 /]# 
```

jburel:
humm I have not modified anything on the pg front.

pwalczysko:
@jburel here https://linuxacademy.com/community/show/21629-docker-failed-to-get-dbus-connection-operation-not-permitted/ it is suggested that this is a Docker + CentOS7 problem

jburel:
the problem is that in docker things are a bit different. The "docker" commands are not shown in the doc. Maybe we should 
```
     if [ "${container:-}" = docker ]; then
        su - postgres -c "/usr/pgsql-10/bin/initdb -D /var/lib/pgsql/10/data --encoding=UTF8"
        echo "listen_addresses='*'" >> /var/lib/pgsql/10/data/postgresql.conf
    else
        PGSETUP_INITDB_OPTIONS=--encoding=UTF8 /usr/pgsql-10/bin/postgresql-10-setup initdb
    fi
    sed -i.bak -re 's/^(host.*)ident/\1md5/' /var/lib/pgsql/10/data/pg_hba.conf
    if [ "${container:-}" = docker ]; then
        sed -i 's/OOMScoreAdjust/#OOMScoreAdjust/' \
        /usr/lib/systemd/system/postgresql-10.service
    fi
    if [ "${container:-}" = docker ]; then
        su - postgres -c "/usr/pgsql-10/bin/pg_ctl start -D /var/lib/pgsql/10/data -w"
    else
        systemctl start postgresql-10.service
    fi

```

pwalczysko:
```
(venv) [omero@b9e774350374 OMERO.server]$ echo $PATH
/opt/omero/server/venv/bin:/opt/omero/server/venv/bin:/opt/omero/server/venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
(venv) [omero@b9e774350374 OMERO.server]$ cd ..
(venv) [omero@b9e774350374 ~]$ omero config set
bash: omero: command not found
```

cannot find ``omero``

jburel:
Sorry there is a section missing in the doc :-)

jburel:
@pwalczysko fixed

pwalczysko:
Screen Shot 2019-12-02 at 15 25 31

Postgres 10 is still hightlighted here. I cannot find this part in the code here though.

jburel:
> 
> Postgres 10 is still hightlighted here. I cannot find this part in the code here though.

fixed

pwalczysko:
I successfully installed and started the OMERO.server on Centos7 now.
Also, all the walkthrough seems sensible and the fixes were done.

But, trying OMERO.web installation just after that, I have problems

```
[omero@4a0db055e33c root]$ omero web config nginx --http "${WEBPORT}" --servername "${WEBSERVER_NAME}" > /home/omero/nginx.conf.tmp
usage: /opt/omero/server/venv/bin/omero [-h] [-v] [-d DEBUG] [--path PATH]
                                        [-C] [-s SERVER] [-p PORT] [-g GROUP]
                                        [-u USER] [-w PASSWORD] [-k KEY]
                                        [--sudo ADMINUSER] [-q]
                                        <subcommand> ...
/opt/omero/server/venv/bin/omero: error: argument <subcommand>: invalid choice: 'web'

choose from:
	admin, chgrp, chown, config, db, delete, download, errors, export, 
	fs, group, help, hql, import, ldap, load, login, logout, node, 
	obj, perf, quit, script, search, server, sessions, shell, tag, 
	testengine, upload, user, version
```

Screen Shot 2019-12-02 at 16 21 27

pwalczysko:
Maybe the problems above https://github.com/ome/ome-documentation/pull/2038#issuecomment-560468033 are caused by the venv not having been activated ? But there was no activation step afaik

Screen Shot 2019-12-02 at 16 24 00

pwalczysko:
Consulting the workflow with @manics , indeed, I should have activated the venv for OMERO.web. But I did not because the doc did not tell me so, see screenshot at https://github.com/ome/ome-documentation/pull/2038#issuecomment-560469624 - this should be probably fixed.

Edit: Maybe it is not the venv, but rather the omero and omero web cmds defs, see below.

pwalczysko:
```
[omero@4a0db055e33c root]$ python3 -mvenv  /opt/omero/web/venv/ 
Error: [Errno 13] Permission denied: '/opt/omero/web/venv/pyvenv.cfg'
[omero@4a0db055e33c root]$ 
```
I tried to activate the web venv - but see above - now I am really stuck

pwalczysko:
I think the problems above come from the fact that my "omero" command is defined for the server only, and not at all for the web. I was able to circumvent the problem above by starting a new shell and then not defining the server environment (or rather, run the "server" omero commands first, then start the new shell). Then, the "omero web" command would still not be recognized, except that I went into the OMERO.server dir and run it from there, this worked.

It seems that at least the missing "omero web" definition should be dealt with in the doc - or some suggestion/guidance about how to deal with the alternating "omero" and "omero web" commands.

pwalczysko:
https://github.com/ome/omero-install/blob/develop/linux/step05_centos7_nginx.sh#L31

This line I ran  - and although I am in docker, it still ended with 

```
[root@4a0db055e33c omero]# systemctl start nginx
Failed to get D-Bus connection: Operation not permitted
```

jburel:
@pwalczysko Web is installed in another virtualenv (the name is different). There is another page for that.
Maybe it is b/c we use the same user that is is confusing.
Let's discuss tomorrow

jburel:
I have made few changes to the omero.web install pages.
Hopefully it will be clearer

jburel:
@pwalczysko I have pushed an extra commit this am so it is not included in the build to clarify ``OMERODIR``

pwalczysko:
Yes, it is clearer, but https://github.com/ome/ome-documentation/pull/2038#issuecomment-560516192 problem with nginx still persists

jburel:
@pwalczysko could you start docker with ``--privileged`` 
i.e. ``docker run --privileged .... ``

will-moore:
As discussed, in page title "OMERO.web installation separately from OMERO.server on CentOS 7 and IcePy 3.6" we could probably remove "separately from OMERO.server" since there is no need to distinguish now. 
Same for text "This is an example walkthrough for installing OMERO.web decoupled from the OMERO.server in a virtual environment", can remove "decoupled from the OMERO.server".

pwalczysko:
>@pwalczysko could you start docker with --privileged
i.e. docker run --privileged .... 

Unfortunately, that did not work either. Same error.

jburel:
Ok the setup will have to be a bit more complicated to test.
In one terminal:
* ``docker run --privileged -it -v /sys/fs/cgroup:/sys/fs/cgroup centos:7 /usr/sbin/init``
* wait a bit You should have something like ``Started Update UTMP about System Runlevel Changes.``
* in another terminal, run ``docker ps``
then ``docker exec -it CONTAINER_ID /bin/bash``
* then follow the instruction you should not have the error now

pwalczysko:
@jburel Thanks. All works now, also the "options" are gone, which is good.

Run the walkthrough on the web install up to the chapter ``Running OMERO.web``
I checked that the omero web is running with the suggested ``curl`` command, and it went well.
Stopping of omero web worked too.

Ready to merge fmpov

joshmoore:
Merging based on the general level of satisfaction from @pwalczysko and @will-moore. Also good to have everything so others can start opening more targeted, polishing PRs. Thanks all!

PR #2040, opened 2019-12-02
review installation instructions on mac

jburel:
* create 2 venv for web and server
cc @will-moore

### DIFF
--- a/omero/sysadmins/unix/server-install-homebrew.rst
+++ b/omero/sysadmins/unix/server-install-homebrew.rst
@@ -100,53 +100,56 @@ Requirements
 
 4. Install Python provided by Homebrew::
 
-    $ brew install python2
+    $ brew install python
 
    Homebrew installs Python in the following location::
 
-    '/usr/local/opt/python/libexec/bin'
+    '/usr/local/opt/python3/libexec/bin'
 
    Follow the instructions from the brew Python install and set your system to use the Homebrew version of Python 
    rather than the Python shipped with OS X. Add the following line to your :file:`.bash_profile`::
 
-    export PATH="/usr/local/opt/python/libexec/bin:$PATH"
+    export PATH="/usr/local/opt/python3/libexec/bin:$PATH"
 
-5. Check that Python is working and is version 2.7.x::
+5. Check that Python is working and is version 3.7.x::
 
-    $ which python
-    /usr/local/opt/python/libexec/bin/python
+    $ which python3
+    /usr/local/opt/python3/libexec/bin/python
 
     $ python --version
-    Python 2.7.13
+    Python 3.7.4
 
 6. For developing with OMERO, or Python in general, we recommend the use of Virtualenv.
    Virtualenv allows development of Python applications without having to
    worry about clashing third-party packages for different Python projects.
 
-   Use pip to get `Virtualenv <https://virtualenv.pypa.io/en/stable/>`__::
+   Create a virtual environment for omero-py using Python 3::
 
-    $ pip install virtualenv
+    $ python3 -mvenv ~/Virtual/omero-server
 
-   With Virtualenv installed, create a virtual environment::
+   This will create a folder to hold Python libraries in the the directory :file:`~/Virtual/omero-server/lib`
 
-    $ virtualenv ~/Virtual/omero
+   Create a virtual environment for omero-web using Python 3::
 
-   This will create a folder to hold Python libraries in the the directory :file:`~/Virtual/omero/lib`
+    $ python3 -mvenv ~/Virtual/omero-web
+
+   This will create a folder to hold Python libraries in the the directory :file:`~/Virtual/omero-web/lib`
 
   .. note:: 
    You can activate the Virtualenv environment that we created using::
 
-    $ source ~/Virtual/omero/bin/activate
+    $ source ~/Virtual/omero-server/bin/activate
 
    This will switch to using Pip and Python in the Virtualenv directory
-   :file:`~/Virtual/omero/bin` and any Pip libraries you install, whilst the Virtualenv is activated, 
-   will be installed to :file:`source ~/Virtual/omero/lib`.
+   :file:`~/Virtual/omero-server/bin` and any Pip libraries you install, whilst the Virtualenv is activated, 
+   will be installed to :file:`source ~/Virtual/omero-server/lib`.
 
   .. note::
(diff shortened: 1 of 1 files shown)


omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -100,61 +100,65 @@ Requirements
 
 4. Install Python provided by Homebrew::
 
-    $ brew install python2
+    $ brew install python
 
    Homebrew installs Python in the following location::
 
-    '/usr/local/opt/python/libexec/bin'
+    '/usr/local/opt/python3/libexec/bin'
```
will-moore: I got this on Mojave (10.14.6):
```
Python has been installed as
  /usr/local/bin/python3
```

will-moore: But actually that failed because homebrew couldn't link. ```brew link python``` failed...

omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -100,61 +100,65 @@ Requirements
...
 
    Follow the instructions from the brew Python install and set your system to use the Homebrew version of Python 
    rather than the Python shipped with OS X. Add the following line to your :file:`.bash_profile`::
 
-    export PATH="/usr/local/opt/python/libexec/bin:$PATH"
+    export PATH="/usr/local/opt/python3/libexec/bin:$PATH"
```
will-moore: My python3 is in ```/usr/local/opt/python/libexec/bin```

omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -100,61 +100,65 @@ Requirements
...
 
-5. Check that Python is working and is version 2.7.x::
+5. Check that Python is working and is version 3.7.x::
 
-    $ which python
-    /usr/local/opt/python/libexec/bin/python
+    $ which python3
+    /usr/local/opt/python3/libexec/bin/python
```
will-moore: As above, when I added ```export PATH="/usr/local/opt/python/libexec/bin:$PATH"```
I get:
```
$ which python
    /usr/local/opt/python/libexec/bin/python
$ python
Python 3.7.4...
```

omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -100,61 +100,65 @@ Requirements
...
    worry about clashing third-party packages for different Python projects.
 
-   Use pip to get `Virtualenv <https://virtualenv.pypa.io/en/stable/>`__::
+   Create a virtual environment for omero-py using Python 3::
 
-    $ pip install virtualenv
+    $ python3 -mvenv ~/Virtual/omero-server
 
-   With Virtualenv installed, create a virtual environment::
+   This will create a folder to hold Python libraries in the the directory :file:`~/Virtual/omero-server/lib`
```
will-moore: the the

omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -100,61 +100,65 @@ Requirements
...
 
-    $ virtualenv ~/Virtual/omero
+   Create a virtual environment for omero-web using Python 3::
 
-   This will create a folder to hold Python libraries in the the directory :file:`~/Virtual/omero/lib`
+    $ python3 -mvenv ~/Virtual/omero-web
+
+   This will create a folder to hold Python libraries in the the directory :file:`~/Virtual/omero-web/lib`
```
will-moore: the the

omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -207,40 +211,46 @@ OMERO configuration
...
+    export OMERODIR=/path/to/OMERO.server-x.x.x-ice36-bxx
     # ...OR locally built server
-    export OMERO_SERVER=/path/to/openmicroscopy/dist
+    export OMERODIR=/path/to/openmicroscopy/dist
 
    and add the OMERO executable to the OS X :envvar:`PATH`::
 
     # Add the OMERO distribution to PATH
-    export PATH=$OMERO_SERVER/bin:$PATH
+    export PATH=$OMERODIR/bin:$PATH
```
will-moore: Do we need to do this now, since in the virtualenv we will have ```omero``` available already?

omero/sysadmins/unix/server-install-homebrew.rst
jburel: ok mine was under python3.

omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -100,61 +100,65 @@ Requirements
...
 
    Using the command-line terminal, reload your :file:`.bash_profile`::
 
     $ source ~/.bash_profile
 
    Now you can activate the Virtualenv environment using::
 
-    $ startVmOmero
+    $ startVmOmeroServer
+    $ startVmOmeroWeb
```
will-moore, line 161: It might be nice to use these commands instead of the longer ```source...``` commands everywhere they're used below (or add the alias alongside the source command) so that by the time you are done, you've learnt them. Maybe the "(Optional)" should be "(Recommended)". I've certainly found it useful to have these aliases.

joshmoore: My preference would be to _not_ have the aliases _anywhere_. If the path to the directory is too long/difficult to remember, I'd suggest we find a location that is more user friendly. `source` (and `.` even more so) is a fundamental shell action (like `ls`, `cd`, and `mv`). Depending on the alias means that a user in another situation may not have learned "source" which would ultimately be more useful.

jburel: the aliases have been there for a while. Happy to remove them 
Is it a blocker?

will-moore:
Looks good, thanks.
I realise that on my machine (not the one I was testing on yesterday) I do need `python3`:

```
$ which python
/Users/willadmin/anaconda2/bin/python
$ which python3
/usr/local/bin/python3
```
So my comment above is not always valid.

omero/sysadmins/unix/server-install-homebrew.rst
joshmoore: Not a blocker (unless someone is relying on them in production?) and not a must from my side. Just spelling out my preference how I think about the doc design.

PR #2043, opened 2019-12-04
Tables

jburel:
* remove page and update links


omero/sysadmins/server-tables.html
```diff
@@ -0,0 +1,8 @@
+<html>
+  <head>
+    <meta http-equiv="refresh" content="1; url=../" />
+    <script>
+      window.location.href = ".."
```
joshmoore: Building now to test, but in my previous redirect I had moved the content from the given file to the index which is why I used `..`. If there's a better location for users who are looking for the old `server-tables` content, feel free to redirect them there.

jburel: I will point to dev page

joshmoore: NB: I'll register this file in my PR since otherwise it will conflict. In future PRs, each should add the file itself: https://github.com/ome/ome-documentation/pull/2042/files#diff-55a1d28abbfc5d72410cace65452f0ebR212

joshmoore:
Tested with 20191204.3 build. Looks good. :+1:

PR #2044, opened 2019-12-04
Install server

jburel:
Change the location of the OMERO.server
This is no longer installed under the home directory of the omero user

related to https://github.com/ome/omero-install/pull/218


will-moore:
Works fine.

Just an idea: what about we have a 'test' after start, equivalent to the 'curl' test on the web-install docs. e.g.

```
omero admin start

# login with the $OMERO_ROOT_PASS password
omero login
Previous session expired for root on localhost:4064
Server: [localhost:4064]
Username: [root] 
Password:                          # e.g. omero_root_password
Created session for root@localhost:4064. Idle timeout: 10 min. Current group: system
```

jburel:
@will-moore something we could add I would prefer to do it in a follow up PR if we want to have that as part of omero-install

PR #2050, opened 2019-12-05
autogen: install pillow

joshmoore:
Prevents error messages in generated output.


joshmoore:
Green in travis and multiple builds of OMERO-docs-superbuild

PR #2065, opened 2019-12-17
Web cron

jburel:
* Manually update the file to match the new configuration
* Add link back to doc
* Change the location of the file otherwise it will be removed by autogen


joshmoore:
Thanks, @jburel . https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-docs/ws/omero/_build/html/sysadmins/unix/install-web.html looks good and considering how esoteric cron might be, I think having an example is worthwhile.

PR #2089, opened 2020-01-10
Changelog setup

jburel:
This PR updates the file holding info about change logs
This file and the history are then combined by ``autogen_docs`` to create the history in the doc

cc @joshmoore @will-moore


joshmoore:
LGTM

PR #2090, opened 2020-01-10
Install dropbox

jburel:
install dropbox from pypi

### DIFF
--- a/omero/sysadmins/dropbox.rst
+++ b/omero/sysadmins/dropbox.rst
@@ -37,6 +37,29 @@ the following more specific requirements:
     operating system. Watching a network-attached share (NAS) is strictly
     ***not*** supported.
 
+Installing DropBox
+------------------
+
+From OMERO 5.6.0 release, the library ``omero-dropbox`` supports Python 3 and
+is now available on PyPI_. We recommend you use a Python virtual environment to install it.
+
+To install ``omero-dropbox`` using venv:
+
+.. parsed-literal::
+
+    $ python3 -m venv myenv
+    $ . myenv/bin/activate
+    $ pip install omero-dropbox==\ |version_py|
+    Successfully installed Pillow-6.2.1 future-0.18.2 omero-py-5.6.1 zeroc-ice-3.6.5
+
+To install ``omero-dropbox`` using conda (preferred):
+
+.. parsed-literal::
+
+    conda create -n myenv -c ome python=3.6 zeroc-ice36-python
+    conda activate myenv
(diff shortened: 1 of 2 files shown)


omero/sysadmins/dropbox.rst
```diff
@@ -37,6 +37,29 @@ the following more specific requirements:
     operating system. Watching a network-attached share (NAS) is strictly
     ***not*** supported.
 
+Installing DropBox
+------------------
+
+From OMERO 5.6.0 release, the library ``omero-dropbox`` supports Python 3 and
```
mtbc:
```suggestion
From the OMERO 5.6.0 release, the library ``omero-dropbox`` supports Python 3 and
```

dominikl:
Trying to follow the doc... first thing I asked myself is, why do I have to create an new venv? And for which user? I guess must be as root using the `/opt/omero/server/venv3` so I installed it with `/opt/omero/server/venv3/bin/pip install omero-dropbox==5.6.dev3` (Typo in the doc there, there's no dev10, only dev3).

Enabled dropbox via (as omero-server user):
```
$ omero admin ice server enable MonitorServer
$ omero admin ice server enable DropBox
```
This should actually be part of the section "Installing DropBox", not only "Permissions".

 Then got a lot of errors like this in the Dropbox.log (testing on the debian10 VM again):
```
2020-01-13 12:47:50,208 INFO  [                        fsclient.DropBox] (MainThread) Monitor parameters = {'default': {'watchDir': '', 'eventTypes': [Creation, Modification], 'pathMode': Follow, 'whitelist': [], 'blacklist': [], 'timeout': 0.0, 'blockSize': 0, 'ignoreSysFiles': True, 'ignoreDirEvents': True, 'dirImportWait': 60, 'fileWait': 15.0, 'throttleImport': 10, 'timeToLive': 0, 'timeToIdle': 600000, 'fileBatch': 10, 'readers': '', 'importArgs': ''}}
2020-01-13 12:47:50,220 ERROR [                        fsclient.DropBox] (MainThread) Failed to use a query service :
Traceback (most recent call last):
  File "/opt/omero/server/OMERO.server/lib/python/fsDropBox.py", line 127, in run
    watchDir = pathModule.path(
AttributeError: type object 'path' has no attribute 'path'
```
Uninstalled it again. Wanted to try conda, but same question here, don't know which user and where am I supposed to create this new venv?

dominikl:
Thanks, much clearer now 👍

jburel:
Merging

PR #2092, opened 2020-01-10
mac install conda

will-moore:
Include conda install of omero-py in mac install page

cc @jburel 
NB: haven't built locally yet (install problems), so might need some fixes.


will-moore:
Looks good.
I haven't tried an end-to-end walk-though, but I think it make sense.

jburel:
merging for release

PR #2100, opened 2020-01-28
Re-use virtualenv

joshmoore:
It's not possible to encode in the script which zeroc-ice
wheel to install, therefore re-use an existing virtualenv
and warn the user that manual installations may be needed.


joshmoore:
Merging for inclusion in (the building of) 5.6.1.

PR #2102, opened 2020-02-20
review tests section

jburel:
We no longer run any unit-tests 
This removes the outdated commands

Review the way single test or group of tests can be run.

related to https://github.com/ome/openmicroscopy/pull/6218

### DIFF
--- a/omero/developers/testing.rst
+++ b/omero/developers/testing.rst
@@ -25,38 +25,24 @@ the Java tests.
 Running tests
 -------------
 
-
 Running unit tests
 ^^^^^^^^^^^^^^^^^^
 
-The unit testing framework is fairly simple. Only methods which contain
-logic written within OMERO are tested. This means that framework
-functionality like remoting or the Hibernate layer is not tested. This is a
-part of integration testing (see below).
+Starting from version 5.5, components have been migrated to their own repository.
 
-You can run the unit tests for any component from its directory by
-entering:
+The following repositories use `Gradle <https://gradle.org/>`_ to run the unit tests:
+  - :omero_subs_github_repo_root:`omero-model`
+  - :omero_subs_github_repo_root:`omero-common`
+  - :omero_subs_github_repo_root:`omero-romio`
+  - :omero_subs_github_repo_root:`omero-renderer`
+  - :omero_subs_github_repo_root:`omero-server`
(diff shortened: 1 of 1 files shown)


dominikl:
Good to merge from my point of view 👍

omero/developers/testing.rst
```diff
@@ -113,17 +99,10 @@ explicitly via:
...
 Individual tests
 """"""""""""""""
 
-Alternatively, you can run individual tests which you may currently be
-working on. This can be done using the ``test`` target. For example:
-
-::
-
-    ./build.py -f components/tools/OmeroJava/build.xml test -DTEST=integration/chown/PermissionsTest
-    ./build.py -f components/tools/OmeroPy/build.xml test -DTEST=test/integration/test_admin.py
```
joshmoore, line 125: Did this information (integration tests for Python) get moved anywhere? I might be missing something.

jburel: This is done via pytest. 
I had to make a no-op for ``test`` target since it was broken for unit-test.
if user wants to run specific tests they go to the corresponding component and use either ``gradle`` or ``pytest``

joshmoore: Sure, I just mean in terms of the docs update.

jburel: For Java, the same example is under the Java  ``Individual test``
For Python, the example is there but admittedly under the ``-s`` option
I will add a section ``Individual test`` under python to make it clearer that the info is not lost

jburel: done.
Actually @will-moore update was not correct since the example was listed under ``OmeroWeb`` not ``OmeroPy``

dominikl:
Looks good. The example tests (test_annotate and test_admin) fail, but I could run them. Was just wondering, we say for OMERO web tests you need `pip install pytest-django`. But you also need `omero-web`, or is that obvious anyway?

jburel:
I was only looking at the unit tests and Java section but I will check the web part

jburel:
@dominikl 
it is there
``including ``omero-py`` and ``omero-web`` if you want to run OmeroWeb tests.``

dominikl:
You are right. I was just looking at the section
```
# for Omeroweb tests
$ pip install pytest-django
```
Didn't look further up. That's alright then 👍

PR #2104, opened 2020-03-03
List preferrred conda installation first

dominikl:
If conda is the preferred way to install, it should be mentioned first.

See also: https://github.com/ome/omero-py/pull/190


jburel:
The order makes sense

PR #2105, opened 2020-03-17
remove max_requests

jburel:
See https://github.com/ome/omeroweb-install/pull/57

cc @manics @joshmoore


joshmoore:
Understood. I'd cc: @knabar and @chris-allan as well, but otherwise: :+1:

chris-allan:
Sounds good to me.

manics (approved):

PR #2108, opened 2020-03-19
Server upgrade

jburel:
Add section indicating to upgrade omero.py

See https://github.com/ome/ome-documentation/issues/2106


mtbc:
rendered

jburel:
@mtbc did you take a screenshot of the correct page? It does not look like it

mtbc:
Ah, true enough, it's the only place I could find the text, but now I see that OMERO-docs-superbuild-build still says rc1c so maybe it didn't get built at all?

PR #2114, opened 2020-03-30
Suggest installing omero-certificates for SSL trouble.

mtbc:
Staged at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/lastSuccessfulBuild/artifact/omero/_build/html/sysadmins/troubleshooting.html#ssl-connection-issues. See https://forum.image.sc/t/30704 and https://forum.image.sc/t/33939 for background. Could say more once we know more but at least this gets some note already into our docs that might help people searching for that error.

### DIFF
--- a/omero/sysadmins/troubleshooting.rst
+++ b/omero/sysadmins/troubleshooting.rst
@@ -461,6 +461,16 @@ The parameter ``@SECLEVEL=0`` must be added to the server SSL configuration:
 
 Restart the OMERO.server as normal for the changes to take effect.
 
+IceSSL connection failure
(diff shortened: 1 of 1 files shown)


joshmoore:
All for having this also linked from under troubleshooting, but I do wonder if it shouldn't be reachable from the standard install & upgrade paths.

manics:
image

What do you think about combining the two sections and making omero-certificates the recommended suggestion?

manics (approved):
Needs a final check after rebuilding but LGTM

manics:
Good to merge

PR #2124, opened 2020-05-01
Centos8

jburel:
Add support for centos8
See https://github.com/ome/omeroweb-install/pull/60 and https://github.com/ome/omero-install/pull/239

### DIFF
--- a/omero/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos8-ice3.6.rst
+++ b/omero/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos8-ice3.6.rst
@@ -1,0 +1,198 @@
+.. walkthroughs are generated using ansible, see 
+.. https://github.com/ome/omeroweb-install
+
+OMERO.web installation on CentOS 8 and IcePy 3.6
+================================================
+
+Please first read :doc:`../../server-centos8-ice36`.
+
+
+This is an example walkthrough for installing OMERO.web in a **virtual environment** using a dedicated system user. Installing OMERO.web in a virtual environment is the preferred way. For convenience in this walkthrough, we will use the **omero-web system user** and define the main OMERO.web configuration options as environment variables. Since 5.6, a new :envvar:`OMERODIR` variable is used, you should first unset :envvar:`OMERO_HOME` (if set) before beginning the installation process.
+
+
+**The following steps are run as root.**
+
+If required, first create a local system user omero-web and create directory::
+
+    useradd -m omero-web
+
+    mkdir -p /opt/omero/web/omero-web/etc/grid
+    chown -R omero-web /opt/omero/web/omero-web
+
+
+
+Installing prerequisites
+------------------------
+
+**The following steps are run as root.**
+
+
+Install dependencies::
+
+    yum -y install epel-release
+
+    yum -y install unzip
+
+    yum -y install python3
+
+    yum -y install nginx
+
+
+Creating a virtual environment
+------------------------------
+
+**The following steps are run as root.**
+
+Create the virtual environment. This is the recommended way to install OMERO.web::
+
+    python3 -mvenv /opt/omero/web/venv3
+
+
+
+Install ZeroC IcePy 3.6::
+
+    /opt/omero/web/venv3/bin/pip install --upgrade https://github.com/ome/zeroc-ice-centos8/releases/download/0.0.1/zeroc_ice-3.6.5-cp36-cp36m-linux_x86_64.whl
+
+
+
+Install OMERO.web::
+
+    /opt/omero/web/venv3/bin/pip install "omero-web>=5.6.1"
+
+Installing OMERO.web apps
+-------------------------
+
+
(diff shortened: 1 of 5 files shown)


dominikl:
Struggling to get the omero-server running...

Minor issue at step *Install Postgresql 11*
```
[root@localhost tmp]# dnf module disable -y postgresql
error: db5 error(5) from dbenv->open: Input/output error
error: cannot open Packages index using db5 - Input/output error (5)
error: cannot open Packages database in /var/lib/rpm
Error: Error: rpmdb open failed
```
As there's a lot of installation with yum going on before this step, I just restarted the VM and tried again, then `dnf ...` command worked fine.

Further installation steps and omero-server setup works fine. But then trying to start the server:
```
[omero-server@localhost server]$ omero admin start
Traceback (most recent call last):
  File "/opt/omero/server/venv3/bin/omero", line 118, in <module>
    rv = omero.cli.argv()
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1754, in argv
    cli.invoke(args[1:])
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1187, in invoke
    stop = self.onecmd(line, previous_args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1264, in onecmd
    self.execute(line, previous_args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1346, in execute
    args.func(args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/plugins/prefs.py", line 79, in open_and_close_config
    return func(*args, **kwargs)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/plugins/admin.py", line 810, in start
    self.startasync(args, config)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/plugins/prefs.py", line 79, in open_and_close_config
    return func(*args, **kwargs)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/plugins/admin.py", line 761, in startasync
    self.checkice()
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/plugins/admin.py", line 1684, in checkice
    popen = self.ctx.popen(["icegridnode", "--version"])
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1445, in popen
    stdout=stdout, stderr=stderr)
  File "/usr/lib64/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/usr/lib64/python3.6/subprocess.py", line 1364, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'icegridnode': 'icegridnode'
```

Carried on, setting up the systemd service.
```
[root@localhost server]# update-rc.d -f omero-server remove
-bash: update-rc.d: command not found
```

Thought maybe the usual `systemctl ...` would work instead
```
[root@localhost server]# systemctl enable omero-server
omero-server.service is not a native service, redirecting to systemd-sysv-install.
Executing: /usr/lib/systemd/systemd-sysv-install enable omero-server
service omero-server does not support chkconfig
```

I'll try the again with the install bash script, maybe there's a difference between that and the generated doc.

jburel:
Ok,
* so what I had to do in docker to get pg to install 
``rm -f /var/lib/rpm/__db*``. 
I did not include it in the doc, I assume it was an issue only on Docker
* for the second error, it is b/c Ice is not in the PATH. I had to do the following for service https://github.com/ome/omero-install/pull/239/commits/3f0e9de60b683ebc453492c6a8f4047c6d4e7414

manics:
Are you mixing `yum` and `dnf`?

jburel:
Only dnf command is ``dnf module disable -y postgresql``

manics:
I'm wondering if mixing up the commands led to your corrupt rpm DB. If you replace the `yum` commands with `dnf` (or `dnf` with `yum`) does it work?

jburel:
No it is not the problem.
The rpm DB is corrupted before the dnf command is used.

jburel:
I have now replaced ``dnf`` command by ``yum`` equivalent i.e. https://github.com/ome/omero-install/pull/239/commits/cb5423db29b07bd1c760a1d1ff2d8955a4defe75
travis is still green (we still need to remove the corrupted file)

dominikl:
Indeed using `yum` instead of `dnf` doesn't solve the rpm db issue, but the `rm ...` does, so that's fine. After adding ice to the PATH I could startup omero. But I can't connect to it. Looks like the server is only listening on IPv6. So I disabled it `omero config set Ice.IPv6 0`. Then it listened on IPv4. Still can't connect. Disabled firewall `systemctl stop firewalld`, still can't connect. Then tried local connect:
```
[omero-server@localhost ~]$ omero login
Server: [localhost:4064]
Username: [omero-server]root
Password:
Ice.ConnectionRefusedException: localhost:4064 isn't running
```

`omero admin diagnostics` is fine by the way. With respect to that and taking into account that Travis is happy, I'd say merge the PR. Maybe it's just some Virtualbox issue on my side.

dominikl:
One more remark: When I used `127.0.0.1` instead of `localhost` the connection was initiated but failed with the well known SSL issue, which could be fixed with the 'certificates' cli plugin. It looks like my Centos8 installation is somehow messed up with respect to IP, firewall, hostname settings maybe (default Centos8 minimal server installation btw, nothing fancy).

jburel:
Merging so I can fix the certificates PR
both omero-install and omeroweb-install PRs have now been included

PR #2129, opened 2020-05-25
pin django-cors-headers for Django 1.11

joshmoore:
https://pypi.org/project/django-cors-headers/3.3.0/ dropped
supported for Django 1.11. Ping to earlier versions to prevent
a forced upgrade of Django to 3.x.

see https://github.com/ome/omero-web-docker/issues/50

### DIFF
--- a/omero/sysadmins/unix/install-web.rst
+++ b/omero/sysadmins/unix/install-web.rst
@@ -144,7 +144,7 @@ for more details on the settings.
 
 In the virtual environment where OMERO.web is installed, install the app as **root**::
 
-  $ pip install django-cors-headers
+  $ pip install 'django-cors-headers<3.3'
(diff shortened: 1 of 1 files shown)


jburel:
Change looks good.
I am wondering if we should not include the step in the "omeroweb-install" guides since it is a step in omero-web-docker.

joshmoore:
Agreed. I was a bit surprised that I didn't find it in omeroweb-install. Happy for my ome-docs PR to be closed if it gets migrated.

jburel:
Let's merge this PR so we are sure it is in the doc for the next release. I will look into updating omeroweb-install

jburel:
https://github.com/ome/omeroweb-install/issues/64

PR #2132, opened 2020-06-12
from omero.rtypes import rdouble, rint, rstring

AarghIMApiratez:
I got that rdouble was not loaded.

also this line is flaw

# create a rectangle shape (added to ROI below)
print(("Adding a rectangle at theZ: %s, theT: %s, X: %s, Y: %s, width: %s,")
   " height: %s" % (z, t, x, y, width, height))

### DIFF
--- a/omero/developers/Python.rst
+++ b/omero/developers/Python.rst
@@ -774,6 +774,7 @@ ROIs
 ::
 
     updateService = conn.getUpdateService()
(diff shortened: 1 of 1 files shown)


joshmoore:
> I got that rdouble was not loaded.

Thanks, @AarghIMApiratez. We've been working more on notebooks recently and this file has fallen behind. Sorry about that.

All the best,
~Josh.

PR #2135, opened 2020-06-28
Drop ubuntu1604

jburel:
Drop support for ubuntu 16.04
This PR replaces https://github.com/ome/ome-documentation/pull/2126


sbesson (approved):
The removal of the Ubuntu 16.04 walkthroughs is in agreement with the set of decisions discussed and reported in https://github.com/ome/ome-documentation/pull/2126#issuecomment-650184000. The version requirements page is updated accordingly. 

A few  high-level questions in the preparation of the upcoming release:

- the installation page liks CentOS8 and Ubuntu 20.04 under the `Recommended` walkthroughs. Should they not be under `Upcoming` to match the semantics of the requirements page?
- previously, we were leaning  towards a re-release of the docs only to fix broken installation walkthroughs. However, this PR contains a more substantial change as we are effectively fully dropping an unsupported system. I do not think silently removing doc pages is an option. Additionally, is thisworth at least a clear announcement to the community, maybe together with user-focused API additions to the `omero-py` (and `omero-web`?) stack?

jburel:
We should probably adjust the requirements page for C8 and Ubuntu 20.04 and mark them as "recommended".
We can announce it and indicate that people can find installation steps in omero-install and omeroweb-install as we previously did for debian 8 (we also moved it to https://github.com/ome/community-docs but we never use it afterwards)

sbesson:
:+1: for promoting the new walkthroughs at the same time as we drop the 16.04 one as this effectively offers at least 2 LTS alternatives.

Let's move the discussion on to the new PR but I would vote for marking C8/U20.04 as `supported/suboptimal` in that case rather than `supported/optimal` ? Primary reason for this semantics is the level of testing on these platforms is not as thorough as on C7 and U18.04.

jburel:
The counterargument to that is that U18.04 was marked as ``recommended`` even with limited testing.
I will push an extra commit to this PR and put them under ``Upcoming`` to avoid confusion we can then promote them next round.

jburel:
@sbesson C8 and U20.04 moved to upcoming

manics:
Ubuntu 18.04 is effectively "supported" by the community, there are enough image.sc posts with people's experience of running it for us to know it works. This isn't the case for Ubuntu 20.04 nor CentOS 8.

jburel:
We are not ready to add links to Ansible guides. So for this round we are good with installation.
Waiting on JSON doc PR #2137 (part of the omero-web release) then we should be good to go on the doc fron

PR #2136, opened 2020-06-30
Temporarily add imagej.net to ignore list

dgault:
Adding imagej.net to the linkcheck ignore list until the site is fully restored. See https://forum.image.sc/t/imagej-wiki-is-down/39672 for further details


jburel:
Closing this PR now that it has been restored and it is conflicting the decoupling of the doc

PR #2153, opened 2020-09-24
omero-py>=5.8.0: import automatically downloads JARs

manics:
Added in 5.8.0 https://github.com/ome/omero-py/pull/162


dominikl:
Looks good to me 👍

PR #2154, opened 2020-09-24
code-contributions: mention auto-formatter/pre-commit in progress

manics:
Follow-up from https://github.com/ome/omero-web/pull/218


joshmoore:
Seems like a good start. I imagine we will need to expand it as Java and Javascript come under the ax.

PR #2155, opened 2020-09-28
use range instead of xrange

jburel:
use range instead of xrange
Doc and example in openmicroscopy are out synch

cc @will-moore


will-moore:
👍  Looks good. `xrange` was Python 2 only.

PR #2156, opened 2020-09-29
Bump OMERO.py and OMERO.web versions for docs build.

mtbc:
Updates the autogen config.


jburel:
you should also bump blitz, server etc.

mtbc:
I should also be adjusting the correct file! :smiley: Will correct it.

jburel:
Merging. I will sort that the other version in a follow-up PR

PR #2161, opened 2020-10-05
Insight v5.5.14 is now released.

mtbc:
Adjust config accordingly.


dominikl:
Don't know what `autogen_db_version.py` does, but the change looks good to me 👍

jburel:
I will add that point to the release steps in the README of insight

PR #2166, opened 2020-11-09
Move "Reusing OMERO sessions" from public to web dev

manics:
"Reusing OMERO sessions" isn't suitable for public sharing since it provides a full read-write session with access to all data. Including it in the public user doc implies it's somewhat equivalent.

Background: https://forum.image.sc/t/omero-public-datasets-with-permanent-session-id/44939

### DIFF
--- a/omero/developers/Web/Webclient.rst
+++ b/omero/developers/Web/Webclient.rst
@@ -95,11 +95,16 @@ under an ``OME`` module.
 Reusing OMERO sessions
 ----------------------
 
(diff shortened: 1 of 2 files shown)


joshmoore:
MSTM. There may eventually be a mid-term place for this, but definitely better to be safe than sorry. Actually, does it need a bigger warning that it provides complete access, i.e. "Do not share"?

jburel:
Looks good.
Thanks

PR #2168, opened 2020-12-04
Remove broken centos 6 link

dominikl:
Removes the broken link to the Centos 6 packages. Also unifies the links to the Centos version specifications, instead of using https://access.redhat.com/support/policy/updates/errata#Life_Cycle_Dates for Centos 8 only (which looked a bit odd to me).


joshmoore (approved):
LGTM

PR #2170, opened 2021-01-06
Update Java instructions to access the Artifactory

sbesson:
The artifactory/maven endpoint is functional hits timeouts likely due to the
listing. In general, the Web UI endpoint should be more scalable for searching
for artifacts


joshmoore:
Merge is now passing; latest failing. Merging.

PR #2171, opened 2021-02-04
remove 9.4 link and add 11 and 12

jburel:
Links to pg 9.4 have now been removed from upstream.
Removing the links from our list and add 11 and 12 to the list

### DIFF
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -337,21 +337,16 @@ If no version is provided, a suitable repository is indicated.
       - Ubuntu
       - Homebrew
       - FreeBSD Ports
-    * - 9.4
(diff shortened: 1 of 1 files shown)


sbesson:
The removal of the entry matches the fact the the version is EOL and the YUM repos have been dropped. :+1:

On a related note, the final release of 9.5 will happen in 1 week after which PSQL 9.5 will become EOL. Should we remove the entry as part of this PR?

jburel:
By the time the doc is published, 9.5 will be removed so I have removed the entry

sbesson (approved):
All makes sense.

The only other thought while reading the change is that RHEL 6 is also EOL and these repos will likely disappear at some point. Probably worth cleaning in a separate PR/capturing as a separate issue?

jburel:
https://github.com/ome/ome-documentation/issues/2172 RHEL 6 removal

PR #2175, opened 2021-03-12
Fix 404 in pytest usage

joshmoore:
The stable URL is still present with the additional "docs" prefix,
but if the page is being removed in "latest", we might as well go
ahead and bump.

Fixes https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/771/


sbesson (approved):
The remaining error in https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/774/console is unrelated and might need to be fixed separately.

PR #2177, opened 2021-03-29
give the default port numbers for ws and wss

glyg:
Those cannot be found in the doc (only in `/etc/omero.properties`)


jburel:
Thanks @glyg

glyg:
thanks for the merge @jburel !

PR #2181, opened 2021-04-27
Certificates

jburel:
Adjust the doc to match changes made in https://github.com/ome/omero-install/pull/252


jburel:
We could drop/adjust the text before we configure the ``certificates`` since it is now required on all

omero/sysadmins/unix/server-centos7-ice36.rst
```diff
@@ -119,6 +119,14 @@ Configure the database and the location of the data directory:
     :start-after: #end-copy-omeroscript
     :end-before: #end-step04
 
+Weaker ciphers like ADH are disabled by default in OpenSSL 1.1+.
```
sbesson: Although the statement is correct, it does not match the fact that CentOS ships `openssl` 1.0.2.
Should we also include here a sentence mentioning that TLS 1.0/1.1 are EOL and getting dropped in the JDK packages?

omero/sysadmins/unix/server-centos7-ice36.rst
```diff
@@ -119,6 +119,15 @@ Configure the database and the location of the data directory:
     :start-after: #end-copy-omeroscript
     :end-before: #end-step04
 
+Weaker ciphers like ADH are disabled by default in new version of OpenSSL
+and TLS versions 1.0 and 1.1 have been dropped from JDK packages.
+This means that it is not possible to connect to an OMERO.server
+using any OMERO clients e.g. the Java Desktop client,
+the OMERO.web client or the CLI and import data. Run:
```
sbesson: This might need rephrasing as this reads as it is still not possible to connect. Maybe `In order to connect,```

```
In order to connect to an OMERO.server using any OMERO clients e.g. the Java Desktop client, the OMERO.web client or the CLI and import data, you need to generate self-signed certificates by running:
```

sbesson (commented):
One suggestion on the `certificates` paragraph to be clearer about the fact that clients should be able connect to the server.

Thinking more about the conditions for getting SSL errors, it might be worth expanding the Troubleshooting section about SSL issues while connecting to the server in https://github.com/ome/omero-documentation/blob/develop/omero/sysadmins/troubleshooting.rst#ssl-connection-issues to cover the client requirements (minimal omero-blitz/omero-py version) as well as the server requirements. This does need to happen as part of this PR which deals with the SSL certificates installation server-side  and I will try to open something separately

sbesson:
Additionally, came across https://docs.openmicroscopy.org/omero/5.6.3/sysadmins/client-server-ssl.html. Two questions:
- are these instructions are now superseded by the `omero-certificates` plugin
-  should this page remain as the reference documentation point for the server SSL certificates, in which case, it could be cross-linked from the installation walkthroughs for more details.

jburel:
I will add a link to the ssl page in the walkthroughs and rephrase the section about TLS1_2

joshmoore:
(Just re-noticed this PR this morning)

omero/sysadmins/unix/server-centos7-ice36.rst
```diff
@@ -119,6 +119,15 @@ Configure the database and the location of the data directory:
     :start-after: #end-copy-omeroscript
     :end-before: #end-step04
 
+Weaker ciphers like ADH are disabled by default in new version of OpenSSL
```
mtbc:
```suggestion
Weaker ciphers like ADH are disabled by default in new versions of OpenSSL
```
(also below)

omero/sysadmins/unix/server-centos7-ice36.rst
```diff
@@ -119,6 +119,17 @@ Configure the database and the location of the data directory:
     :start-after: #end-copy-omeroscript
     :end-before: #end-step04
 
+Weaker ciphers like ADH are disabled by default in new versions of OpenSSL
+and TLS versions 1.0 and 1.1 have been dropped from JDK packages.
+In order to connect to an OMERO.server using any OMERO clients
+e.g. the Java Desktop client, the OMERO.web client or the CLI and import data, 
+you need to generate self-signed certificates by running:
```
joshmoore: This might could use an `or-` statement about bringing your own certificate. That's possible by setting a configuration property but I don't think we have or could easily have an example for that though.

joshmoore (commented):
The fact that the block is repeated multiple times makes me wonder if it might (also?) need a home in the higher-level page, but that might be something we can come back to. In general, :+1: for upping the visibility of this requirement.

omero/sysadmins/unix/server-centos7-ice36.rst
jburel: I have centralised the text and added a note to the server-installation page.
Since we do not have an example about bringing own certificates. I have omitted it.

omero/sysadmins/unix/walkthrough/walkthrough_centos7.sh
```diff
@@ -72,7 +73,7 @@ $VENV_SERVER/bin/pip install omero-server[default]
 $VENV_SERVER/bin/pip install "omero-py>=5.8.0"
 #start-release-ice36
 cd /opt/omero/server
-SERVER=https://downloads.openmicroscopy.org/omero/5.6/server-ice36.zip
```
joshmoore: Why the change here?

omero/sysadmins/unix/ciphers
```diff
@@ -0,0 +1,5 @@
+Weaker ciphers like ADH are disabled by default in new versions of OpenSSL
```
joshmoore, line 1: Ah nice. This is an include. First I went looking for this page. Don't know if we want to identify them somehow, but big :+1: for the solution.

joshmoore (commented):
Few comments but looks nice.

omero/sysadmins/unix/walkthrough/walkthrough_centos7.sh
jburel: This should not there

jburel: sorted

omero/sysadmins/unix/ciphers
jburel: They could have ``.txt`` as an extension

jburel:
Any objection to merge this PR so I can move forward with https://github.com/ome/omero-install/issues/256
Or I push away the change made in this PR  in the debian9 installation file

joshmoore:
Assuming the client-ssl link is now on every walkthrough, LGTM :+1:

jburel:
Thanks

PR #2182, opened 2021-06-01
Update details on OMERO setup at IMCF / Biozentrum

ehrenfeu:
* update link to point directly to IMCF (where there is information about the OMERO instance)
* update amount of users and used storage space
* restructure (and update) resources a bit to make it more readable

### DIFF
--- a/omero/sysadmins/server-setup-examples.rst
+++ b/omero/sysadmins/server-setup-examples.rst
@@ -62,13 +62,13 @@ Stats
 -  36 TiB of data storage space, of which 16.6 TiB is currently in use
 -  Performance statistics to come
 
-Biozentrum, University of Basel
--------------------------------
+IMCF, Biozentrum, University of Basel
+-------------------------------------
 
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Thanks Niko

joshmoore:
❤️

sbesson:
@ehrenfeu it looks like https://www.biozentrum.unibas.ch/imcf has gone AWAL (`The requested page could not be found`). DO you know if the website is getting migrated and if yes what the new location is supposed to be?

ehrenfeu:
Hi @sbesson - you're right, the website is *currently* being migrated. The link should be re-established afterwards, at least that was the point of having permanent deep-links...

Will keep an eye on it though, thanks!

sbesson:
Thanks for the feedback and no problem. We'll temporarily ignore these broken links and get back to you in a few days/weeks if it's still 404

ehrenfeu:
Works again now. Thanks for pointing this out, it had actually been forgotten in the migration...

PR #2184, opened 2021-06-15
The key is dry_run, not dry-run

stephenogg:


jburel:
Thanks @stephenogg for spotting it
This now matches the output of ``omero import --help``
Screenshot 2021-06-16 at 11 51 43

PR #2186, opened 2021-07-02
specify min java version

jburel:
The Java version in ``conf.py`` is mainly used for the server installation
Client side requirement is still at Java 1.8
see https://github.com/ome/omero-insight/blob/master/build.gradle#L19

cc @sukunis

### DIFF
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -73,7 +73,7 @@ rst_epilog += """
 .. |version_web|  replace:: %s
 .. |iceversion| replace:: 3.6.5
 .. |postgresversion| replace:: 11
-.. |javaversion| replace:: 11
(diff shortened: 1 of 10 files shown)


omero/conf.py
```diff
@@ -74,6 +74,7 @@
 .. |iceversion| replace:: 3.6.5
 .. |postgresversion| replace:: 11
 .. |javaversion| replace:: 11
```
joshmoore: Is this effectively `javaversion_recommended`?

jburel: For the server installation yes it is.
The value was used for both clients and servers section leading to the confusion.
I will rename the variable so it is more explicit and to avoid misusage in the future

joshmoore (approved):
WFM

PR #2188, opened 2021-07-13
Mention OMERODIR in the server upgrade page

sbesson:
See https://forum.image.sc/t/omero-server-upgrade-error-from-5-5-1-to-5-6-3/55088

There is no mention of the `OMERODIR` envvar in the server upgrade page which can lead to confusion for users upgrading from OMERO < 5.6.

### DIFF
--- a/omero/sysadmins/server-upgrade.rst
+++ b/omero/sysadmins/server-upgrade.rst
@@ -300,7 +300,8 @@ Environment variables
 If you changed the directory name where the |release| server code
 resides, make sure to update any system environment variables. Before
 restarting the server, make sure your :envvar:`PATH` system environment
-variable is pointing to the new location.
+variable is pointing to the new location. Also make sure the :envvar:`OMERODIR`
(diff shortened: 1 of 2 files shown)


dominikl:
Looks good. Just wonder if it would make sense to also link to the https://docs.openmicroscopy.org/omero/5.6.2/sysadmins/unix/server-installation.html#environment-variables site just for further information about the `OMERODIR` variable.

sbesson:
:+1: pushed two extra commits to cross-link to the reference section in the installation page and expand the rationale for the `OMERODIR` envvar. I'll relist for review tomorrow.

joshmoore (approved):
Unfortunate that this is as complicated as it is for the user, but :+1: for helping as much as we can.

sbesson:
@dominikl have you had a chance to glance over the last two changes?

dominikl:
Sorry, thought Josh already approved it. Yes, looks good, ready to merge imo 👍

PR #2191, opened 2021-07-23
Fix link to example scripts

### LINKED ISSUE #2189: Link to example server scripts not found
The `ome/omero-example-scripts` repository doesn't exist any longer, so the link https://docs.openmicroscopy.org/omero/5.6.3/developers/scripts/index.html#finding-scripts is broken.

dominikl:
Fixes https://github.com/ome/omero-documentation/issues/2189 

Just fix broken link to example scripts repo.

### DIFF
--- a/omero/developers/scripts/index.rst
+++ b/omero/developers/scripts/index.rst
@@ -46,7 +46,7 @@ include:
 
 - `OMERO scripts <https://github.com/glencoesoftware/omero-user-scripts>`_ -
   Glencoe Software
-- `Example scripts <https://github.com/ome/omero-example-scripts>`_
+- `Example scripts <https://github.com/openmicroscopy/omero-example-scripts>`_
   - OME Team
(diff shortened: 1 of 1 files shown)


sbesson (approved):

joshmoore:
@sbesson : was there a reason that this repo didn't get migrated to github.com/ome? I'm not remembering one.

sbesson:
I think the primary reason is that this repository was forked from https://github.com/ome/omero-user-scripts and there is (was?) no possibility for an organization to fork itself.
In modern GitHub, I assume the forking would have bene replaced by using  templates.

joshmoore:
Probably something like that but with all the migrations from `openmicrosopy` to `ome` we could have transferred, no?

sbesson:
This is the error message I get when trying to transfer

Screenshot 2021-08-04 at 09 43 29

We need to remove the repo from the network but I haven't investigated this further.

PR #2196, opened 2021-11-19
Migrate the IDR information to the server-setup-examples section

sbesson:
Point to the IDR deployment page for more information about the architecture

--depends-on https://github.com/IDR/idr.openmicroscopy.org/pull/148

### DIFF
--- a/omero/sysadmins/server-setup-examples.rst
+++ b/omero/sysadmins/server-setup-examples.rst
@@ -117,3 +117,10 @@ Network infrastructure
 The server is hosted inside the faculty of medicine where the network works at
 100 Mbit/s. There are are 4 Gbit/s ports on the server but only one is
 currently in use.
+
+.. seealso::
+
+    `IDR deployment page <https://idr.openmicroscopy.org/deployment>`_
+        Information about the deployment of the
+        `Image Data Resource <https://idr.openmicroscopy.org>`_ which hosts
+        over 10 million images - see https://idr.openmicroscopy.org/studies.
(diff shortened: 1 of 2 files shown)


omero/sysadmins/server-setup-examples.rst
```diff
@@ -117,3 +117,11 @@ Network infrastructure
 The server is hosted inside the faculty of medicine where the network works at
 100 Mbit/s. There are are 4 Gbit/s ports on the server but only one is
 currently in use.
+
+.. seealso::
+
+    `IDR deployment page <https://idr.openmicroscopy.org/about/deployment>`_
+        Information about the deployment of the
+        `Image Data Resource <https://idr.openmicroscopy.org>`_ which hosts
+        over 10 million images - see
```
joshmoore: reviewing: https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/lastSuccessfulBuild/artifact/omero/_build/html/sysadmins/server-setup-examples.html

I think I wouldn't use "see" here since it makes it seem like that's where the info has. Can "over 10 million images" simply link to the studies page if someone wants that context? I could also see not using the "seealso" box but having this as a header like the others.

sbesson: Thanks, I agree the see also box seems to refer to the previous section, a63032f rewrites the information as a new section to prevent the confusion

joshmoore:
* :+1: much more understandable, thank you. Happy to go with it.
* Likely out of scope:
  - I'm unsure of the ordering logic for the page
  - If we ever get to the point that we have screenshots on this page, I could see re-generating the IDR metrics table from source ;)

sbesson:
Merging but I concur there is likely some restructuration to be done around this page/content especially if we had more examples. If there were enough exampes, I could probably imagine this metadata made available via a map representation with the different sites and their specifications.

PR #2197, opened 2021-11-30
GHA update

jburel:
The aim of this PR is to be the doc up-to-date when changes occur in upstream repositories i.e. 
- https://github.com/ome/openmicroscopy
- https://github.com/ome/omero-install
- https://github.com/ome/omeroweb-install

This combines with https://github.com/ome/omero-documentation/pull/2195 should simplify the release of the documentation

How does it work:
* Every day, the ``update`` action will be triggered e.g. https://github.com/jburel/ome-documentation/actions/runs/1521384260
* A Pull Request will be created if changes occur in the upstream repositories.

This workflow will be active when this  PR is merged 

cc @sbesson @joshmoore 

The change in autogen file will  be removed when/if we switch to rtd.

Edit: The workflow is now a bit more complicated due to use of ``omero db script``

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,33 +1,54 @@
 name: Update
 
 on:
-  schedule:
-    - cron: '* 0 * * *'
+  push:
 jobs:
   build:
     runs-on: ubuntu-20.04
+    env:
+      ICE_HOME: /opt/ice-3.6.5 # location where Ice is installed
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
           python-version: 3.8
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Install slice2java
+        run: |
+          sudo apt-get install -y libmcpp-dev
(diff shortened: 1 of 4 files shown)


omero/autogen_docs
```diff
@@ -34,9 +35,10 @@ sed 1,5d advanced-help.txt > omero/downloads/inplace/advanced-help.txt
 omero import --javahelp 2> java-help.txt || echo "Dumped"
 sed 1,5d java-help.txt > omero/downloads/cli/help.out
 
-echo "Generating DB script example"
-omero db script --password secretpassword 2>&1 | sed "s|Saving to.*/|Saving to /home/omero/|" > omero/downloads/cli/db-script-example.txt
-rm OMERO*sql
+#echo "Generating DB script example"
```
joshmoore: Was this supposed to be protected by an `if DB` statement?

jburel: This was commented out during testing and I forgot about it
The workflow will be a bit more complex for that section to work but since we are planning to move to rtd, we need to  everything updated via action. I will push extra commits to sort this section

jburel: This is now back on see https://github.com/jburel/ome-documentation/actions/runs/1529949753
This requires a more complex setup in the action unfortunately see a024358

.github/workflows/update.yml
```diff
@@ -0,0 +1,58 @@
+name: Update
```
joshmoore: Could the name be a bit more descriptive than "Update"?

joshmoore (commented):
Mostly looks like a few `if`-catches for edge cases for this new build environment. Other than one minor comment about the use of "Update" :+1:

jburel:
Several PRs have been merged in omero-install and omeroweb-install. 
Merging this PR so we can test the workflow and check output tomorrow

PR #2201, opened 2021-12-06
Clarify tables version to be installed

jburel:
This fixes a problem with the installation of ``tables``


sbesson (approved):
This matches the version mentioned in https://github.com/ome/omero-cli-server/blob/fdaa0d3a9e252f37c9773275d66537d39bff24af/setup.py#L60-L62. 

A possible alternative would be to install all dependencies using `pip install omero-cli-server`. The advantage would be to reduce the maintenance of the documentation and  the number of places to keep in sync. The downside is that it feels confusing to install a package called `*server` in a client environment.

jburel:
I thought about the same thing and due to the name I opted not to go that route

PR #2206, opened 2021-12-17
add include label

jburel:
Add label when creating the PR 
see https://github.com/peter-evans/create-pull-request
cc @sbesson


sbesson (approved):
Makes sense to me. I assume the permissions should be sufficient to create the label in addition to opening the PR.

jburel:
The permissions should be enough for a label.

PR #2214, opened 2022-01-26
Removes specific 3.6 python version

### LINKED ISSUE #2213: omero-py install instructions should not pin pyhthon3.6
Hi!

The current version of the installation instructions for `omero-py` either from conda or pip, instruct the users to use python 3.6 when newer versions of python work just fine. As python 3.6 gets older (I think it is not supported by numpy anymore) I think this should be clarified. 

Here is the file:

https://github.com/ome/omero-documentation/blob/759dcafeec65e31e79920ffcd4742ed34764643f/omero/users/cli/installation.rst

I'm opening a PR with a proposed fix straight away.

glyg:
Fixes #2213

### DIFF
--- a/omero/users/cli/installation.rst
+++ b/omero/users/cli/installation.rst
@@ -5,7 +5,7 @@ Installation
     but may be supported in the future.
 
 Since OMERO 5.6, only Python 3 is supported.
-We assume that you have already installed Python 3.6.
+We assume that you have already installed Python 3.6 or higher. You can ensure that your python executable is correct with the ``python --version`` command.
 
 We recommend installing client library ``omero-py`` and the |CLI| plugins
 in a Python virtual environment.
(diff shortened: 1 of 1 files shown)


jburel:
Thanks. The changes look. We are working on updating the documentation newer python versions
see https://github.com/ome/omero-install/pull/260 and/or https://github.com/ome/omero-install/pull/259

sbesson:
Should we get this merged ahead of the upcoming OMERO.server/documentation release @jburel ?

Note that as per https://github.com/ome/conda-omero-py/issues/26 there is also a question of whether we should update the documentation to recommend a Conda installation using the new `conda-forge` package. If we do not have the confidence to make the switch at this point, this is already a good step forward.

jburel:
Let's merge this PR, we can adjust if need be

PR #2216, opened 2022-02-16
Update django-redis version

jburel:
Update version when generating doc

### DIFF
--- a/autogen_omero.sh
+++ b/autogen_omero.sh
@@ -35,9 +35,8 @@ fi
 set +u # PS1 issue
 . $DOCVENV/bin/activate
 set -u
-python -m pip install "omero-web>=5.6.0"
+python -m pip install "omero-web[redis]"
 python -m pip install future 'ansible<2.7'
-python -m pip install "django-redis==5.0.0"
 python -m pip install -U PyYAML==5.1
 python -m pip install scc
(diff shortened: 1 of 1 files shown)


autogen_omero.sh
```diff
@@ -37,7 +37,7 @@ set +u # PS1 issue
 set -u
 python -m pip install "omero-web>=5.6.0"
 python -m pip install future 'ansible<2.7'
-python -m pip install "django-redis>=4.4,<4.9"
+python -m pip install "django-redis==5.0.0"
```
sbesson: As we work on the Django ecosystem upgrade, for me, this raises the question of whether these Django extra packages `redis`, `cors-pipelines` should not be managed at the `omero-web` level via `extra` dependencies.

jburel: This is certainly a good time to have the discussion

jburel: I think they should be included at the omero-web level, at least the ``redis`` one.
@chris-allan @will-moore @joshmoore what is your feeling?

joshmoore: Certainly no objection for an `extra[]` from my side.

jburel: PR https://github.com/ome/omero-web/pull/359

PR #2217, opened 2022-02-16
remove centos8

jburel:
Remove reference to CentOS 8.
Walkthrough will automatically removed when omero-install and omeroweb-install are updated.


sbesson:
From my side, similar question as https://github.com/ome/omero-documentation/pull/2135 i.e. would we redirect URLs like https://docs.openmicroscopy.org/omero/latest/sysadmins/unix/server-centos8-ice36.html and https://docs.openmicroscopy.org/omero/latest/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos8-ice3.6.html and if yes, to which location?

jburel:
One option is to not to point to any walkthrough but to https://docs.openmicroscopy.org/omero/5.6.3/sysadmins/unix/server-installation.html
so user can select one of the "supported" platforms

jburel:
@sbesson I forgot to relist this PR since you last reviewed it. Now that centOS 8 steps have been removed from [omeroweb-install](https://github.com/ome/omeroweb-install/pull/82) and [omero-install](https://github.com/ome/omero-install/pull/263)
This PR needs to be integrated too.

Okay with the redirects?

sbesson (approved):

PR #2221, opened 2022-03-30
Add more examples, correct existing ones

### LINKED ISSUE #2220: Correct search documentation
Add more examples to search documentation on https://docs.openmicroscopy.org/omero/5.6.3/developers/Modules/Search.html. Examples should use 
1. Quotation marks
2. Wildcards
3. Combination of 1. and 2.

In the process, correct the erroneous statements in https://docs.openmicroscopy.org/omero/5.6.3/developers/Modules/Search.html - 

Reported in https://forum.image.sc/t/seach-omero-tags-for-exact-match/65113/

pwalczysko:
Fixes https://github.com/ome/omero-documentation/issues/2220

cc @will-moore @joshmoore

### DIFF
--- a/omero/developers/Modules/Search.rst
+++ b/omero/developers/Modules/Search.rst
@@ -69,7 +69,7 @@ Assuming these entries above for Image.name:
 
 With the same entries as above and adding a wildcard:
 
--  searching for **\*FP** returns 1, 2, 3 and 4. As this example shows, **leading wildcards in the Graphical User Interface are allowed**, but disallowed when using the API directly, see below in the developers section.
+-  searching for **\*FP** returns 1, 2, 3 and 4. As this example shows, **leading wildcards in the Graphical User Interface are allowed**, but must be explicitly enabled when using the API directly, see below in the developers section.
 -  searching for **GF\*** returns 1, 2, 3 and 4.
 -  searching for **GFP-\*** returns no results, but **GFP.\*** returns 1, 2, 3 and 4. Only hyphen and underscore do not return results in this situation, the other non-alpha-numerical characters do.
 -  searching for **"\*FP-H2B"** returns no results. This is not a wildcard query, but a term query, and is the same as searching for **"FP-H2B"**.
--- a/omero/developers/Modules/Search.rst
+++ b/omero/developers/Modules/Search.rst
@@ -39,15 +39,15 @@ Queries
 Search queries are very similar to Google searches. When search terms
 are entered without a prefix ("name:"), then the default field will be
 used which combines all available fields. Otherwise, a prefix can be
-added to restrict the search. The search terms will be by default parsed as if there was "OR" operator
-between them (see example in the "Indexing" paragraph). 
-Further, tokenizing happens on all non-alphanumerical signs, such as underscore, 
+added to restrict the search. The search terms will be parsed as if there was an "OR" operator
+between them (see examples in the "Indexing" paragraph).
+The tokenizing happens on all non-alpha-numerical characters, such as underscore,
 hyphen etc. Again, when the term is tokenized, 
-the default parsing of the tokens is with "OR" operator between the tokens. 
+the query is built by combining the tokens with an "OR".
 The search terms or the tokens created from them as above 
 must precisely match the indexed entries. 
 This means for example that a search term `tes` 
-is **not** matching the indexed entry `test` and the search 
+will **not** match the indexed entry `test` and the search
 will accordingly give no result. 
 
 Indexing
@@ -66,20 +66,20 @@ FullTextAnalyzer <src/main/java/ome/services/fulltext/FullTextAnalyzer.java>`.
 
 Assuming these entries above for Image.name:
 
--  searching for **GFP-H2B** returns 1, 2, 3 and 4, because of the tokenizing on the hyphen and joining of the tokens by an **OR**. This behavior is changed by adding wildcards or quotations, see below.
--  searching for **"GFP H2B"** only returns 1 and 2, since the quotes enforce the exact sequence of the terms.
+-  searching for **GFP-H2B** returns 1, 2, 3 and 4, because of the tokenizing on the hyphen and joining of the tokens by an **OR**.
+-  searching for **"GFP H2B"** only returns 1 and 2, since the quotes enforce the exact sequence of the terms and the query is built with an **AND**.
 -  searching for **GFP H2B** returns 1, 2, 3 and 4, since the two terms
    are joined by an **OR**.
 -  searching for **"GFP-H2B"** returns 1 and 2.
 
 With the same entries as above and adding a wildcard:
 
--  searching for **"GF\*"** returns no results.
+-  searching for **\*FP** returns 1, 2, 3 and 4. As this example shows, **leading wildcards in the Graphical User Interface are allowed**, but disallowed when using the API directly, see below in the developers section.
 -  searching for **GF\*** returns 1, 2, 3 and 4.
--  searching for **"GFP-\*"** returns 1, 2, 3 and 4.
--  searching for **GFP-\*** returns no results. 
--  searching for **\*FP-H2B** returns no results.
--  searching for **"\*FP-H2B"** returns no results.
+-  searching for **GFP-\*** returns no results, but **GFP.\*** returns 1, 2, 3 and 4. Only hyphen and underscore do not return results in this situation, the other non-alpha-numerical characters do.
+-  searching for **"\*FP-H2B"** returns no results. This is not a wildcard query, but a term query, and is the same as searching for **"FP-H2B"**.
+-  searching for **"GF\*"** returns no results. Again, not a wildcard query, the same as searching for **"GF"**.
+-  searching for **"GFP-\*"** returns 1, 2, 3 and 4. Again, not a wildcard query, the same as searching for **"GFP-"**.
 
 Information for developers
 --------------------------
--- a/omero/developers/Modules/Search.rst
+++ b/omero/developers/Modules/Search.rst
@@ -39,7 +39,16 @@ Queries
 Search queries are very similar to Google searches. When search terms
 are entered without a prefix ("name:"), then the default field will be
 used which combines all available fields. Otherwise, a prefix can be
-added to restrict the search.
+added to restrict the search. The search terms will be by default parsed as if there was "OR" operator
+between them (see example in the "Indexing" paragraph). 
+Further, tokenizing happens on all non-alphanumerical signs, such as underscore, 
+hyphen etc. Again, when the term is tokenized, 
+the default parsing of the tokens is with "OR" operator between the tokens. 
+The search terms or the tokens created from them as above 
+must precisely match the indexed entries. 
+This means for example that a search term `tes` 
+is **not** matching the indexed entry `test` and the search 
+will accordingly give no result. 
 
 Indexing
 --------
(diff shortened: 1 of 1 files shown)


omero/developers/Modules/Search.rst
```diff
@@ -57,10 +66,21 @@ FullTextAnalyzer <src/main/java/ome/services/fulltext/FullTextAnalyzer.java>`.
...
+-  searching for **GFP-H2B** returns 1, 2, 3 and 4, because of the tokenizing on the hyphen and joining of the tokens by **OR**.
+This behavior is changed by adding wildcards or quotations, see below.
+-  searching for **"GFP H2B"** only returns 1 and 2, because the quotes enforces the exact sequence of the terms.
+-  searching for **GFP H2B** returns 1, 2, 3 and 4, since the two terms
    are joined by an **OR**.
+-  searching for **"GFP-H2B"** returns 1 and 2.
+
+With the same entries as above and adding a wildcard:
+
+-  searching for **"GF\*"** returns no results.
```
joshmoore: This one surprises me a bit. Perhaps the quotes go a step further than just dropping syntax like `-` and `_` and even disable the `*`.

pwalczysko: Double-checked the behaviour. It is as specified.

pwalczysko: Verified all queries also in OMERO.insight. No difference with OMERO.web.

joshmoore: Definitely believe you, we should just use your examples to try to also explain the semantics.

omero/developers/Modules/Search.rst
```diff
@@ -39,7 +39,16 @@ Queries
 Search queries are very similar to Google searches. When search terms
 are entered without a prefix ("name:"), then the default field will be
 used which combines all available fields. Otherwise, a prefix can be
-added to restrict the search.
+added to restrict the search. The search terms will be by default parsed as if there was "OR" operator
```
will-moore: Maybe  "By default, the search terms will be parsed as if there was *an* "OR" operator...
 - Actually, what is non-default? Do you need the "by default" in this sentence?

omero/developers/Modules/Search.rst
```diff
@@ -39,7 +39,16 @@ Queries
...
+between them (see example in the "Indexing" paragraph). 
+Further, tokenizing happens on all non-alphanumerical signs, such as underscore, 
+hyphen etc. Again, when the term is tokenized, 
+the default parsing of the tokens is with "OR" operator between the tokens.
```
will-moore: Hmmm - I wonder if this paragraph needs to separate out the tokenizing from the querying...
The tokenizing is just all "non-alpha-numerical characters" (not signs).
The querying is about combing the tokens with "AND" and "OR", but I think it's confusing to mention "OR" in the same sentence as tokenising, since tokenising is independent of operators?

omero/developers/Modules/Search.rst
```diff
@@ -39,7 +39,16 @@ Queries
...
-added to restrict the search.
+added to restrict the search. The search terms will be by default parsed as if there was "OR" operator
+between them (see example in the "Indexing" paragraph). 
+Further, tokenizing happens on all non-alphanumerical signs, such as underscore, 
+hyphen etc. Again, when the term is tokenized, 
+the default parsing of the tokens is with "OR" operator between the tokens. 
+The search terms or the tokens created from them as above 
+must precisely match the indexed entries. 
+This means for example that a search term `tes` 
+is **not** matching the indexed entry `test` and the search
```
will-moore: `tes` will **not** match the indexed...

omero/developers/Modules/Search.rst
```diff
@@ -57,10 +66,20 @@ FullTextAnalyzer <src/main/java/ome/services/fulltext/FullTextAnalyzer.java>`.
 
 Assuming these entries above for Image.name:
 
--  searching for **GFP-H2B** returns 1 and 2.
--  searching for **"GFP H2B"** also returns 1 and 2.
--  searching for **GFP H2B** returns 1, 2, and 3, since the two terms
+-  searching for **GFP-H2B** returns 1, 2, 3 and 4, because of the tokenizing on the hyphen and joining of the tokens by an **OR**. This behavior is changed by adding wildcards or quotations, see below.
```
will-moore: I'm not sure you need "This behaviour is changed... " sentence since that's true for all queries.

omero/developers/Modules/Search.rst
pwalczysko: @joshmoore I found https://stackoverflow.com/questions/35797085/how-does-lucene-work-with-quotes-and-wildcards - according to that answer, the ``"GF*"`` is interpreted as if it would be ``GF`` which does not igve any results in our case. The wording is that
`` Again, phrase query. Not a wildcard query in any way. You can't use wildcards in phrases. The analyzer will remove the punctuation, making this identical to the last one.`` 
Happy for me to use this wording, warning the user off using wildcards in phrases and move on here ?

omero/developers/Modules/Search.rst
```diff
@@ -57,10 +66,20 @@ FullTextAnalyzer <src/main/java/ome/services/fulltext/FullTextAnalyzer.java>`.
...
+-  searching for **"GFP H2B"** only returns 1 and 2, since the quotes enforce the exact sequence of the terms and the query is built with an **AND**.
+-  searching for **GFP H2B** returns 1, 2, 3 and 4, since the two terms
    are joined by an **OR**.
+-  searching for **"GFP-H2B"** returns 1 and 2.
+
+With the same entries as above and adding a wildcard:
+
+-  searching for **\*FP** returns 1, 2, 3 and 4. As this example shows, **leading wildcards in the Graphical User Interface are allowed**, but disallowed when using the API directly, see below in the developers section.
+-  searching for **GF\*** returns 1, 2, 3 and 4.
+-  searching for **GFP-\*** returns no results, but **GFP.\*** returns 1, 2, 3 and 4. Only hyphen and underscore do not return results in this situation, the other non-alpha-numerical characters do.
```
pwalczysko, line 73: @will-moore @joshmoore This clearly shows that hyphen and underscore are somehow preferred/special, not like other non-alphanumericals. I do not think I can google this out, as it must be something specific for our implementation.

pwalczysko:
@will-moore @joshmoore I thought I explained everything, but it is only the [quotations and wildcard combinations](https://github.com/ome/omero-documentation/pull/2221#discussion_r841945057) which seems to be "logically" explained. There is [one more inexplicable behaviour](https://github.com/ome/omero-documentation/pull/2221#discussion_r845483976). I do not believe this will be (soon or satisfactorily, or ever) explained by me, so would suggest to leave it at the examples level ?

omero/developers/Modules/Search.rst
```diff
@@ -39,7 +39,16 @@ Queries
...
+added to restrict the search. The search terms will be parsed as if there was an "OR" operator
+between them (see examples in the "Indexing" paragraph).
+The tokenizing happens on all non-alpha-numerical characters, such as underscore,
+hyphen etc. Again, when the term is tokenized, 
+the query is built by combining the tokens with an "OR".
+The search terms or the tokens created from them as above 
+must precisely match the indexed entries. 
+This means for example that a search term `tes` 
+will **not** match the indexed entry `test` and the search
+will accordingly give no result.
```
will-moore: Does the following sound OK (trying to separate the tokenization and "OR")?

The search term is first split into "tokens" and these are combined into a search query. The tokenizing happens on all non-alpha-numerical characters, such as underscore, hyphen etc.  The query is built by combining the tokens with an "OR" operator (see examples in the "Indexing" paragraph). The search terms or the tokens created....

omero/developers/Modules/Search.rst
```diff
@@ -57,10 +66,20 @@ FullTextAnalyzer <src/main/java/ome/services/fulltext/FullTextAnalyzer.java>`.
 
 Assuming these entries above for Image.name:
 
--  searching for **GFP-H2B** returns 1 and 2.
--  searching for **"GFP H2B"** also returns 1 and 2.
--  searching for **GFP H2B** returns 1, 2, and 3, since the two terms
+-  searching for **GFP-H2B** returns 1, 2, 3 and 4, because of the tokenizing on the hyphen and joining of the tokens by an **OR**.
+-  searching for **"GFP H2B"** only returns 1 and 2, since the quotes enforce the exact sequence of the terms and the query is built with an **AND**.
```
will-moore: replace 'terms' with 'tokens'? Same in following line.

will-moore:
A couple of minor suggestions.

omero/developers/Modules/Search.rst
pwalczysko: Yes, I am adding the word "space", as now your formulation encompasses also the most common case of what is perceived by user as  "multiple terms entry" such as ``GFP H2B kidney``

pwalczysko:
Fixes pushed. All built successfully on https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/lastSuccessfulBuild/artifact/

omero/developers/Modules/Search.rst
```diff
@@ -57,10 +61,20 @@ FullTextAnalyzer <src/main/java/ome/services/fulltext/FullTextAnalyzer.java>`.
...
 Assuming these entries above for Image.name:
 
--  searching for **GFP-H2B** returns 1 and 2.
--  searching for **"GFP H2B"** also returns 1 and 2.
--  searching for **GFP H2B** returns 1, 2, and 3, since the two terms
+-  searching for **GFP-H2B** returns 1, 2, 3 and 4, because of the tokenizing on the hyphen and joining of the tokens by an **OR**.
+-  searching for **"GFP H2B"** only returns 1 and 2, since the quotes enforce the exact sequence of the tokens and the query is built with an **AND**.
+-  searching for **GFP H2B** returns 1, 2, 3 and 4, since the two tokens
    are joined by an **OR**.
+-  searching for **"GFP-H2B"** returns 1 and 2.
```
joshmoore: This could be combined with the second example, no? So does that mean that the hyphen is just always factored out?

omero/developers/Modules/Search.rst
```diff
@@ -57,10 +61,20 @@ FullTextAnalyzer <src/main/java/ome/services/fulltext/FullTextAnalyzer.java>`.
...
+
+With the same entries as above and adding a wildcard:
+
+-  searching for **\*FP** returns 1, 2, 3 and 4. As this example shows, **leading wildcards in the Graphical User Interface are allowed**, but disallowed when using the API directly, see below in the developers section.
```
joshmoore:
```suggestion
-  searching for **\*FP** returns 1, 2, 3 and 4. As this example shows, **leading wildcards in the Graphical User Interface are allowed**, but must be explicitly enabled when using the API directly, see below in the developers section.
```

omero/developers/Modules/Search.rst
```diff
@@ -57,10 +61,20 @@ FullTextAnalyzer <src/main/java/ome/services/fulltext/FullTextAnalyzer.java>`.
...
+-  searching for **GF\*** returns 1, 2, 3 and 4.
+-  searching for **GFP-\*** returns no results, but **GFP.\*** returns 1, 2, 3 and 4. Only hyphen and underscore do not return results in this situation, the other non-alpha-numerical characters do.
+-  searching for **"\*FP-H2B"** returns no results. This is not a wildcard query, but a term query, and is the same as searching for **"FP-H2B"**.
```
joshmoore: I really like these, but they might get their own sections that just explains, "a wildcard in quotes doesn't count"

omero/developers/Modules/Search.rst
pwalczysko: Will do. Yes, in all my experiments this holds.

omero/developers/Modules/Search.rst
pwalczysko: Surprising new result. 
``"GFP*H2B"`` does return 1 and 2. So more precisely, wildcard in quotes does not count as wildcard, but as a non-alphanumerical character on which to tokenize. (?)

joshmoore: I think that explanation makes a lot of sense.

will-moore (approved):
Thanks @pwalczysko - looks good 👍

PR #2224, opened 2022-04-15
update graph

jburel:
In preparation for the upcoming server release


sbesson (approved):

sbesson:
My understanding is that this change corresponds to the security fixes made in OMERO 5.6.1. Not from https://github.com/ome/omero-model/blob/3sdb8ba94b0dab6211cbb6175712e406af197752/src/main/resources/mappings/meta.ome.xml#L68, these properties are still formally marked as deprecated but the PR might reflect the fact they are no longer exposed via the API.

PR #2225, opened 2022-04-15
drop pg 9.6

jburel:
PostgreSQL has been removed from omero-install and ansible role

### DIFF
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -337,13 +337,6 @@ If no version is provided, a suitable repository is indicated.
       - Ubuntu
       - Homebrew
       - FreeBSD Ports
-    * - 9.6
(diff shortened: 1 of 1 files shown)


sbesson:
This table captures the state of the distribution by the various Linux distros right? While PSQL 9.6 has been removed from the RHEL/CentOS repos, I have not checked whether the same holds for Debian/Ubuntu.

Otherwise, looks good to me

jburel:
you could still get it for Debian 11 for example e.g. https://apt.postgresql.org/pub/repos/apt/dists/bullseye-pgdg/9.6/
I opted to remove to avoid confusion since this is not really a version we recommend

sbesson:
Noted, I am fine with removing it completely as per the unrecommended argument. Is there a corresponding support level we can use in the table above to express the status of PSQL 9.6 on OMERO 5.6? Or should it be kept as deprecated?

jburel:
I think keeping it as ``deprecated`` is fine.

sbesson (approved):

PR #2226, opened 2022-04-15
bump versions

jburel:
Bump versions of the various components


sbesson (approved):
I do not have the details of all the latest component releases but this looks good at a glance

PR #2233, opened 2022-04-22
Convert pyramids before import warning

will-moore:
See https://forum.image.sc/t/omero-pyramid-generation-time/33463/27

### DIFF
--- a/omero/sysadmins/limitations.rst
+++ b/omero/sysadmins/limitations.rst
@@ -59,7 +59,10 @@ that users avoid these issues by converting
 their data to `pyramidal OME-TIFF <https://www.openmicroscopy.org/2018/11/29/ometiffpyramid.html>`_
 files before importing into OMERO. A number of suitable tools are available such as
 `bioformats2raw & raw2ometiff <https://www.glencoesoftware.com/blog/2019/12/09/converting-whole-slide-images-to-OME-TIFF.html>`_,
-`bfconvert`, `Kheops <https://github.com/BIOP/ijp-kheops>`_, `tifffile`, `aicsimageio`, `libvips` and `QuPath`.
+`bfconvert <https://docs.openmicroscopy.org/latest/bio-formats/users/comlinetools/conversion.html>`_,
+`Kheops <https://github.com/BIOP/ijp-kheops>`_, `tifffile <https://pypi.org/project/tifffile/>`_,
+`aicsimageio <https://github.com/AllenCellModeling/aicsimageio>`_,
+`libvips <https://github.com/libvips/libvips>`_ and `QuPath <https://qupath.github.io/>`_.
 
 Large images with floating-point pixel data
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/omero/sysadmins/limitations.rst
+++ b/omero/sysadmins/limitations.rst
@@ -46,17 +46,20 @@ File format support
 Large images
 ^^^^^^^^^^^^
 
-When you import an image over a certain size, OMERO will generate a Pyramid of lower resolution
+When you import an image over a certain size, OMERO will generate a pyramid of lower resolution
 images if it doesn't already exist in the file. The threshold size is configurable using
 :property:`omero.pixeldata.max_plane_height` and
 :property:`omero.pixeldata.max_plane_width` but set to 3192x3192 pixels by
-default. However, this process can be very resource-intensive for larger images, for example see
+default. However, this process can be very resource-intensive, depending on the size of the
+image as well as the image format and any data compression used, for example see
 `PixelData threads and pyramid generation issues <https://forum.image.sc/t/pixeldata-threads-and-pyramid-generation-issues/49794>`_.
 
-It is recommend that users avoid these issues by converting
+The OMERO pyramid generation process should be considered as deprecated and instead it is recommended
+that users avoid these issues by converting
 their data to `pyramidal OME-TIFF <https://www.openmicroscopy.org/2018/11/29/ometiffpyramid.html>`_
-files before importing into OMERO, as described at
-`Converting Whole Slide Images to OME-TIFF: A New Workflow <https://www.glencoesoftware.com/blog/2019/12/09/converting-whole-slide-images-to-OME-TIFF.html>`_.
+files before importing into OMERO. A number of suitable tools are available such as
+`bioformats2raw & raw2ometiff<https://www.glencoesoftware.com/blog/2019/12/09/converting-whole-slide-images-to-OME-TIFF.html>`_,
+`bfconvert`, `Kheops<https://github.com/BIOP/ijp-kheops>`_, `tifffile`, `aicsimageio`, `libvips` and `QuPath`.
 
 Large images with floating-point pixel data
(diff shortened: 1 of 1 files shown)


omero/sysadmins/limitations.rst
```diff
@@ -43,15 +43,28 @@ user, you cannot make the group into a private group.
 File format support
 -------------------
 
+Large images
+^^^^^^^^^^^^
+
+When you import an image over a certain size, OMERO will generate a Pyramid of lower resolution
```
sbesson: maybe we worth saying that its only when the XY dimensions exceed a certain size

sbesson: no capital for `Pyramid`

omero/sysadmins/limitations.rst
```diff
@@ -43,15 +43,28 @@ user, you cannot make the group into a private group.
...
+images if it doesn't already exist in the file. The threshold size is configurable using
+:property:`omero.pixeldata.max_plane_height` and
+:property:`omero.pixeldata.max_plane_width` but set to 3192x3192 pixels by
+default. However, this process can be very resource-intensive for larger images, for example see
```
sbesson: The resource-intensiveness largely depends on a set of factor. Size is one of them but compression/decompression and concurrency can be other issues. Thus I would not attempt to be too specific about the cause for resource intensiveness

omero/sysadmins/limitations.rst
```diff
@@ -43,15 +43,28 @@ user, you cannot make the group into a private group.
...
+`PixelData threads and pyramid generation issues <https://forum.image.sc/t/pixeldata-threads-and-pyramid-generation-issues/49794>`_.
+
+It is recommend that users avoid these issues by converting
```
sbesson: I would propose to add a first statement along the lines of 

"The OMERO pyramid generation process should be considered as deprecated and instead it is recommended..."

omero/sysadmins/limitations.rst
```diff
@@ -43,15 +43,28 @@ user, you cannot make the group into a private group.
...
+their data to `pyramidal OME-TIFF <https://www.openmicroscopy.org/2018/11/29/ometiffpyramid.html>`_
+files before importing into OMERO, as described at
```
sbesson: maybe this requires `for instance` or a similar statement to clarify this is not the only workflow

sbesson (changes requested):
Overall, this largely corresponds to what I had in mind in terms of clarifying the pyramid generation service. Thanks for opening it.

A few suggestions made inline. Happy to discuss alternate formulations.

Probably the overarching statement is whether we are all happy with calling the pyramid generation service as deprecated. If so, I could also imagine adding it in the release announcement so that people start making adjustements to their workflow /cc @jmarie

jburel:
@sbesson we can certainly indicate that it is deprecated in the announcement. If we want to mark it as deprecated in the code, we will have to re-publish packages

will-moore:
What does "deprecated" really mean? The `RoiService` has been deprecated for about 10 years, and Insight for almost as long.

omero/sysadmins/limitations.rst
will-moore: Do we have any examples of other workflows or tools we can point to for creating pyramidal OME-TIFF?

sbesson: A few examples of open-source libraries/plugins/applications that can generate multi-resolution OME-TIFF and come to mind:

- BIo-Formats command-line utility i.e. `bfconvert`
- the https://github.com/BIOP/ijp-kheops ImageJ /plugin
- `tifffile` / `aicsimageio`
- `libvips`
- QuPath

sbesson:
> What does "deprecated" really mean? The RoiService has been deprecated for about 10 years, and Insight for almost as long.

I think the definition in https://github.com/ome/omero-documentation/blob/7e0751ef0bb79518288d5be84e5e5a768fd380b2/omero/sysadmins/version-requirements.rst#support-levels largely applies here i.e. the service is functional but suboptimal in many cases and we are not planning to continue its development and/or add new features.
Unlike the 2 examples mentioned above, in this case, there is a rich ecosystem of tools allowing to generate multi-resolution images in a scalable way so we are not leaving the community high and dry but we  want to encourage them to rethink their loading workflow.

omero/sysadmins/limitations.rst
will-moore: That's covered by the next sentence.

will-moore:
@sbesson @jburel Pushed some fixes

omero/sysadmins/limitations.rst
```diff
@@ -43,15 +43,31 @@ user, you cannot make the group into a private group.
...
+:property:`omero.pixeldata.max_plane_width` but set to 3192x3192 pixels by
+default. However, this process can be very resource-intensive, depending on the size of the
+image as well as the image format and any data compression used, for example see
+`PixelData threads and pyramid generation issues <https://forum.image.sc/t/pixeldata-threads-and-pyramid-generation-issues/49794>`_.
+
+The OMERO pyramid generation process should be considered as deprecated and instead it is recommended
+that users avoid these issues by converting
+their data to `pyramidal OME-TIFF <https://www.openmicroscopy.org/2018/11/29/ometiffpyramid.html>`_
+files before importing into OMERO. A number of suitable tools are available such as
+`bioformats2raw & raw2ometiff<https://www.glencoesoftware.com/blog/2019/12/09/converting-whole-slide-images-to-OME-TIFF.html>`_,
```
sbesson: This probably requires a space before `<` to render properly as an hyperlink

omero/sysadmins/limitations.rst
```diff
@@ -43,15 +43,31 @@ user, you cannot make the group into a private group.
...
+`bfconvert`, `Kheops<https://github.com/BIOP/ijp-kheops>`_, `tifffile`, `aicsimageio`, `libvips` and `QuPath`.
```
sbesson: Same as above

sbesson: Do we need hyperlinks for all the conversion tools/utilities?

omero/sysadmins/limitations.rst
```diff
@@ -43,15 +43,31 @@ user, you cannot make the group into a private group.
...
+
 Large images with floating-point pixel data
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 :model_doc:`Pyramids <omero-pyramid/>` of image tiles are currently not
```
sbesson, line 70: While the content of this section remains unchanged i.e. floating-point data is not support in the OMERO pyramid generation service, we could add a concluding sentence explaining that floating-point pyramidal OME-TIFF can be generated using the same tools as described above.

sbesson (commented):
A few formatting/linking questions but otherwise this reads finr to me.

omero/sysadmins/limitations.rst
will-moore: Well, I could find a few URLs via google, but so could anyone else. I don't know if the ones I find are more likely to be correct or stable than ones that a user might find. To be consistent, I could either add links that I find, or remove links from Kheops and link to `blog post` for bf2raw?

will-moore:
Added links and other fixes

sbesson (approved):
Thanks. This reads well to me at least.

PR #2237, opened 2022-04-28
Improve update action

jburel:
* Set the title of the PR and indicate with repositories have been modified 
* Improve the commit message 
* Delete branch

cc @joshmoore


joshmoore:
Definitely :+1: the general idea. What would be the easiest way to generate one for review?

jburel:
This is an example. I made a minor modification to jburel/openmicroscopy on develop branch
https://github.com/jburel/ome-documentation/pull/14

.github/workflows/update.yml
```diff
@@ -54,8 +54,31 @@ jobs:
...
+          fi
+          echo "::set-output name=value::$value"
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4
         with:
           labels: include
           token: ${{secrets.ACTION_API_TOKEN}}
+          commit-message: ${{ steps.upstream_repo.outputs.value }}
+          delete-branch: true
+          title: 'Changes from upstream repositories'
```
joshmoore: Is it possible that multiple repositories create one or could `${value}` be used here as well?

jburel: Value will indicate the repositories that have been modified e.g. ``openmicroscopy, omero-install``

joshmoore: Sorry, I noticed I was unclear. I mostly meant for `title`. So you're thinking is since there are multiple, we don't try to encode them into the title?

jburel: HAppy to change the title

jburel: https://github.com/jburel/ome-documentation/pull/16

joshmoore:
Thanks for the update in https://github.com/jburel/ome-documentation/pull/16. Happy to move forward with this and we can keep iterating.

PR #2238, opened 2022-05-03
Code samples fixes

Tom-TBT:
After following the examples, I noticed a few things that I could fix, such as missing parenthesis, imports or indentation.

It's not much, and not thorough either, but I figured that it could still help the next person.


jburel:
Thanks for submitting the adjustments
The changes have been made in the code see https://github.com/ome/openmicroscopy/blob/develop/examples/Training/python/Read_Data.py
but not propagated to the doc.

PR #2241, opened 2022-05-24
use conda-forge channel

jburel:
Remove usage of ``ome`` channel

### DIFF
--- a/omero/sysadmins/unix/server-install-homebrew.rst
+++ b/omero/sysadmins/unix/server-install-homebrew.rst
@@ -124,16 +124,19 @@ Using conda (preferred)
 #. Install Conda.
    See `miniconda <https://docs.conda.io/en/latest/miniconda.html>`_ for more details.
 
-#. Create virtual environments named ``omeropy`` and ``omeroweb``::
+#. Create virtual environments named ``omeropy``::
 
     $ conda create -n omeropy -c conda-forge python=3.8 zeroc-ice omero-py
+
+#. Create virtual environments named ``omeroweb``, activate it and install dependencies::
+
+    $ conda create -n omeroweb -c conda-forge python=3.8 zeroc-ice omero-py
+    $ conda activate omeroweb
+    $ pip install "omero-web>=\ |version_web|"
 
 #. Activate the virtual environments::
 
     $ conda activate omeropy
-
-    # In a different terminal:
-    $ conda activate omeroweb
 
 #. You can now use the ``omero`` command. You will also need to ensure you are in 
    the appropriate environment when you install additional modules::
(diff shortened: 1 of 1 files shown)


omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -126,8 +126,8 @@ Using conda (preferred)
 
 #. Create virtual environments named ``omeropy`` and ``omeroweb``::
 
-    $ conda create -n omeropy -c ome python=3.6 zeroc-ice36-python omero-py
-    $ conda create -n omeroweb -c ome python=3.6 zeroc-ice36-python omero-web
+    $ conda create -n omeropy -c conda-forge python=3.6 zeroc-ice36-python omero-py
```
sbesson: Should this be updated to use Python 3.8?

Also `zeroc-ice36-python` should now be `zeroc-ice` and can be ommitted as it is a dependency.

omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -126,8 +126,8 @@ Using conda (preferred)
...
+    $ conda create -n omeroweb -c conda-forge python=3.6 zeroc-ice36-python omero-web
```
sbesson: There is no https://anaconda.org/conda-forge/omero-web. Should this be pip installed for now?

sbesson (changes requested):

omero/sysadmins/unix/server-install-homebrew.rst
jburel: yes
For 3.8 I was thinking too, but was planning to do a "python" review in a follow-up PR

jburel:
Also the command ``brew link python`` is probably not installing 3.7 (I have not checked)

omero/sysadmins/unix/server-install-homebrew.rst
```diff
@@ -126,8 +126,7 @@ Using conda (preferred)
 
 #. Create virtual environments named ``omeropy`` and ``omeroweb``::
 
-    $ conda create -n omeropy -c ome python=3.6 zeroc-ice36-python omero-py
-    $ conda create -n omeroweb -c ome python=3.6 zeroc-ice36-python omero-web
+    $ conda create -n omeropy -c conda-forge python=3.8 zeroc-ice omero-py
```
sbesson, line 129: The command does not really map the sentences above which talk about 2 virtual environments. Should this block also create an `omeroweb` environment installing `omero-py` from `conda-forge` and add an extra `pip install omero-web`command post activation below?

sbesson (changes requested):

omero/sysadmins/unix/server-install-homebrew.rst
jburel: I will review.  I also think we should remove that walkthrough since we are no longer testing it.

PR #2243, opened 2022-05-26
update action to use the action-ice

jburel:
This action was introduced while working on ``omego``. See https://github.com/ome/omego/blob/master/.github/workflows/workflow.yml

This simplifies the action


joshmoore (approved):
:+1:

PR #2245, opened 2022-05-30
Update extlinks to comply with Sphinx 5 deprecation warnings

jburel:
See https://github.com/ome/bio-formats-documentation/pull/250 for background


sbesson (changes requested):
The CI logs still indicate a log of warnings related to deprecated extlink formatting.

Looking at the files, it seems that the `extlinks` dictionary were in `common/conf.py` but not int`omero/conf.py` (also it looks like the dictionary is duplicated but that's likely outside the scope of this PR)

jburel:
the duplication is probably left over from the split "doc/contributing"

sbesson (approved):
Spotted a few warnings (unrelated to this work) in the logs which I will capture as a separate issue.

PR #2250, opened 2022-06-27
Warning

jburel:
Doc warning noticed when using newer version of sphinx.
See also https://github.com/ome/openmicroscopy/pull/6327 and https://github.com/ome/omeroweb-install/pull/84


sbesson:
Is there a public build log with the original warnings?

jburel:
This is interesting. I had 4.4.0 installed locally in one conda env, I had 5.0.2 in a different one. It looks that the warnings are not there with 5.0.2 which is the version used by rtd. So we could ignore those changes. Nonetheless, they are probably more inline with the definition of variables in ``conf.py``

sbesson:
Thanks, I was asking because I initially looked at https://readthedocs.org/api/v2/build/17287235.txt and could not find these warnings.

No objection to using extlinks whenever it's justified although looking at the changes, I am slightly concerned by changes of this type:

```diff
- in `github <https://github.com/ome/openmicroscopy>`_.
+ in :omero_subs_github_repo_root:`Github <openmicroscopy>`.
```

Not only is the the diff longer (while the extlink is partly meant to shorten reference) but I find the diff makes the link extremely hard to read and maintain. What does `omero_subs_github_repo_root` stands for ?

jburel:
for ``https://github.com/ome`` as defined in ``conf.py``

prior to this PR it was used in a limited number of places so
renaming should not be a problem
```
git grep "omero_subs_github_repo_root"
conf.py:    'omero_subs_github_repo_root': (github_root + 'ome/%s', None),
developers/Model.rst:which can be mapped to Hibernate mapping files. See an example in :omero_subs_github_repo_root:`omero-model` at:
developers/Model.rst:however, the files in :omero_subs_github_repo_root:`omero-model` under :model_sourcedir:`src/main/resources/mappings`
developers/Model.rst:is then used to turn the mapping files into generated Java code in :omero_subs_github_repo_root:`omero-model` under the
developers/Model.rst::omero_subs_github_repo_root:`omero-server`.
developers/Model.rst:External to the server code is the :omero_subs_github_repo_root:`omero-blitz` layer. These classes are in the
developers/Model.rst:In :omero_subs_github_repo_root:`omero-blitz`, the generated Ice files along with the hand-written Ice files from
developers/Model.rst:relationships can be defined simply in the mapping files. One example in :omero_subs_github_repo_root:`omero-model` is
developers/Server/Properties.rst:configurers" in the following file in :omero_subs_github_repo_root:`omero-server`:
developers/build-system.rst::omero_subs_github_repo_root:`omero-build` and :omero_subs_github_repo_root:`omero-gradle-plugins`. 
developers/build-system.rst:#. in :omero_subs_github_repo_root:`omero-model`, adjust the version of ``ome:formats-gpl`` in 
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-model`
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-common`
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-romio`
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-renderer`
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-server`
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-blitz`
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-gateway-java`
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-py`
developers/testing.rst:  - :omero_subs_github_repo_root:`omero-web`
```

sbesson:
I am still not convinced that these extlinks help readability and maitenance but if we consider they are, I would rename to prevent confusion. Looking at other changes

```
:omero_subs_github_repo_root:`prod-playbooks <prod-playbooks>`
```

or

```
:omero_subs_github_repo_root:`homebrew-alt <homebrew-alt>`
```

don't make sense as these are not OMERO-specific repositories. This decision will be required in order to move forward with https://github.com/ome/openmicroscopy/pull/6327

jburel:
Since the warnings are no longer printed out when using sphinx 5.0.x, we can ignore and close this PR.
My main concern was that it could turn into something else now that we delegate our doc generation to a 3rd party.

PR #2251, opened 2022-06-27
Add name and description opts, add link to it from limitations

pwalczysko:
see https://github.com/ome/omero-documentation/issues/2249 for background.

Adds -n and -x options.

@sbesson Also added one example of usage. To be honest, I was confused myself, cf. paragraph "java options" of the same doc which seems to be asking for two double-dashes. Nevertheless, tested the example which I added as well as the variants -n and --name and -x and --description, all works.

### DIFF
--- a/omero/users/cli/import.rst
+++ b/omero/users/cli/import.rst
@@ -32,8 +32,6 @@ Some of the options available to the import command are:
 
 .. option:: -h, --help
 
-    Examples of options available to the import command,
-
 .. option:: -s SERVER, -p PORT, -U USERNAME, -g GROUPNAME
 
     To avoid prompts for servername, port, username and group, use::
@@ -50,7 +48,6 @@ Some of the options available to the import command are:
 
     See :doc:`import-target` for more information on import targets.
 
-
 .. option:: -n NAME, --name NAME
 .. option:: -x DESCRIPTION, --description DESCRIPTION
 
@@ -59,10 +56,9 @@ Some of the options available to the import command are:
         $ omero import image.tif -n "control image1" -x "PBS Control"
         $ omero import image.tif --name image2.tif --description "negative control"
 
+.. option:: --logprefix [LOGPREFIX]
 
-  --logprefix [LOGPREFIX]               Directory or file prefix for --file and --errs
-  --file [FILE]                         File for storing the standard output from the Java process
-  --errs [ERRS]                         File for storing the standard error from the Java process
+    Directory or file prefix for --file and --errs
 
 .. option:: --file FILE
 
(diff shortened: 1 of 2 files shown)


omero/users/cli/import.rst
```diff
@@ -40,13 +40,14 @@ Some of the options available to the import command are:
...
 
-.. option:: -d DATASET_ID, -r SCREEN_ID, -T TARGET, --target TARGET
+.. option:: -d DATASET_ID, -r SCREEN_ID, -T TARGET, --target TARGET, -n NAME, --name NAME, -x DESCRIPTION, --description DESCRIPTION
     
     To import images into a Dataset::
 
         $ omero import image.tif -d 2
         $ omero import image.tif -T Dataset:id:2
         $ omero import image.tif -T Dataset:name:Sample01
+        $ omero import image.tif --name "Control image1" --description "PBS Control" -T Dataset:name:Controls
```
sbesson: Since name and description are othogonal to the target containers, I originally thought of splitting them into their own option sections, what do you think?

sbesson (commented):
The double dash is used to separate the arguments passed to the underlying Java command-line importer i.e. `omero import -- --java_arg_1 value_1 --java_arg_2 value_2`

The confusion stems from the fact the name and description arguments are defined both at the Python and at the Java level. I agree the Python form (without the double dash) should be preferred in the documentation.

One suggestion about splitting the addition into two separate blocks. Otherwise, looks great

pwalczysko:
@sbesson fixes pushed, also fixed some older remnants of imprecisely added blocks in https://github.com/ome/omero-documentation/pull/2251/commits/816256f26e3407e1f91e776e3cbf2f5722f41ac4

omero/users/cli/import.rst
```diff
@@ -50,9 +48,17 @@ Some of the options available to the import command are:
...
-  --errs [ERRS]                         File for storing the standard error from the Java process
+.. option:: -n NAME, --name NAME
+.. option:: -x DESCRIPTION, --description DESCRIPTION
+
+    To change the name of an image and add a description::
+
+        $ omero import image.tif -n "control image1" -x "PBS control"
+        $ omero import image.tif --name image2 --description second_batch
+
+.. option:: --logprefix [LOGPREFIX]
```
sbesson: Interestingly `--logprefix` is defined 3 options below so this can be simply removed.

sbesson (changes requested):
Looks good. One duplicated option to remove and then it should be ready to merge

omero/users/cli/import.rst
pwalczysko: HAH. Great, so this one was defined **only** three times :)

pwalczysko:
Fix pushed

sbesson (approved):

PR #2260, opened 2022-07-04
Remove doc version value

jburel:
make the release process more visible


README.rst
```diff
@@ -134,17 +146,6 @@ To switch back to a development version run::
 Specifying ``major``, ``minor`` or ``patch`` depending on whether the development branch will be a `major, minor or patch release <https://semver.org/>`_. This will also add the ``SNAPSHOT`` suffix.
```
sbesson: Interestingly, I completed skipped that step for the OMERO 5.6.5 documentation release. This is still a requirement?

Since we are now heavily relying on build number prefixes to update the docs i.e. `5.6.5-1, `5.6.5-2`, is this also something that `bumpversion` can handle?

sbesson (commented):
The re-organization makes sense. One open question about the release workflow though

README.rst
jburel: Technically speaking we could get rid of it. 
It is no longer used since we now use the openmicroscopy version see https://github.com/ome/omero-documentation/blob/develop/omero/conf.py#L70

sbesson (approved):
So merge and push a new tag to fix https://github.com/ome/omero-documentation/issues/2198#issuecomment-1171287930?

jburel:
I will push a new tag

PR #2262, opened 2022-07-08
Add PostgreSQL 13 and 14 to the version requirements table

sbesson:
Tentatively mark both versions under the same support levels as PostgreSQL 12 for the OMERO 5.6 series but as discussed this morning testing is on its way.

### DIFF
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -317,8 +317,22 @@ OMERO support policies
       - |Recommended|
       - |Recommended|
     * - 12
-      - Oct 2019
+      - Sep 2019
       - Nov 2024
+      - |Upcoming|
+      - |Upcoming|
+      - |Supported|
+      - |Supported|
+    * - 13
+      - Sep 2020
+      - Nov 2025
+      - |Upcoming|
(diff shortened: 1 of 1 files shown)


jburel:
See also https://github.com/ome/omero-install/pull/267

joshmoore:
:+1: though:
 - a) I could see marking something else for OMERO6
 - b) maybe we shouldn't even be listing OMERO6 🙂 

Screen Shot 2022-07-11 at 12 23 54

sbesson:
That's larger than the original scope of this PR but since you are asking, I am totally in favor of option b). Trying to anticipate the requirement for the next major release of OMERO had its merits but to this date we have no defined OMERO 6 either in terms of content or in terms of timeline.

Given the longevity of the OMERO 5.6 line, we are already adjusting the recommendations for the third-party dependencies and operating systems to match the reality of software support. As such, I would be fine with focusing this page on OMERO 5.6

joshmoore:
Also happy to hold off. To some extent what we're almost doing is the reverse of Postgresql. We will likely be bumping the suggested version in the 5.6.x line which means maybe we have columns per _patch_ releases.

PR #2268, opened 2022-08-08
Update link to RFC2616 section

sbesson:
The former link was redirecting but the rewrite rule seems to have changed recently leading to a 404 error message after redirection

See https://latest-ci.openmicroscopy.org/jenkins/job/OMERO-docs/1080/ and former builds


will-moore (approved):
New link looks good. 👍

sbesson:
Looks like the redirect got fixed under the hood. Merging this anyways as it removes another layer of redirection.

PR #2270, opened 2022-08-18
add whitenoise to upgrade

jburel:
Improve upgrade page see https://forum.image.sc/t/omero-web-upgrade-to-django-3-2-issue/68344/9

### DIFF
--- a/omero/sysadmins/omeroweb-upgrade.rst
+++ b/omero/sysadmins/omeroweb-upgrade.rst
@@ -45,6 +45,11 @@ benefits to upgrading:
 
       $ pip install --upgrade 'omero-py>=\ |version_py|'
 
+Upgrade ``whitenoise`` to at least version ``5.3.0``.
+
+.. parsed-literal::
+
+      $ pip install --upgrade 'whitenoise>=5.3.0'
 
 Configuration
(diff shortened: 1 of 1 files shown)


joshmoore:
Is this temporary since it's an extra step for the user? Would adding this pin in omero-web suffice? (with a new release of omero-web)

jburel:
``whitenoise`` is currently not a dependency of ``omero-web``. It requires an extra installation step e.g. https://omero.readthedocs.io/en/stable/sysadmins/unix/install-web/walkthrough/omeroweb-install-centos7-ice3.6.html
That's why it was also added to the upgrade page
Including it in ``omero-web`` will definitely simplify the installation.

joshmoore:
Ah. Do you know the reasoning for only having `pip install whitenoise` in the docs and not in the repo itself?

jburel:
By not in the repo, do you mean a dependency of omero-web?
I do not know the reason why it was not included in the first place.
This is certainly a candidate to introduce similar to what we do with ``"django-cors-headers`` recently

jburel:
Closing this PR, ``whitenoise`` is now a dependency of omero-web. This included in 5.16.0

PR #2274, opened 2022-09-12
fix the link

jburel:
fix the link. this will be overridden when a new release openmicroscopy
The correct fix is https://github.com/ome/omero-model/pull/86 but it will require a full stack release


will-moore (approved):
URL looks good (old one is 404).
👍

PR #2276, opened 2022-09-20
tmp remove the include labels

jburel:
Label should be added back when https://github.com/ome/omero-model/pull/86 is merged and the change integrated in openmicroscopy


jburel:
Merging so we can fix the build

PR #2285, opened 2022-12-08
Upgrade

jburel:
Add some info about newer versions of omero.web and omero.py
This currently in README of omero-web and omero-py respectively.

### DIFF
--- a/omero/developers/Web.rst
+++ b/omero/developers/Web.rst
@@ -15,7 +15,10 @@ to provide modular web tools, such as the webclient, webgateway and
 the JSON api app. This modular framework
 makes it possible to extend OMERO.web with your own apps.
 
-Since version ``5.14.0``, ``OMERO.web`` depends on Django to ``3.2.x``.  Plugin developers should read the `Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_ detailing how to migrate their plugin(s) to Django 3.2.x.
+Since version 5.14.0, OMERO.web depends on Django 3.2.x. 
+Plugin developers should read the
+`Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_
+detailing how to migrate their plugin(s) to Django 3.2.x.
 
 OMERO.web infrastructure
 ------------------------
--- a/omero/developers/Web.rst
+++ b/omero/developers/Web.rst
@@ -15,7 +15,7 @@ to provide modular web tools, such as the webclient, webgateway and
 the JSON api app. This modular framework
 makes it possible to extend OMERO.web with your own apps.
 
-Since version ``5.14.0``, ``OMERO.web`` requires an update of Django to ``3.2.x``.  Plugin developers should read the `Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_ detailing how to migrate their plugin(s) to Django 3.2.x.
+Since version ``5.14.0``, ``OMERO.web`` depends on Django to ``3.2.x``.  Plugin developers should read the `Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_ detailing how to migrate their plugin(s) to Django 3.2.x.
 
 OMERO.web infrastructure
 ------------------------
--- a/omero/developers/Web.rst
+++ b/omero/developers/Web.rst
@@ -15,7 +15,7 @@ to provide modular web tools, such as the webclient, webgateway and
 the JSON api app. This modular framework
 makes it possible to extend OMERO.web with your own apps.
 
-Since vesion ``5.14.0``, ``OMERO.web`` requires an update of Django to ``3.2.x``.  Plugin developers should read the `Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_ detailing how to migrate their plugin(s) to Django 3.2.x.
+Since version ``5.14.0``, ``OMERO.web`` requires an update of Django to ``3.2.x``.  Plugin developers should read the `Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_ detailing how to migrate their plugin(s) to Django 3.2.x.
 
 OMERO.web infrastructure
 ------------------------
--- a/omero/developers/Web.rst
+++ b/omero/developers/Web.rst
@@ -15,6 +15,8 @@ to provide modular web tools, such as the webclient, webgateway and
 the JSON api app. This modular framework
 makes it possible to extend OMERO.web with your own apps.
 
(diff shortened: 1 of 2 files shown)


omero/developers/Web.rst
```diff
@@ -15,6 +15,8 @@ to provide modular web tools, such as the webclient, webgateway and
 the JSON api app. This modular framework
 makes it possible to extend OMERO.web with your own apps.
 
+Since vesion ``5.14.0``, ``OMERO.web`` requires an update of Django to ``3.2.x``.  Plugin developers should read the `Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_ detailing how to migrate their plugin(s) to Django 3.2.x.
```
sbesson:
```suggestion
Since version ``5.14.0``, ``OMERO.web`` requires an update of Django to ``3.2.x``.  Plugin developers should read the `Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_ detailing how to migrate their plugin(s) to Django 3.2.x.
```

Also since Django will be a transitive dependency for developers, maybe `required an update of Django` a should be rephased as `depends on Django 3.2` ?

omero/sysadmins/server-upgrade.rst
```diff
@@ -324,6 +324,8 @@ up to date to ensure that security updates are applied:
       $ pip install --upgrade 'omero-py>=\ |version_py|'
 
 
+Since version ``5.13.0,`` the use of ``omero certificates`` is required to ensure that an OMERO server installation has, at minimum, a self-signed certificate.
```
sbesson: :+1: for adding a note in the upgrade page. Is it worth putting under a dedicated heading as it's not really related to `Dependencies`?

Two comments:
- `Since version 5.13.0` refers to OMERO.py 5.13.0 but server certificates might already have been required e.g. for Java clients
- `the use of ``omero certificates`` is required`: I would rephrase the requirement since what is needed is the presence and the configuration of a server certificate. `omero certificates` is definitely the recommended tool for generating self-signed certificates but it is also possible to deploy them manually e.g. as described in https://omero.readthedocs.io/en/stable/sysadmins/client-server-ssl.html.

sbesson (changes requested):

omero/sysadmins/server-upgrade.rst
jburel: @sbesson proposed changes in [8e221b9](https://github.com/ome/omero-documentation/pull/2285/commits/8e221b9d1f1ed808916e7571bcba7f6274b24008)

omero/sysadmins/server-upgrade.rst
```diff
@@ -324,6 +324,11 @@ up to date to ensure that security updates are applied:
       $ pip install --upgrade 'omero-py>=\ |version_py|'
 
 
+SSL verification
```
sbesson: `SSL verification` is a bit confusing as this section does not give give instructions on how to verify.

Looking at https://omero--2285.org.readthedocs.build/en/2285/sysadmins/server-upgrade.html#upgrade-steps, all sub-sections were originally named after actions which clash with this title name. Should this upgrade guide suggest that `omero certificates` should be executed unless a manual certificate has been manually installed?

sbesson (changes requested):
Looks better, but I think we want to be more precise on what the sysadmin should do during the upgrade process.

Only other comment is that this PR makes inconsistent usage of verbatim mode for referring to version numbers. I don't think we are using verbatim anywhere else so we might want to use normal text and keep verbatim for commands like `omero certificates`

omero/sysadmins/server-upgrade.rst
jburel: Reviewed in [f253aec](https://github.com/ome/omero-documentation/pull/2285/commits/f253aecf4e69ab76d8485d60bf7291a4ce214371)

omero/developers/Web.rst
```diff
@@ -15,6 +15,8 @@ to provide modular web tools, such as the webclient, webgateway and
 the JSON api app. This modular framework
 makes it possible to extend OMERO.web with your own apps.
 
+Since version ``5.14.0``, ``OMERO.web`` depends on Django to ``3.2.x``.  Plugin developers should read the `Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_ detailing how to migrate their plugin(s) to Django 3.2.x.
```
sbesson:
```suggestion
Since version 5.14.0, OMERO.web depends on Django 3.2.x. 
Plugin developers should read the
`Guide <https://github.com/ome/omero-web/blob/master/MIGRATION_TO_DJANGO_32_GUIDE.md>`_
detailing how to migrate their plugin(s) to Django 3.2.x.
```

omero/sysadmins/server-upgrade.rst
```diff
@@ -324,6 +324,13 @@ up to date to ensure that security updates are applied:
       $ pip install --upgrade 'omero-py>=\ |version_py|'
 
 
+Server certificate
+^^^^^^^^^^^^^^^^^^
+Since OMERO 5.6.2,  the recommended way to ensure that all OMERO server installations have, at minimum, a self-signed certificate is to use the `OMERO server certificate management plugin <https://github.com/ome/omero-certificates>`_. The plugin will generate or update your self-signed certificates and configure the OMERO.server. Each time, you run the command to generate or update the certificates is run, the server needs to be restarted.
+If you prefer to configure the OMERO server manually, check
+:doc:`/sysadmins/client-server-ssl`.
+In OMERO.py version ``5.13.0``, Anonymous Diffie-Hellman (ADH) default configuration has been removed. Please ensure that self-signed certificates have been generated and the OMERO.server configured accordingly.
```
sbesson:
```suggestion
In OMERO.py version 5.13.0, the Anonymous Diffie-Hellman (ADH) default configuration has been removed. Please ensure that self-signed certificates have been generated and the OMERO.server configured accordingly.
```

omero/sysadmins/server-upgrade.rst
```diff
@@ -324,6 +324,13 @@ up to date to ensure that security updates are applied:
       $ pip install --upgrade 'omero-py>=\ |version_py|'
 
 
+Server certificate
+^^^^^^^^^^^^^^^^^^
+Since OMERO 5.6.2,  the recommended way to ensure that all OMERO server installations have, at minimum, a self-signed certificate is to use the `OMERO server certificate management plugin <https://github.com/ome/omero-certificates>`_. The plugin will generate or update your self-signed certificates and configure the OMERO.server. Each time, you run the command to generate or update the certificates is run, the server needs to be restarted.
```
sbesson: ` Each time, you run the command to generate or update the certificates is run, the server needs to be restarted.`

I don't understand this sentence

omero/sysadmins/server-upgrade.rst
```diff
@@ -324,6 +324,13 @@ up to date to ensure that security updates are applied:
...
+If you prefer to configure the OMERO server manually, check
```
sbesson:
```suggestion
If you prefer to configure the OMERO server certificate manually, check
```

sbesson (changes requested):

omero/sysadmins/server-upgrade.rst
jburel: I changed the sentence half way through and did not read it again.
Maybe something like ``for the configuration to take effect, the server needs to be restarted`` or we omit it

sbesson (approved):

PR #2290, opened 2023-01-12
Changes from upstream repositories:  openmicroscopy

snoopycrimecop:
- This PR contains changes from upstream repositories

Auto-generated by [create-pull-request][1]

[1]: https://github.com/peter-evans/create-pull-request


jburel:
I will push a tag with the changes to update the doc

sbesson:
Is OMERO.web  scheduled for release soon? https://github.com/ome/omero-web/pull/434 includes an update to the configuration page that should be propagated to the doc (automatically?) and could benefit from being included in the next tag.

jburel:
@sbesson yes. We have one more PR to include then we will release a new version.

jburel:
I assume you are referring to https://github.com/ome/omero-web/pull/429.
The doc was not automatically updated since no command is run to do so after web changes. I will have to check

jburel:
Looking at the code
When omero-web is released, we should have the property upgraded
i.e.
```
DOCVENV/bin/pip install "omero-web[redis]"

echo "Generating configuration properties page"
$DOCVENV/bin/omero config parse --rst | sed "s|$SUFFIX||" | sed "s|$WORKSPACE|/home/omero|" > omero/sysadmins/config.rst

```
This could be changed and instead of using the release version, we could use ``master``.
But this means that parameters could be inadvertently added to the doc before a release of ``omero-web``

sbesson:
> This could be changed and instead of using the release version, we could use master.

I strongly vote against this exactly for the reason you are bringing up. We should never publish documentation for unreleased code under https://omero.readthedocs.io/en/stable/.

jburel:
> > This could be changed and instead of using the release version, we could use master.
> 
> I strongly vote against this exactly for the reason you are bringing up. We should publish documentation for unreleased code under https://omero.readthedocs.io/en/stable/.

Exactly. so we are on the same page.

jburel:
@sbesson omero-web changes are now included. 
I will merge and tag

jburel:
Doc updated on rtd

PR #2292, opened 2023-01-31
Update threadpool config docs

kkoz:
Update the omero config docs to match the Javadocs in https://github.com/ome/omero-server/pull/155. In particular, specify that `max_threads` and `idle_timeout` don't do anything and that `min_threads` actually controls the maximum number of threads in the threadpool.

### DIFF
--- a/omero/sysadmins/config.rst
+++ b/omero/sysadmins/config.rst
@@ -452,11 +452,11 @@ but your database installation will need to be configured
 to accept *at least* as many, preferably more, connections
 as this value.
 
-The related values omero.threads.max_threads and
+The related values omero.threads.min_threads and
 omero.threads.background_threads do *not* need to be
 increased by the same amount. A system will be more stable
-if background_threads is less than max_threads and
-max_threads is less than poolsize.
+if background_threads is less than min_threads and
+min_threads is less than poolsize.
 
 Default: `10`
 
@@ -1224,9 +1224,11 @@ Default: `600000`
 
 omero.threads.background_threads
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Number of threads from the max_threads pool that can
+Number of threads from the min_threads pool that can
 be used at any given time for background tasks like
-import.
+import. Note that if this value is less than min_threads,
+min_threads will limit the number of background
+tasks which can run simultaneously.
 
 Default: `10`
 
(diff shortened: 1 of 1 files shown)


jburel:
Apologies If I was not clearer earlier today
the content of ``config.rst`` is automatically generated (see https://github.com/ome/omero-documentation/blob/develop/.github/workflows/update.yml) The text will need to be adjusted in https://github.com/ome/omero-server/blob/master/src/main/resources/omero-server.properties.

sbesson:
Two quick thoughts:

- in terms of timeline, the current workflow requires a full release of the JARs (and the server?) in order to automatically update the documentation, is that correct?
- having the documentation staged at  https://omero--2292.org.readthedocs.build/en/2292/sysadmins/config.html is useful to review  the proposed change. Would it still make sense to iterate on this until we agree on the description before merging (?) and backporting the final version of the text to the properties file ?

jburel:
It will require a release yes
 I was pointing that the changes also need to be made in omero-server otherwise next time we release the text will be reverted back
I followed the same approach previously with some pg link see https://github.com/ome/omero-documentation/pull/2274

joshmoore:
If the intent is to simply fix the docs ASAP, :+1:. However, if there's a chance that things will stay this way, I'd suggest we swap min and max threads to say that "min_threads" does nothing for less long-term confusion.

In my brief discussion with Chris, we didn't talk about idle_timeout so I don't know if that has any hope of ever being resurrected or if it should actually be deprecated here.

kkoz:
Thanks @jburel I will make the changes in omero-server as well once we finalize our decision about what to do.
@joshmoore I don't mind swapping min_threads and max_threads in theory, but the value is actually both. I think that to the user max_threads is a more important number, so that's what I emphasized in the docs. However, technically what we're setting here is core threads, which is the number of threads that we want to hang around forever without being cleaned up. So once min_threads threads are started, there will never be fewer than that many in the pool. So one question is how to communicate that most effectively. Another question is, if we swap min_threads and max_threads, then when people update their systems they will suddenly be increasing the number of max threads for USER and BACKGROUND tasks from 5 to 50 (with default settings). This could have serious performance implications. I think we might need a more staged deprecation process where we require users to explicitly set a new config `user_background_threads` in order to increase the thread pool size above min_threads. So I would advocate in the short term for just updating the docs and not changing how the configs actually affect the code.

kkoz:
@jburel I updated the text in `omero-server.properties` in https://github.com/ome/omero-server/pull/155.

jburel:
Closing this PR. Changes are automatically done via https://github.com/ome/omero-documentation/pull/2300

PR #2293, opened 2023-02-06
Rsp

jburel:
Exclude royalsocietypublishing from the linkcheck 
This leads at the moment to a 403


jburel:
Merging to fix latest-ci

PR #2304, opened 2023-03-23
Indicate third-party password providers should be lazy initialized

sbesson:
See https://github.com/ome/omero-server/pull/160

Clarifies the expectation for a third-party password provider with the current declaration order for the services.  A minimal example of password provider extension following these guidelines and compatible with OMERO.server 5.6.7 can be found in https://github.com/sbesson/omero-passwordprovider-extension.


jburel:
Changes look good. 
While looking at it I noticed a problem in the redirect (fixed) and in the autogen script (fixed)
Proposing to merge and tag as 5.6.7-2

PR #2307, opened 2023-03-28
remove redirect leading to 443

jburel:
this is currently leading to error on latest CI. Removing the redirect

### DIFF
--- a/omero/developers/index.rst
+++ b/omero/developers/index.rst
@@ -31,7 +31,7 @@ with commercial licenses and customization available from |Glencoe|. You can
 read about how OMERO has developed since the project started in the
 :doc:`/users/history`.
 
-.. _GNU General Public License (GPL): https://www.gnu.org/licenses/gpl-3.0.html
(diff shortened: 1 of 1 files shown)


jburel:
ci is back to normal. So we can either include or close

omero/developers/index.rst
```diff
@@ -31,7 +31,7 @@ with commercial licenses and customization available from |Glencoe|. You can
 read about how OMERO has developed since the project started in the
 :doc:`/users/history`.
 
-.. _GNU General Public License (GPL): https://www.gnu.org/copyleft/gpl.html
+.. _GNU General Public License (GPL): https://www.gnu.org/licenses/gpl-3.0.html
```
sbesson: An alternative would be
```suggestion
.. _GNU General Public License (GPL): https://www.gnu.org/licenses/licenses.html#GPL
```

As this link is specific to GPL 3 while some of the code is under GPL 2

jburel: This will work.
We did not see the error again on latest-ci

sbesson (approved):
Assuming CI is passing

PR #2310, opened 2023-04-11
Update of system requirements

chdietrich:
Hi all,

This PR makes some changes to the system requirements.
It's far from perfect, but probably still an improvement.
I look forward to any opinions.

**Changes:**

- Removed "1.33GHz":
You won't get anything (new or quad-core) with a lower clock speed.

- Replaced the "server future-proofed for 3-4 years" with something current
Dual socket 2014 CPU → Single socket 2023 CPU
SAS HDD → SSD
256 GB RAM → 8x 32 GB (match CPU memory channels)

- CPU
"limited by GHz" → "limited by single thread performance" (clock speed != performance)
Remove the reference to a dual-socket CPU. If you don't need more than 
96 cores / 2 TB RAM / 128 PCIe lanes, a dual-socket system is more of a disadvantage.

- Recommended client
2GB RAM → 8 GB RAM
HDD → SSD
I really don't know what makes sense for OMERO Insight, but you don't want to work on a 
current Windows or Linux desktop with just 2 GB RAM and a HDD nowadays.

Best regards,
Christian

### DIFF
--- a/omero/sysadmins/system-requirements.rst
+++ b/omero/sysadmins/system-requirements.rst
@@ -13,19 +13,17 @@ by any recently bought hardware.
 
 An OMERO.server specification for between 25-50 users might be:
 
--  Quad core 1.33GHz Intel or AMD CPU
+-  Quad core Intel or AMD CPU
 -  8GB RAM
--  500MB hard drive space for OMERO.server distribution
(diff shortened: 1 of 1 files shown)


jburel:
Thanks @chdietrich

PR #2313, opened 2023-05-10
fix failure on rtd

jburel:
see https://github.com/readthedocs/readthedocs.org/issues/10290


jburel:
Build is green. Merging so we can fix other PR

PR #2316, opened 2023-05-16
Drop OMERO 6.0 column from version requirements table

sbesson:
As I was looking into making amendments to the Ubuntu section and cover the state of Ubuntu 20.04/22.04, I found the OMERO 6 column both unnecessary and misleading. A similar concern had been raised in https://github.com/ome/omero-documentation/pull/2262#issuecomment-1180223773.
This PR proposes to focus the documentation on the current stable OMERO 5.6.x series which is what the system administrators will be primarily interested. As noted in the related PR, I could also see an argument for dropping the older OMERO 5.4 and OMERO 5.5 columns

### DIFF
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -2,8 +2,8 @@
 Version requirements
 ********************
 
-Summary of changes for OMERO 5.6 and provisional changes for 6.0
-================================================================
+Summary of changes for OMERO 5.6
+================================
 
 We aim to support OMERO on the environments specified below, based
 on the availability of support by upstream developers and operating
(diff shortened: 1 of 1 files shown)


joshmoore:
Considering the general (lack of) discussion around "OMERO 6", :+1: from me.

sbesson:
Note that the base requirements for OMERO.server and OMERO.web on the current mainline will likely be updated over the next few months to reflect the support expectations with several operating systems coming end of life in 2024. As a comparison point, see [this post](https://www.glencoesoftware.com/blog/2023/06/26/Updates-to-OMERO-Plus-OS-support.html) which documents Glencoe Software position in terms of upgrade support.

Unless there is a plan to establish an OMERO 6 timeline, it probably does not hurt to get this in before updating the support tables and reduce the number of rows to update.

jburel:
title will need to be adjusted i.e. remove ``provisional changes for 6.0``

jburel (approved):

PR #2319, opened 2023-06-03
exclude ci links from check

jburel:
merge-ci is currently down


jburel:
merge-ci is back so we can remove the exclusion

PR #2323, opened 2023-06-13
Add install docs for ubuntu 22.04

dominikl:
See title. Respective omero-install doc: https://github.com/ome/omero-install/pull/271 .


jburel:
We should probably indicate which version of postgresql is installed since we specify the Python and Java version
I believe the walkthrough  file will need to be updated. The marker might have changed following the simplification of the installation script

jburel:
Working on it due to conflict

PR #2341, opened 2023-08-30
remove reference to build number

jburel:
We use GHA. we no longer have a build number
see https://downloads.openmicroscopy.org/omero/5.6.8/artifacts/


dominikl (approved):
👍

PR #2343, opened 2023-08-30
Cleanup readme

jburel:
This PR removes some outdated information

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,26 +13,12 @@ Initial setup
 Sphinx
 ------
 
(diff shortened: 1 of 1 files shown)


pwalczysko:
This link http://sphinx.pocoo.org/tutorial.html which is cited in the doc is not working anymore (404). Do you want to point to a different sphinx tutorial instead ?

jburel:
@pwalczysko link updated

pwalczysko (approved):
Sorry, did not approve before because the RtD build was failing. Now green though.

PR #2344, opened 2023-09-04
remove pg version 10, no longer available for downloads

jburel:
No more version 10 available at https://download.postgresql.org/pub/repos/yum/


sbesson:
Is this also an incentive to modify the level of support for PSQL 10 in OMERO 5.6 which is currently `supported/suboptimal` to `supported/deprecated`?

dominikl (approved):
👍  Looks good to me.

PR #2345, opened 2023-09-05
use sphinx_rtd_theme

jburel:
Set rtd theme

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -21,3 +21,7 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+
+python:
+    install:
+    - requirements: requirements.txt
--- a/common/conf.py
+++ b/common/conf.py
@@ -234,7 +234,7 @@ rst_epilog = """
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = "sphinx_rtd_theme"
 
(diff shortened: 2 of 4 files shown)


sbesson:
This seems to fix the theme regression issue that has been observed between https://omero.readthedocs.io/en/v5.6.8-2/ and https://omero.readthedocs.io/en/v5.6.8-3/.

Is there an official link describing this as a the new canonical way to declare the RTD theme?
Will this be a change to apply across all rtd repositories (i.e. lots of upcoming noise)?

jburel:
We will have to propagate across all repos. 
I have not seen a page
But RTD might have dropped it at some point because it did not have support for Sphinx 7 see https://github.com/readthedocs/sphinx_rtd_theme/issues/1463

The issue is now solved

sbesson (approved):
I believe the breaking change was introduced in https://docs.readthedocs.io/en/stable/changelog.html#version-10-2-0 and more specifically https://github.com/readthedocs/readthedocs.org/pull/10638

jburel:
```
We are not installing sphinx-rtd-theme automatically on new projects. However, we are setting sphinx_rtd_theme automatically on their docs/conf.py.
```
This did not happen for us since we are using ``omero/conf.py`` (even if it is not a new project) so some of our repos will be okay

snoopycrimecop:
Empty PR description. Please add a short summary of the PR scope and some testing instructions.

PR #2347, opened 2023-09-13
pin version to match rtd ones

jburel:
The requirements file is used by merge-ci job
this PR pins the requirements so it matches the version used on rtd

--exclude

### DIFF
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -392,8 +392,8 @@ def copy_legacy_redirects(app, exception):
     ]
     if app.builder.name == 'html':
         for html_src_path in redirect_files:
-            target_path = app.outdir + '/' + html_src_path
-            src_path = app.srcdir + '/' + html_src_path
+            target_path = (app.outdir / html_src_path)
+            src_path = (app.srcdir / html_src_path)
             if os.path.isfile(src_path):
                 target_dir = os.path.dirname(target_path)
                 if not os.path.exists(target_dir):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-Sphinx
-sphinx-rtd-theme
+sphinx==7.2.0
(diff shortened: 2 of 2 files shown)


sbesson (approved):
Pinning the versions in the requirements file is definitely one way to fixes the mismatch between the readthedocs build environment and the OME Jenkins CI build environment. Additionally, this meets the recommendations of the [readthedocs guide](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#use-a-requirements-file-for-python-dependencies) and should reduce the occasional cases of failures associated with upstream releases containing backward-incompatible changes.

No objection from my side to the fix. Note this raises a few wider policy questions that will need to be decided by the docs maintainers:
1- does the same policy need be enforced across the 20+ RTD projects owned by OME?
2- should these requirements files be periodically reviewed and updated?

jburel:
We have to review the repositories in order to make sure we have the correct version of the config file so the pinning can occur at the same time
A review will need to happen at regular intervals

jburel:
it requires Python 3.9
This is the error on merge-ci (python 3.8 is installed)
```
Ignored the following versions that require a different python version: 7.2.0 Requires-Python >=3.9; 7.2.1 Requires-Python >=3.9; 7.2.2 Requires-Python >=3.9; 7.2.3 Requires-Python >=3.9; 7.2.4 Requires-Python >=3.9; 7.2.5 Requires-Python >=3.9; 7.2.6 Requires-Python >=3.9
```

jburel:
--exclude

sbesson (commented):
While working on https://github.com/ome/omero-documentation/issues/2434, I noticed the requirements.txt are pinned to versions older than the ones proposed in this PR. Should this effort be restarted in 2026?

jburel:
We are now using Python 3.11 on merge-ci

PR #2352, opened 2023-10-06
Cap table to 3.9

jburel:
Cap tables installations

### DIFF
--- a/omero/conf.py
+++ b/omero/conf.py
@@ -175,9 +175,7 @@ rst_epilog += """
 .. |javaversion_recommended| replace:: 11
 .. |javaversion_min| replace:: 8
 .. |version_dropbox|  replace:: %s
-.. |version_tables_cap|  replace:: 3.9
 .. |version_openmicroscopy|  replace:: %s
-
 .. |Broken| image:: /images/broken.png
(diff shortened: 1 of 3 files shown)


sbesson (changes requested):
See https://github.com/PyTables/PyTables/releases/tag/v3.9.1 looks like Pytables has officially dropped support for Python 3.8  in the 3.9.x line so I think we can use this semantics in the documentation.

sbesson (changes requested):
As per https://github.com/PyTables/PyTables/issues/1062#issuecomment-1752482749 and the yanking of PyTables 3.9.0 combined with the correct metadata requirements in PyTables 3.9.1, it feels like there is no more need to have specific capping instructions either in the developer documentation or in the troubleshotting and `pip install tables` should install the latest supported PyTables version on any Python 3.8+ environment.

jburel:
@sbesson I have now removed the cap and only indicated the Python supported version

sbesson (approved):
:+1: The intent is to raise the minimal supported version of Python to 3.8 anyways right?

jburel:
It is yes.

PR #2355, opened 2023-11-02
Update python Tables code to use 'Image' column

will-moore:
See https://github.com/ome/omero-documentation/pull/2354

This PR updates the Python tables code with improvements suggested at that PR, in particular the use of "Image" column so that data can be associated with images.

### DIFF
--- a/omero/developers/Python.rst
+++ b/omero/developers/Python.rst
@@ -656,52 +656,62 @@ Write data
 OMERO tables
 ^^^^^^^^^^^^
 
--  **Create a name for the Original File (should be unique)**
+-  **Create some sample data. One table row per Image in Dataset**
+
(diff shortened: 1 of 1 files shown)


jburel:
@will-moore Do you remember why this PR was not merged alongside #2357?

will-moore:
No, I guess it just never got tested, reviewed, approved etc.

pwalczysko:
Strangely, the build here started to work... What would you prefer @will-moore , that we work on the #2520 which is rebased or should I close it ?

will-moore:
Thanks @pwalczysko - If #2520 is building then let's go with that. The original request on forums for better docs is still valid I think. Closing this one...

PR #2359, opened 2023-11-07
Ubuntu2204 rocky9

jburel:
This combines the web installation and server installation.
This is to avoid issue with autogen PR


jburel:
See https://github.com/ome/omero-install/pull/278

dominikl:
Looks good, thanks @jburel  👍

PR #2366, opened 2023-11-16
Replace "help" with "guides"

joshmoore:
I noticed that the index.rst still mentioned "help" and I wanted to update just that listing, but then I noticed that `:help:` is used pervasively throughout the guides. Happy to close this if this is considered confusing, but I'd rather not take on getting rid of `:help:` at the moment.


jburel:
replaced by #2372

PR #2412, opened 2024-04-26
Move background reading section up one level

sbesson:
This section contains a few essential pages notably around versions requirements and the current layout hides these pages from the left-side navigation

### DIFF
--- a/omero/sysadmins/index.rst
+++ b/omero/sysadmins/index.rst
@@ -54,11 +54,12 @@ an existing component, or just looking for more background information, there
 is a section about the server within the :doc:`/developers/index`;
 the best starting point is the :doc:`/developers/Server` for developers.
 
+******************
 Background reading
-##################
+******************
 
 .. toctree::
-    :maxdepth: 2
+    :maxdepth: 1
     :titlesonly:
 
     whatsnew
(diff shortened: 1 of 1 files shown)


sbesson:
See https://omero--2412.org.readthedocs.build/en/2412/sysadmins/index.html

Also opened to alternate suggestions about the section title (`Background reading` does not convey the importance of the underlying pages).

will-moore:
"Prerequisite reading" or "Important background reading" or "Essential background reading" ?

omero/sysadmins/index.rst
```diff
@@ -54,11 +54,12 @@ an existing component, or just looking for more background information, there
 is a section about the server within the :doc:`/developers/index`;
 the best starting point is the :doc:`/developers/Server` for developers.
 
+******************
 Background reading
```
pwalczysko:
```suggestion
Requirements, examples, limitations
```

pwalczysko (changes requested):
Maybe https://github.com/ome/omero-documentation/pull/2412#discussion_r1581083044 - imho, we should not try to shoehorn very important and slightly heterogeneous topics into one title - suggesting to name the most important ones explicitly

omero/sysadmins/index.rst
sbesson: Two other options would be to split into separate sections or move these pages at the top-level of the hierarchy (i.e. don't try and nest them under a heading as we can't rely define it)

pwalczysko (approved):

PR #2420, opened 2024-05-17
Fix omero.security.ignore_case markup 

sbesson:
See https://omero.readthedocs.io/en/v5.6.11-1/sysadmins/config.html#omero.security.ignore_case for the current rendering.

With this PR the warning should display correctly.
The change has also been applied upstream in https://github.com/ome/omero-common/pull/44

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,9 +13,10 @@ Initial setup
 Sphinx
 ------
 
-The Sphinx documentation system can be obtained by issuing::
+The Sphinx documentation system and the readthedocs theme can be obtained by issuing::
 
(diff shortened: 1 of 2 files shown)


jburel:
I think it will be better to point to https://github.com/ome/omero-documentation/blob/develop/requirements.txt
It is now recommended by RTD to pin the version.
This is what is currently used when building the documentation

jburel (approved):

jburel:
Warning correctly displayed at https://omero--2420.org.readthedocs.build/en/2420/sysadmins/config.html

PR #2424, opened 2024-05-21
Clarify CSRF token usage

DavidStirling:
Due to Django's [default security settings](https://docs.djangoproject.com/en/5.0/ref/csrf/), users connecting to the JSON API over HTTPS URLs tend to get "CSRF token invalid" messages when trying to use the API with the current instructions. This PR adds a bit of text to explain the need to supply a `Referer` header for the token checks to succeed.

### DIFF
--- a/omero/developers/json-api.rst
+++ b/omero/developers/json-api.rst
@@ -319,6 +319,10 @@ cookie of any request or from the following URL:
       "data": "eNoVq528bOqlhQqbCzKuviODTRX3PUO2"
     }
 
+If you are using the API over HTTPS, you will also need to set the ``Referer`` key in 
+request headers when supplying the CSRF token to other endpoints. The referer URI 
+needs to be an address on the same domain as the server for the token to be accepted.
+
 
 Login
 ^^^^^


omero/developers/json-api.rst
```diff
@@ -319,6 +319,10 @@ cookie of any request or from the following URL:
       "data": "eNoVq528bOqlhQqbCzKuviODTRX3PUO2"
     }
 
+If you are using the API over HTTPS, you will also need to set the ``Referer`` key in 
+request headers when supplying the CSRF token to other endpoints. The referer URI
```
will-moore, line 323: This looks like very useful info which I think I would have found useful in the past when failing to fix CORS issues! 

Just reading about the miss-spelling of the ``Referer`` header key at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy 
"The original header name [Referer](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer) is a misspelling of the word "referrer". The Referrer-Policy header does not share this misspelling."

So instead of "referer" in this line, you could use "referrer" but it might be clearer to repeat the ``Referer``.

DavidStirling: I did check this when testing and found that it specifically requires `Referer` rather than `Referrer`, although it's not case-sensitive. Could add "Yes, it really is spelled like that!" 😄

PR #2450, opened 2024-07-26
Update prep 5.6.12

jburel:
Add information about upgrade to OMERO 5.6.12

### DIFF
--- a/omero/sysadmins/server-upgrade.rst
+++ b/omero/sysadmins/server-upgrade.rst
@@ -25,6 +25,9 @@ the level of detail; upgrading should be a straightforward process.
     may be required beyond simply running the SQL upgrade scripts
     described below.
 
+.. warning::
+    Upgrading to OMERO 5.6.12 or higher requires an upgrade to OMERO.py 5.19.4 or higher
+
 Upgrade checklist
 -----------------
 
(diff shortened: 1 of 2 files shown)


omero/sysadmins/server-upgrade.rst
```diff
@@ -25,6 +25,9 @@ the level of detail; upgrading should be a straightforward process.
     may be required beyond simply running the SQL upgrade scripts
     described below.
 
+.. warning::
+    Upgrading to OMERO 5.6.12 or higher requires an upgrade to OMERO.py 5.19.4 or higher
+
```
sbesson, line 28: An alternative place for putting this warning would be the `Dependencies` section which talks specifically about upgrading OMERO.py - https://omero.readthedocs.io/en/stable/sysadmins/server-upgrade.html#dependencies

How is the minimal OMERO.py version maintained for this section? Is it automatically bumped to the latest version of is it bumped when there are specific requirements like the configurable OMERO.table module

jburel: Good point,  I have moved it there.
the versions are usually automatically bumped, I had to disable it due to the omero-common formatting
Now that the problem will be fixed with 5.6.12, I will re-activate it

sbesson (approved):
Based on https://omero--2450.org.readthedocs.build/en/2450/sysadmins/server-upgrade.html#dependencies and https://omero--2450.org.readthedocs.build/en/2450/sysadmins/troubleshooting.html#tables-service-inactive

PR #2461, opened 2024-11-19
Mark Java 8 as deprecated

jburel:
Mark Java 8 as deprecated


sbesson (approved):
Makes sense to me. Although the server components should retain runtime compatiblity, Java 8 has been effectively untested by the OME CI for months.

This section will probably need to additional attention to communicate the plan for more recent Java LTS versions but this is outside the scope of the upcoming OMERO release

PR #2477, opened 2025-03-03
Trouble-shooting LUTs display refers to OMERO-web LUTs docs

will-moore:
See https://github.com/ome/omero-web/pull/603#pullrequestreview-2650501988

### DIFF
--- a/omero/sysadmins/index.rst
+++ b/omero/sysadmins/index.rst
@@ -167,6 +167,7 @@ tasks on behalf of all users.
 
     server-permissions
     restricted-admins
+    lookup-tables
 
 .. seealso::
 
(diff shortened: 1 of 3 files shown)


will-moore:
@jburel @sbesson I added a new LUTs page. Not sure under which section to link it on the index page, but I think the Sysadmin part of the docs makes sense since it's not really developers who will add LUTs.

sbesson:
Thanks @will-moore, in https://github.com/ome/omero-web/pull/603#issuecomment-2697651008, I had overlooked this section of the documentation which is indeed relevant for OMERO users with administrator privileges. No objection to adding this content here from my side

jburel:
We can link to the new page from installation or upgrade in follow-up PR.
Merging so it is ready for omero-web release

PR #2478, opened 2025-03-13
Install Python bindings in .venv3

jburel:
The problem was noticed when running on Ubuntu 24.04 and Python 3.12
The ice-python binding are not installed in the virtual env ``.ven3`` (this is expected by merge-ci for example)
Previously it is built from source when omero-web is installed in the virtual env 
See for example  https://github.com/ome/omero-documentation/actions/runs/13621954191/job/38072830354
```
  Running setup.py install for zeroc-ice: started
  Running setup.py install for zeroc-ice: still running...
  Running setup.py install for zeroc-ice: still running...
```
This won't work when we do the switch
This wheel will need to be updated when we switch again to Python 3.12 on Ubuntu 24.04

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -28,6 +28,10 @@ jobs:
         run: |
           cd $space/OMERO.server
           ./build.py
+      - name: Create Virtual environment
+        run: |
+          python -m venv $space/.venv3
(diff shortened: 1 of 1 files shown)


jburel:
I might move the logic to ome/action-ice

PR #2485, opened 2025-04-08
Fix recommended RAM

dominikl:
8x 32GB RAM is a bit excessive. Especially as the next paragraph says:
``` In summary, depending on hardware layout 16, 24 or 32 GB of RAM would be ideal for your OMERO server. If you have a separate database server more than 16 GB of RAM may not be of much benefit to you at all.```

This PR changes the recommendation to 32GB (which we use for nightshade too, afaik).


pwalczysko (approved):

pwalczysko:
PR makes sense. The 8 x is:

1. excessive
2. in conflict with paragraph which follows
3. our production server at UoD is 32 GB RAM, running for >10 years

PR #2491, opened 2025-06-11
Remove flags with undefined params

pwalczysko:
As a quick fix of https://github.com/ome/omero-documentation/issues/2490 - this wfm.

Possibly a bigger rehaul of the walkthrough is needed with addition of an envirnoment file.

Happy to discuss.

Nevertheless, the changes here are making the workflow from bronken -> working.

cc @jburel


jburel:
the change must be done in omeroweb-install
those pages are auto-generated

PR #2496, opened 2025-07-17
Update getObjects group filtering

### LINKED ISSUE #466: `getObjects` `"group"` option is not used as a query filter
Hi ome :)

When searching for an object as detailed [in the documentation](https://omero.readthedocs.io/en/stable/developers/Python.html#read-data), at least when the connection has been set over all the groups with `conn.SERVICE_OPTS.setOmeroGroup(-1)`, filtering by passing `opts = {"group": group.id}` has no effect.

 

```python
...
group = conn.getObject("ExperimenterGroup", 4)
conn.SERVICE_OPTS.setOmeroGroup(-1)
projects = conn.getObjects("Project", opts={'group': group.id})

print(f"Found {len(projects)}") # All the projects in the DB
```

Restricting the connection to the desired group fixes the issue — but then it is useless to pass the group id in `opt`

```python
...
group = conn.getObject("ExperimenterGroup", 4)
conn.SERVICE_OPTS.setOmeroGroup(group.id)
projects = conn.getObjects("Project")

print(f"Found {len(projects)}") # Only the projects in group 4
```

will-moore:
Fixes https://github.com/ome/omero-py/issues/466

### DIFF
--- a/omero/developers/Python.rst
+++ b/omero/developers/Python.rst
@@ -215,8 +215,9 @@ Read data
     # Load first 5 Projects, filtering by default group and owner
     my_exp_id = conn.getUser().getId()
     default_group_id = conn.getEventContext().groupId
+    # To query across all groups, use -1
+    conn.SERVICE_OPTS.setOmeroGroup(default_group_id)
     for project in conn.getObjects("Project", opts={'owner': my_exp_id,
-                                                'group': default_group_id,
                                                 'order_by': 'lower(obj.name)',
                                                 'limit': 5, 'offset': 0}):
         print_obj(project)


glyg:
Thanks  a lot for the quick fix Will

I understand the need for `ops` to be supple enough in its handling of the keys, but the fact that an unknown key is silently ignored can be a bit misleading, maybe a note on the matter in the docs could help (while this PR is opened :wink: ).

jburel:
@will-moore you may also want to adjust the code examples used to generate that documentation 
see https://github.com/ome/openmicroscopy/blob/develop/examples/Training/python/Read_Data.py#L54

will-moore:
@jburel Done

will-moore:
@glyg Hopefully once this docs bug is fixed there should be less scope for users picking invalid/ignored opts. I realise that docs for `opts` is limited but that's outside the scope of this quick fix.
@jburel Training samples code was updated. Good to merge this?

PR #2506, opened 2025-12-04
Fixes GitHub workflow generating documentation from release artifacts

sbesson:
This effort was primarily motivated to propagate some of the property changes associated with the [OMERO.web 5.30.0 release](https://www.openmicroscopy.org/2025/12/03/omero-web-5.30.0.html) to the OMERO documentation.

A few issues were found as part of an initial investigation:

- the auto-generation workflow has been broken for several months with https://github.com/ome/omero-documentation/pull/2500 being the latest open PR and many failing actions
- the `omero config parse` command is skipped in `autogen_openmicroscopy.sh` and several configuration properties are out-of-sync
- the workflow builds OMERO.server from source which has two downsides:
  * it increases significantly build times
   * it includes changes from merged but unreleased upstream changes into PRs against the reference documentation
- several recent builds failed in `autogen_db_version.py` with  an error of type `packaging.version.InvalidVersion: Invalid version: '4.3__0'`
- the create-pull-request action failed on my fork and is behind

To address all of the above, the following changes are proposed:

- OMERO.server is no longer built from source but the latest GitHub release asset is downloaded using `gh release download`, similarly to the OMERO.py GitHub workflow. The `SUFFIX` and `BUILD` variables are removed or disabled.
- `autogen_db_version.py` is updated to only detect SQL files and folders corresponding to release versions of the OMERO database i.e. under the form `<MAJOR>.<MINOR>__<PATCH>`, excluding historical schema versions as well as all intermediate DEV database versions. The usage of `packaging.version` is dropped in favour of the `get_mmap()`
- the `peter-evans/create-pull-request` action is updated to `v7` and uses `GITHUB_TOKEN` under the hood

Note this workflow is currently configured to run in a `cron` matter making changes hard to evaluate. As a proof of concept, https://github.com/sbesson/omero-documentation/pull/8 was generated by https://github.com/sbesson/omero-documentation/actions/runs/19940358557 which includes an extra commit on top of this branch to force the workflow and a build time <2min

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -36,7 +36,7 @@ jobs:
           $space/.venv3/bin/pip install https://github.com/glencoesoftware/zeroc-ice-py-ubuntu2204-x86_64/releases/download/20221004/zeroc_ice-3.6.5-cp310-cp310-linux_x86_64.whl
       - name: Update documentation
         run: |
-          WORKSPACE=$space BUILD=false SUFFIX=/dist bash autogen_omero.sh
+          WORKSPACE=$space bash autogen_omero.sh
       - name: Find modified files
         id: upstream_repo
         run: |
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -8,6 +8,7 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       ICE_HOME: /opt/ice-3.6.5 # location where Ice is installed
+      GH_TOKEN: ${{ github.token }}
     steps:
       - uses: actions/checkout@v4
       - name: Install Ice Java and Python binding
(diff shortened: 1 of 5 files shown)


.github/workflows/update.yml
```diff
@@ -21,20 +22,21 @@ jobs:
           cd ..
           WORKSPACE=$(pwd)
           echo "space="${WORKSPACE} >> $GITHUB_ENV
-          git clone https://github.com/ome/openmicroscopy OMERO.server
           git clone https://github.com/ome/omeroweb-install
           git clone https://github.com/ome/omero-install
-      - name: Build OMERO.server
+      - name: Download latest release of OMERO.server
```
jburel, line 27: This means that the configuration changes will only be included in the documentation after a release of the OMERO.server and not when PRs are merged. 
This means that as we have seen previously the configuration text might be incorrect or not correctly displayed. We will not be able to catch that. This is probably okay to pay that cost since we are also reducing the build time significantly

sbesson: Agreed, for simple changes like typos/descriptions, the generation step creates a bit of unnecessary friction as you might need to wait until the next release.

On the other side:
- the generation of the configuration properties page has been turned off in ba0ee84f589ddc56f14009f0fd5322f835c1bbdf for more than a year. So reenabling those is an improvement
- building OMERO.server from source across these repositories is currently not a viable option. We'll need to look into this issue separately as it is currently failing the builds on ome/openmicroscopy and blocking the release workflow of OMERO.server.

PR #2513, opened 2026-01-08
Remove postgres12 link as it is end of life

pwalczysko:
Removing postgres12 link as it is breaking the linkcheck in https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/11/console and postgres12 is end-of-life.

cc @will-moore 

The result of this PR is built in https://omero--2513.org.readthedocs.build/en/2513/sysadmins/version-requirements.html#version-provided-by-distribution

### DIFF
--- a/omero/sysadmins/version-requirements.rst
+++ b/omero/sysadmins/version-requirements.rst
@@ -242,7 +242,7 @@ If no version is provided, a suitable repository is indicated.
       - CentOS/RHEL
       - Ubuntu
     * - 12
-      - 9 (`postgresql <https://download.postgresql.org/pub/repos/yum/12/redhat/rhel-9-x86_64/>`__)
+      - 9
       - 24.04 (`postgresql <https://apt.postgresql.org/pub/repos/apt/dists/noble-pgdg/>`__)
     * - 13
       - 9


khaledk2:
I think it would be a good idea to remove PostgreSQL versions 12 and 13, as they are no longer supported, and add versions 17 and 18.

pwalczysko:
> I think it would be a good idea to remove PostgreSQL versions 12 and 13, as they are no longer supported, and add versions 17 and 18.

@khaledk2 Thank you, but the above table is one which says "Version provided by distribution". I do not think that the version provided by the respective (OS) distribution changed (even though these are EOL as you correctly remark).

The table of the recommended versions is in another place of that doc, see https://omero--2513.org.readthedocs.build/en/2513/sysadmins/version-requirements.html#omero-support-policies where the 17 and 18 are included and 12 and 13 are "unsupported/old"

will-moore (commented):
Removal makes sense for v 12 (404).

I noticed that https://download.postgresql.org/pub/repos/yum/13/redhat/rhel-9-x86_64/ is working for v 13 (which currently has no postgres link). Do you want to add that?

pwalczysko:
> I noticed that https://download.postgresql.org/pub/repos/yum/13/redhat/rhel-9-x86_64/ is working for v 13 (which currently has no postgres link). Do you want to add that?

No, as postgres 13 is also EOL and that link will break in very near future if added.

will-moore (approved):
OK 👍

pwalczysko:
@khaledk2 @will-moore  Thanks both for your comments. Imho they show that the "Versions provided by distribution" table is not giving us much joy, already for the second time (see discussion on it in https://github.com/ome/omero-documentation/pull/2507#discussion_r2627389887). I think we could improve on that in the next PR, but atm I will go ahead and merge this.

pwalczysko:
@khaledk2 @will-moore after studying https://www.postgresql.org/download/linux/redhat/, I think that the "Version provided by distro" table for postgres is actually accurrate. We do claim that 15, 16 are provided, which matches https://www.postgresql.org/download/linux/redhat/. The fact that we still list 12.13 is also in lines with the strategy taken on similar table in Python, see https://github.com/ome/omero-documentation/pull/2507#discussion_r2627389887. This means: there are distros of OS-es around which provided posgres 12, 13 (just as they provided python 3.9) and thus the claim in our table is valid imho.

PR #2514, opened 2026-01-09
Update Django documentation links to point at 5.2 LTS

sbesson:
Remove links to obsolete Django versions

### DIFF
--- a/common/conf.py
+++ b/common/conf.py
@@ -188,7 +188,7 @@ extlinks = {
     'zeroc': ('https://zeroc.com/%s', None),
     'zerocforum': ('https://forums.zeroc.com/discussion/%s', None),
     'zerocdoc': ('https://doc.zeroc.com/%s', None),
-    'djangodoc': ('https://docs.djangoproject.com/en/4.2/%s', None),
+    'djangodoc': ('https://docs.djangoproject.com/en/5.2/%s', None),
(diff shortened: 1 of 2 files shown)


pwalczysko (changes requested):
When I go to some ``:djangodoc:`` link in the docs which are built from this PR, e.g. https://omero--2514.org.readthedocs.build/en/2514/sysadmins/config.html#omero-web-caches, and click on it, I am ending on https://docs.djangoproject.com/en/4.2/topics/http/sessions/#using-cached-sessions (= 4.2, I was expecting 5.2). Is it because the build from this PR somehow does not take the ``common/conf.py`` change into account ?

sbesson:
I think you are correct and `omero/conf.py` is the correct place to update. Fixed in d536f6b and the external link has been removed from `common/conf.py`.

There is a likely a case for removing `common/conf.py` entirely as I believe it made sense when multiple documentation sets were in this repository but that's outside the scope of this PR

pwalczysko:
Thank you @sbesson , checked several :djangodoc: links and with you last commit all point to 5.2. 

lgtm

pwalczysko (approved):

knabar (approved):
Built the documentation and ran the link check, all Django documentation links point to 5.2.

PR #2524, opened 2026-02-16
Update autogen_dependencies.sh to consume the OMERO.server

sbesson:
See https://github.com/ome/omero-documentation/pull/2523#issuecomment-3907380981

The current auto-generation logic download a version of OMERO.server behind a NGINX redirect which might be out-of-sync with the latest released GitHub artifact which is downloaded in the GitHub workflow. Similarly to other scripts, this passes the `WORKSPACE` environment variable and introspects the version of the OMERO server component from the existing bundle.

### DIFF
--- a/omero/autogen_dependencies.sh
+++ b/omero/autogen_dependencies.sh
@@ -11,6 +11,9 @@
 #    - check version of the dependencies
 
 PREFIX="omero-"
+WORKSPACE=${WORKSPACE:-$(pwd)}
+WORKSPACE=${WORKSPACE%/}  # Remove trailing slashes
+export OMERODIR=${WORKSPACE}/OMERO.server
 
 # General java packages
 # Java packages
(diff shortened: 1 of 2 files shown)


pwalczysko:
OK, I will make an attempt to understand it:

The script you are editing is mentioning that it is being used by the CI jobs. Hence my attempt:

https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-server/80/consoleFull

is using (confusingly for me) the OMERO.server 5.6.3-xxx 

(e.g. 

```
DIST=OMERO.server-5.6.3-594-ea2e4b837b-ice36-b61
00:51:48 + unzip OMERO.server-5.6.3-594-ea2e4b837b-ice36-b61.zip
...
```

and the effect of this PR will be that tomorrow it will be using the ``OMERO.server-5.6.17xxx`` ?

sbesson:
@pwalczysko here CI refers to GitHub workflows which is the environments where these scripts are executed. I don't know whether these are also used in UoD Jenkins CI but this is not the scope of this work.

My understanding of the current workflow is as follows:
- https://github.com/ome/omero-documentation/blob/7e7e5b065df5dc12de382f593c687b3a3dc601b0/.github/workflows/update.yml#L30 downloads the latest release artifact from https://github.com/ome/openmicroscopy/releases
- this artifact is used by the script in several auto-generated tasks defined in https://github.com/ome/omero-documentation/blob/develop/omero/autogen_docs
- https://github.com/ome/omero-documentation/blob/develop/omero/autogen_dependencies.sh retrieves the latest GitHub tag on github.com/ome/openmicroscopy and compare it to https://github.com/ome/omero-documentation/blob/7e7e5b065df5dc12de382f593c687b3a3dc601b0/omero/conf_autogen.py#L3
- if the versions differ, it downloads a server zip from https://downloads.openmicroscopy.org/omero/5.6/server-ice36.zip and parses the versions of the component

In practice, this creates a possible race condition on release that can be illustrated by https://github.com/ome/omero-documentation/pull/2523/changes/83fc944ee890041130e70a091475bcbae2aedd8a#diff-7b02055e006552117176ae6e5776b970d87d12350f2795eaac4ebb4a23a17ada. Following the tag pushed to https://github.com/ome/openmicroscopy/releases/tag/v5.6.17, `version_openmicroscopy` was upgraded to 5.6.17. However, the https://downloads.openmicroscopy.org redirects on UoD system were not yet updated and the rest of the workflow downloaded the 5.6.16 server and used it for all other components (like Bio-Formats). Due to the version check, there is no automatic logic allowing to correct these versions until the next OMERO.server release.

Fixing these versions by hand is an option. This tries to workaround these automation issues:
- re-using the release artifacts already downloaded by the workflow, currently OMERO.server 5.6.17 rather than downloading a server that might be out of sync
- removing the logic comparing the versions. Outside the release scenario, https://github.com/ome/omero-documentation/blob/7e7e5b065df5dc12de382f593c687b3a3dc601b0/omero/autogen_dependencies.sh#L76-L91 should be idempotent.

pwalczysko:
@sbesson Thank you very much for the great explanation. Sorry for being so far off-target with my first guess.

I think I understood. If I may try to put it in simple words, GitHub is more reliable/authoritative than the redirects on UoD systems.
This PR is reflecting that. Fully agreed.

pwalczysko (approved):

PR #2536, opened 2026-05-25
Replace faqs.org link

dominikl:
See title. Because they've moved http://www.faqs.org/ behind a captcha test.

### DIFF
--- a/omero/sysadmins/server-ldap.rst
+++ b/omero/sysadmins/server-ldap.rst
@@ -164,7 +164,7 @@ Compound Filters
 ^^^^^^^^^^^^^^^^
 
 .. note:: OMERO uses standard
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Makes complete sense. As a follow-up, we might want to review other RFC links in this document and switch from https://datatracker.ietf.org/ to https://www.rfc-editor.org/.

This should be merged once it has been validated by a passing linkcheck build.

pwalczysko:
from https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs/142/consoleFull - and the build is green

``01:54:10 (developers/Web/CSRF: line   14) ok        https://www.rfc-editor.org/rfc/rfc2616.html#section-9.1.1``

PR #2541, opened 2026-07-30
Add note about POST requests for public user

knabar:
https://github.com/ome/omero-web/pull/679 allows `POST` requests for certain functions (as of now, just `webgateway/table/*/slice`) even if `omero.web.public.get_only` is enabled.

This PR adds a note highlighting this in the documentation.

### DIFF
--- a/omero/sysadmins/public.rst
+++ b/omero/sysadmins/public.rst
@@ -64,8 +64,8 @@ To set this up on your OMERO.web installation:
      $ omero config set omero.web.public.password '<password>'
 
 - By default the public user is only allowed to perform GET requests.
-  (There are limited built-in exceptions allowing POST requests for certain
-  URLs that require it, but do not write or modify any data.) This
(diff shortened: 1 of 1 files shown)


sbesson (changes requested):
Thanks, the text looks good although I would remove the parentheses

knabar:
> Thanks, the text looks good although I would remove the parentheses

I removed them, but am not totally happy with the disconnect between the first and third sentences now. I can't think of a good rephrasing though without making everything a lot more verbose.

sbesson:
Re-reading, I wonder if a better flow would be to move the third sentence first i.e.:
- start with the description of the defulatbehavior (the public user cannot create, delete, modify)
- then clarify the implementation details (only allowing GET requests as well as a few exceptions)
- and document the configuration allowing  a more permissive behaviour.

sbesson (approved):
I think this reads well and is a great compromise between being transparent on which requests are blocked by default and staying not too verbose.

